PR #6, opened 2026-06-13
deps: modernize requirements for Python 3.12 (verified install + tests)

cgfixit:
Re-audited the proposed 'hardened' requirements against actual code usage and the
live package index/advisories, then installed + tested the result on Python 3.12.

Key corrections vs the proposal:
- Drop langgraph-checkpoint/-sqlite: unused (soul persistence is stdlib sqlite3;
  graph.py uses only StateGraph/END). Their CVEs don't apply to this app.
- langchain-core 1.3.3 (proposed 1.3.2 was still vulnerable to CVE-2026-44843).
- Keep numpy 1.26.4 (numpy 2.x breaks chromadb's onnxruntime dependency).
- pytest 9.0.3 (fixes CVE-2025-71176; proposed 8.3.4 still vulnerable).
- Keep nltk (required by retrieval/stemmer.py); remove unused matplotlib + pytest-asyncio.
- fastapi 0.136.3 pulls starlette>=1.0.1 (fixes Range-DoS + BadHost CVEs).
- langgraph 1.1.10 (conservative-mature track; '1.1.15' does not exist).

Adds constraints.txt (full transitive pin tree), pytest.ini, a two-step CPU-torch
install in README, and REQUIREMENTS_AUDIT.md documenting the CVE table and the
Python 3.12 install/import/unit/runtime test results.

https://claude.ai/code/session_01WKi36FBRzLtMKXN2tMt715


cgfixit:
“Embedded client users are NOT affected. Applications using only chromadb.PersistentClient in-process without a running server, exposed REST API, or trust_remote_code parameters are safe.”
That’s exactly PsyClaw’s configuration (embedded PersistentClient, binds to 127.0.0.1, no server, no trust_remote_code, embeddings supplied directly) — so it’s not exploitable at runtime

PR #14, opened 2026-06-16
Config-driven prompt filter & performance optimizations

cgfixit:
## Summary
This PR refactors the prompt injection filter to be entirely configuration-driven, adds caching to hot paths to reduce overhead, and includes several performance optimizations across the codebase.

## Key Changes

**Prompt Filter (utils/sanitizer.py)**
- Moved hardcoded `BANNED_PATTERNS` and `MAX_INPUT_CHARS` into `config.yaml` under `policy.prompt_filter`
- Added `_load_filter()` with `@lru_cache` to compile regex patterns once per config file instead of on every call
- Made filter configurable with `enabled`, `banned_patterns`, and `max_input_chars` settings
- Updated `check_input()` to return the query (enabling inline usage) and accept `config_path` parameter
- Updated `sanitize_chunk()` to accept `config_path` and use cached compiled patterns
- Improved error messages and documentation

**Rate Limiting (gate.py)**
- Added `_sweep_rate_limits()` to evict stale client entries and prevent unbounded memory growth
- Refactored `check_rate_limit()` to use a local `recent` list, reducing dict mutations
- Sweep runs at most once per window to keep the common path cheap

**Embeddings (retrieval/embeddings.py)**
- Added `_embeddings_cfg()` with `@lru_cache` to parse config once per path
- Fixed file descriptor leak by using context manager (`with open()`) instead of `yaml.safe_load(open())`
- Eliminates redundant config parsing on every embedding call

**Stemming (retrieval/stemmer.py)**
- Compiled token regex `_TOKEN_RE` once at import instead of on every `tokenize_and_stem()` call
- Added `@lru_cache(maxsize=100_000)` to `stem_token()` to cache stemming results
- Reduces regex compilation and stemming computation on repeated tokens

**Keyword Search (retrieval/hybrid_search.py)**
- Replaced full `sorted()` with `heapq.nlargest()` for top-k selection
- Reduces complexity from O(n log n) to O(n log k) per keyword query

**Indexing (retrieval/indexer.py)**
- Updated `sanitize_chunk()` call to pass `config_path` parameter

**CI (`.github/workflows/ci.yml`)**
- Added `cc` branch to CI triggers
- Changed pytest output from `--tb=no` to `--tb=short` for better debugging
- Removed `|| echo` fallback to properly fail on test failures

## Implementation Details
- All caching uses `@lru_cache` with appropriate `maxsize` values to balance memory and performance
- Config loading is cached per path to support multiple config files in tests
- Patterns are compiled with `re.IGNORECASE` once and reused, avoiding recompilation
- The filter can be disabled entirely via config for deployments that don't need it

https://claude.ai/code/session_01ExrmzbcxDWZm5sMnpuLRWB

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,9 +46,9 @@ jobs:
       - name: Run safe tests + coverage
         shell: bash
         run: |
-          pytest tests/test_sanitizer.py tests/test_rate_limit.py -q --tb=no \
+          pytest tests/test_sanitizer.py tests/test_rate_limit.py -q --tb=short \
             --cov=utils/sanitizer --cov=utils \
-            --cov-report=xml --cov-report=html --cov-report=term-missing || echo 'Safe tests done'
+            --cov-report=xml --cov-report=html --cov-report=term-missing
 
       - name: Upload Coverage Report
         uses: actions/upload-artifact@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: PsyClaw CI (Simplified + Coverage)
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, cc ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, cc ]
 
 jobs:
   test:
--- a/config.yaml
+++ b/config.yaml
@@ -76,20 +76,24 @@ policy:
     send_local_context_to_grok: false
   prompt_filter:
     enabled: true
+    # Regex patterns (single-quoted so backslashes reach the regex engine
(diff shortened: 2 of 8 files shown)


cgfixit:
Closing in favor of the cherry-pick route. The identical diff (8 files, +157/−59 — config-driven sanitizer, embeddings FD-leak fix, BM25 `heapq.nlargest`, stemmer cache, rate-limiter memory bound, CI triggers) has been brought onto `main` via #23 (merged as `b6f3fca`), which keeps `cc` as a standalone branch with no history join into `main`.

`cc` itself is untouched by this — its branch pointer (`f9c4849`) is unchanged. No action needed here.

---

PR #18, opened 2026-06-16
fix(personality): resolve the 11 remaining red unit tests (PersonalityManager v1.3 contract)

cgfixit:
## PR Set 1 of 2 — Unit-test deep-dive (follow-up to the suite audit in PR #16)

This is the scheduled **unit-test review** PR. The previous run (merged **PR #16**) audited the
whole suite, fixed the 3 collection errors, and left two red groups:

- **8 `test_sanitizer`** failures → owned by open **PR #14** (`cc → main`, config-driven filter). **Left untouched here.**
- **11 PersonalityManager** failures (`test_personality.py` ×8 + `test_personality_changes.py` ×3) → flagged as *"design decision required"*. **This PR resolves them.**

### Ground truth this run (Python 3.12 semantics, emulated RAG + API smoke)
```
before this PR:  pytest tests/ -q  →  19 failed, 47 passed, 0 collection errors
after  this PR:  pytest tests/ -q  →   8 failed, 58 passed, 0 collection errors
```
The only remaining red is the 8-test `test_sanitizer` suite (PR #14). A runtime smoke
(`PersonalityManager` propose/apply/reload/record lifecycle + `gate.py`/`graph.py` import) passes.

---

### The decision: align production code to the tests

The failing tests are not wrong — they encode the **v1.3 behaviors the architecture PDF already
promises** but `utils/personality.py` never shipped (most importantly **Invariant #5: a forensic
`audit_log` event on soul drift**). So the fix aligns the module to the tests rather than
weakening the tests.

| Failing expectation | Root cause (was) | Fix in `utils/personality.py` |
|---|---|---|
| `patch("utils.personality.audit_log")` | module never imported `audit_log` | `from utils.logger import audit_log`; best-effort call on **drift** + **apply** (wrapped so an audit-sink failure can't crash the manager) |
| `get_version() == 1` (int) | returned string `vN_<sha8>_<date>` → `'>=' str vs int` TypeError | `get_version()` → `int` (latest `soul_versions.id`); descriptive string moved to **new** `get_version_label()` |
| `propose_evolution(...) → {status, proposed_soul, current_sha, proposed_sha}` | returned `{diff, injection_flags, safe_to_apply, reason}` | return value is now a **superset** — adds the new keys, keeps the old ones so `gate.py`'s `/personality` endpoint callers don't break |
| `pm.conn` persistent + `row["query_hash"]` | each method opened/closed its own conn; no attr | single long-lived `self.conn` (`check_same_thread=False`, `row_factory=sqlite3.Row`, `busy_timeout=5000`) used everywhere |
| `pm.reload_soul()` | method was `reload()` only | added `reload_soul` as an alias of `reload()` |
| `pm.maintenance(ttl_days=…)` + TTL prune on init | prune only lived inside `record_interaction` | added `maintenance(ttl_days=None)`; **called from `__init__`** (v1.3 prune-on-boot) |
| "creates default soul when missing" (`"PsyClaw"` in content) | `_load_soul` set `soul_core=""`, wrote nothing | `_load_soul` writes `DEFAULT_SOUL` when `soul.md` is absent, then version-tracks it |
| atomic write — no `.tmp` left after apply | plain `write_text`, no atomicity | `apply_evolution` stages to `soul.md.tmp` then `os.replace`; still keeps a `.bak` |

### Backward-compatibility (callers checked before changing)
- `gate.py` `/personality` (`get_version`, `propose_evolution`, `apply_evolution`, `reload`) — int
  version is JSON-safe; `propose_evolution` superset is additive; `reload()` retained.
- `graph.py` (`get_system_prompt_additive`, `record_interaction(query_hash, outcome)`) — unchanged signatures.

### Reviewer note — thread safety
The shared `self.conn` is correct for PsyClaw's single-user local gateway (LM Studio), and
`busy_timeout` absorbs the rare overlapping write (e.g. a manual reload). If PsyClaw ever becomes
multi-tenant / high-QPS, revisit with a per-request connection or a write queue. Called out in the
module docstring and in `tests/TEST_SUITE_AUDIT.md` §8.

### Still out of scope (by design)
- **`test_sanitizer` (8)** — config-driven prompt filter, owned by **PR #14**. Touching it here
  would collide with that branch.
- The non-hermetic smells in `test_personality_changes.py` (hard-coded
  `/home/workdir/artifacts/...` path, at-import global monkeypatch of `audit_log`) are documented
  in the audit §3.6/§4; they don't block the suite now that the code satisfies the contract, and
  are best cleaned up alongside the PR #14 merge to avoid churn.

### For the next Claude Code session
1. Merge **PR #14** (`cc → main`) to clear the last 8 (`test_sanitizer`) and flip CI fully green.
2. Then drop the error-swallowing `|| echo` in `.github/workflows/ci.yml` and run the **whole**
   `pytest tests/` as the gate (audit §3 / §5.6).
3. Optionally de-non-hermeticize `test_personality_changes.py` (audit §3.6).

See `tests/TEST_SUITE_AUDIT.md` (§8 added here) for the full living record.

---
_Generated by [Claude Code](https://claude.ai/code/session_0169dmEttvQxToJtjwZNXBgP)_

### DIFF
--- a/tests/TEST_SUITE_AUDIT.md
+++ b/tests/TEST_SUITE_AUDIT.md
@@ -276,3 +276,34 @@ textual conflicts with PR #14 (which also edits `retrieval/stemmer.py`), the ste
 **Still red (intentionally, out of scope here):** `test_sanitizer` (8 — config-driven filter, **PR #14**),
 `test_personality` (8) and `test_personality_changes` (3) — pending the PersonalityManager
 contract decision tracked in the companion issue.
+
+---
+
+## 8. Follow-up PR — PersonalityManager contract resolved (2026-06-16)
+
+The PersonalityManager contract decision deferred in §3.5 / §5.2 has now been **made and
+implemented** in `utils/personality.py`. The direction chosen is **align production code to the
+tests**, because the tests encode the v1.3 behaviors the architecture PDF already promises
+(forensic audit on drift = Invariant #5, integer version, TTL maintenance, atomic soul writes).
+
+| Test expectation (was failing) | Change in `utils/personality.py` |
+|---|---|
+| `patch("utils.personality.audit_log")` | `from utils.logger import audit_log`; called (best-effort) on drift + apply |
+| `get_version() == 1` (int) | `get_version()` now returns the latest `soul_versions.id` as `int`; the old `vN_<sha8>_<date>` string moved to `get_version_label()` |
+| `propose_evolution(...) → {status, proposed_soul, current_sha, proposed_sha}` | return value is now a **superset** — adds those keys while keeping `diff / injection_flags / safe_to_apply / reason` (so `gate.py` callers don't break) |
+| `pm.conn` persistent, `row["query_hash"]` | single long-lived `self.conn` (`check_same_thread=False`, `row_factory=sqlite3.Row`, `busy_timeout=5000`) used by every method |
+| `pm.reload_soul()` | added as an alias of `reload()` |
+| `pm.maintenance(ttl_days=…)` + TTL prune on init | added `maintenance(ttl_days=None)`; called from `__init__` (v1.3 prune-on-boot) |
+| "creates default soul when missing" (`"PsyClaw"` in content) | `_load_soul` writes `DEFAULT_SOUL` when `soul.md` is absent, then version-tracks it |
+| atomic write — no `.tmp` left | `apply_evolution` stages to `soul.md.tmp` then `os.replace`; keeps a `.bak` |
+
+**Result:** `pytest tests/` → **58 passed, 8 failed, 0 collection errors** (was 47/19). The 11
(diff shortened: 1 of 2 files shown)


cgfixit:
Closing as superseded. The PersonalityManager v1.3 contract fix this PR implements has already landed on `main` via #20/#21 (commits `83fcef2` + `4c86260`) — `main` now has `get_version() -> int`, the `audit_log` import + drift/apply events, persistent `self.conn`, `reload_soul` alias, `maintenance()` + prune-on-init, default-soul creation, and atomic `.tmp` + `os.replace` writes. The 11 personality tests are green on `main`, and this branch now conflicts (`dirty`) against the merged version.

No code is lost by closing — `main` already satisfies the contract this PR targeted.

---

PR #58, opened 2026-06-19
ci: real ChromaDB + BM25 RAG smoke (replace hardcoded mock)

cgfixit:
## What

Replaces the CI "Emulated RAG Query Smoke" step with a **real** retrieval smoke that indexes the committed corpus into ChromaDB + BM25 and runs an actual `HybridRetriever` query against it.

## Why

The old step ran a base64-encoded `MockRetriever` whose `hybrid_search()` returned a hardcoded `{"text": "...", "score": 0.92}` dict. It:
- never touched ChromaDB, BM25, embeddings, or the real `HybridRetriever` — so it could pass even if retrieval was completely broken;
- printed the stale **"PsyClaw"** name (pre-rename).

It proved nothing about RAG. @cgrady92 added a populated corpus (`data/corpus/cyclaw_overview.md`) and wanted the smoke to actually index ChromaDB and answer a query the corpus document covers — this PR does that.

## How

New `tests/ci_rag_smoke.py`:
1. `retrieval.indexer.build_index()` → builds a real ChromaDB (semantic) + BM25 (keyword) index from `data/corpus`.
2. Instantiates the real `HybridRetriever` and runs `hybrid_search()` for *"What fusion method does CyClaw use to blend semantic and keyword results?"* — answered by `cyclaw_overview.md` ("RRF Fusion … blend semantic and keyword results").
3. Asserts a hit is returned and its fused score clears the configured `retrieval.min_score` gate — a genuine **vault hit**, exiting non-zero on a real regression.

No LLM is involved (CI has no LM Studio), so the smoke stops at retrieval — the part ChromaDB + BM25 actually own. Generation stays covered by the mocked gate/graph unit tests.

The prep step no longer overwrites the committed corpus with a stub, so the smoke indexes the real shipped document.

## Verification (local, Python 3.12.3)

```
=== Real Offline RAG Query Smoke (ChromaDB + BM25) ===
Building real index from data/corpus ...
[Indexer] Indexed 1/1 chunks
Query:         What fusion method does CyClaw use to blend semantic and keyword results?
Hits returned: 1
Top source:    data/corpus/cyclaw_overview.md
Top score:     0.050166 (min_score gate: 0.028 )
Real RAG query passed (vault hit above gate)   # exit 0
```

## Follow-up idea

The corpus is a single document today, so retrieval is degraded-to-semantic (BM25 has one doc). Adding 2–3 more corpus docs would let the smoke assert true **hybrid** fusion (RRF blending distinct semantic + keyword winners) and rank ordering — a natural next step now that real indexing is in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DNADoXjjDvJBuwQ2QBMQk6)_


cgfixit:
CI.yml is now a true RAG query emulation rather than a smoke test

PR #114, opened 2026-06-20
fix(retrieval): ChromaDB cosine distance metric (#99 #1)

cgfixit:
## What — implements PR #99 finding **#1** (HIGH)

ChromaDB collections were created with the default **`l2`** space, but embeddings are L2-normalized (`embeddings.py`). For unit vectors squared-Euclidean distance ∈ [0,4], so `hybrid_search`'s `score = 1 - distance` produced a **non-cosine, partly-negative** scale. Now:
```python
collection = client.create_collection(collection_name, metadata={"hnsw:space": "cosine"})
```
→ `distance = 1 - cos ∈ [0,2]`, so `1 - distance` is genuine cosine similarity.

**Ranking is unchanged for unit vectors** (L2 is monotonic in cos), so RRF order and the `min_score` gate are unaffected — only the surfaced `semantic_score` becomes correct. Requires a one-time **reindex**; CI's `Real RAG Query Smoke` calls `build_index()` (and `index/` is gitignored/built fresh), so it validates automatically.

## ⚠️ Scope change — #6 deferred to a follow-up

This PR originally also implemented **#6** (RRF-scaling single-path results). **CI surfaced that #6 needs a `min_score` re-tune I shouldn't guess**, so I reverted it here:

> The committed smoke corpus is 1–2 chunks. BM25 IDF over a single-doc corpus goes **negative**, so the keyword path returns nothing and the smoke query is served **single-path (semantic)**. Pre-#6 that returned a raw cosine score (~0.5, above `min_score=0.028`); #6's RRF-scaling forced it to `1/60 = 0.0167 < 0.028` → the smoke correctly failed as a vault miss. Making legit single-path results pass requires lowering/retuning `min_score` against real score data — a calibration decision for the maintainer.

So **#6 will land separately**, paired with the `min_score` re-tune (and possibly per-mode thresholds). This PR is now the clean, ranking-invariant, smoke-safe `#1` change only.

## Risk
Minimal — one-line collection metadata; ranking-invariant; CI smoke rebuilds and validates.

---
Part of the PR #99 backlog (`docs/ACTION_PLAN_PR99_2026-06-20.md`, PR #106).

### DIFF
--- a/retrieval/hybrid_search.py
+++ b/retrieval/hybrid_search.py
@@ -129,34 +129,6 @@ class HybridRetriever:
                 ))
         return hits
 
-    def _as_rrf_scaled(self, hits: List[SearchResult]) -> List[SearchResult]:
-        """Re-score a single surviving retrieval path onto the RRF scale.
-
-        When only one path returns, its raw scores (semantic ``1 - cos`` ∈ [-1,1],
-        or unbounded BM25) are NOT comparable to ``min_score``, which is tuned for
-        RRF magnitudes (~``1/rrf_k``). The gating ``score`` is therefore replaced
-        with the rank-based RRF term ``1/(rrf_k + rank)`` so the confidence gate
-        behaves consistently under degraded retrieval (PR #99 #6). The raw value is
-        preserved in ``semantic_score`` / ``keyword_score`` for provenance.
-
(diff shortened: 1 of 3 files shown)


cgfixit:
If chroma embeddings get jacked this was me ha

PR #181, opened 2026-06-21
[[[!!!]]].    - https://github.com/CGFixIT/CyClaw/pull/181#issuecomment-4763651067

cgfixit:
Follow-up to merged PR #179 (dependency manifest alignment).

## What changed
- Both `test` and `verify-skills` jobs now pass `-c constraints.txt` when installing from `requirements.txt`.
- This enforces the full direct pin set (including pydantic family sync, torch==2.6.0+cpu CVE gate, chromadb local-only, etc.) that was expanded in #179.
- Updated header and step comments to reference v1.5.0 + the python-development / cyclaw-advisor alignment work.
- No behavior change for happy path (pins were already identical); this just makes the hermetic gate explicit and consistent with Dockerfile + pyproject.toml best practice.

## Why (python-development rationale)
- Single source of truth (pyproject.toml) + constraints for any pip-based path prevents future drift.
- CI cache already listed constraints.txt; now we actually use it.
- Keeps the existing torch index dance (required for CPU wheel safety) while adding the constraint layer.
- Low-risk, high-signal improvement for reproducible CI.

## Verification
- Existing verify-install jobs and full matrix should continue to pass (same versions, just stricter constraint application).
- Future `uv` migration path remains open (this change is a stepping stone).

Ready for review/merge. This closes the loop on making the install gate match the declared manifests across Docker, CI, and local dev.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,9 @@
 #   - Explicit hermetic prepare (mkdirs, minimal soul.md, env var) cross-platform
 #   - Real RAG smoke validates ChromaDB 1.5.6 + BM25 + RRF on committed corpus
 #   - Coverage scoped to exercised modules only
-#   - Aligns with v1.4.0 invariants (reproducible CI security gate, Python 3.12, stable Chroma pin)
+#   - Aligns with v1.5.0 invariants + PR #179 dependency manifest alignment
+#     (constraints.txt now expanded and authoritative for pip hermetic installs;
+#      pyproject.toml remains single source of truth + uv recommended path)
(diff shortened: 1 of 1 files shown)


cgfixit:
Note2self check out versions dependabot suggested in PR #180

--

Bumps the pip-all group with 7 updates in the / directory:

Package	From	To
fastapi	0.137.2	0.138.0
langgraph	1.2.5	1.2.6
langchain-core	1.4.7	1.4.8
chromadb	1.5.6	1.5.9
pytest	9.1.0	9.1.1
torch	2.6.0+cpu	2.12.1
pydantic-core	2.46.4	2.47.0
Updates fastapi from 0.137.2 to 0.138.0

PR #187, opened 2026-06-22
ci: drop ineffective pip cache from lint workflow

cgfixit:
## What
Removes `cache: 'pip'` from the `actions/setup-python` step in `.github/workflows/lint.yml`.

## Why / benefit
The lint job installs **only** `ruff` (`run: pip install ruff`) and references no dependency manifest. However, `setup-python`'s pip cache computes its key from the repo's `requirements.txt` / `pyproject.toml` hashes. As a result:
- The restored cache never contains the `ruff` wheel (it was keyed on manifests that the job doesn't install), so there is effectively **no cache hit benefit**.
- The cache key churns on every unrelated dependency change, adding save/restore overhead for a wheel that installs in ~1s.

Dropping it simplifies the workflow and removes a misleading cache configuration. Net effect on runtime is neutral-to-positive.

## Risk to monitor
- Single-line removal; no behavior change to the actual lint step (`ruff check --select E,F,I,B,C4,S .`).
- The action SHA pin on `setup-python` is unchanged.
- YAML validated locally with `yaml.safe_load`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DStiigNTPQhXJ4qBkFnJCw)_


cgfixit:
@copilot do you concur? (Don't edit anything just yes or no for this one)

PR #327, opened 2026-06-29
fix(graph): query/soul-aware prompt budget — deterministic no-stall bound + richer context

cgfixit:
## Summary

Third and final follow-up to #325/#326 for the LM Studio **"stuck at 0% processing"** issue. A deeper 2-agent audit found a **residual gap** the earlier PRs left open, and this PR closes it deterministically while tuning for a larger context window (per operator decision to run LM Studio at 10000–12288 context).

## The residual gap (why #326 wasn't fully sufficient)

#326 bounded the **retrieved-context block**, but that budget was **fixed** — it ignored the **query** and **soul**. So the *total* prompt was still unbounded:

```
prompt = soul_preamble + "USER QUERY: " + query + framing + context + trailing
                ^bounded? no        ^no            ^#326 only bounded this part
```

Worst case (filter enabled, query ≤4000 chars ≈1000 tok): `soul + 1000 + 2000(ctx) + 2048(out)` ≈ **5,500+ tokens** — a 5000-token LM Studio context could still stall. Filter disabled → query up to 65k chars → massive overflow.

## Fix: query/soul-aware adaptive budget

`graph.py` now reserves room for `soul + query + framing` out of `retrieval.max_context_tokens` and gives the remainder to retrieved context, so the **total prompt input is deterministically ≤ `max_context_tokens`**. The no-stall guarantee becomes a simple operator-checkable formula:

```
LM Studio context length  >=  max_context_tokens + max_tokens + ~1500 headroom
```

| File | Change |
|---|---|
| `graph.py` | New `_context_char_budget()` (query/soul-aware) used by `local_llm_node` and `offline_best_effort_node` (`limit=5`). Offline path is now **richer but bounded** (was a fixed `char_cap=300`). Warning log when the assembled prompt still exceeds the budget (large query/soul) so any stall is diagnosable. |
| `config.yaml` | `retrieval.max_context_tokens` 2000→**4000** (richer RAG + offline context); `models.local_llm.max_tokens` 2048→**3000** (longer answers, both paths). Documented the no-stall formula + **recommended LM Studio context 10000–12288**. |
| `tests/test_graph.py` | New `TestLocalLlmPromptBudget`: total prompt bounded; context collapses to the floor on a huge query; small docs preserved (regression); offline prompt bounded. |

## api_key (the recurring question) — verified, no change needed

`self.api_key = str(llm_cfg.get("api_key") or "").strip()`; header sent only when truthy.

| config.yaml form | header sent | default LM Studio | "Require API Key" on |
|---|---|---|---|
| commented out | none | ✅ works | ✗ needs a key |
| `api_key: 123456` (int) | `Bearer 123456` | ✅ (coerced) | ✅ if matches |
| `api_key: "123456"` (str) | `Bearer 123456` | ✅ | ✅ if matches |

Wrong key with "Require API Key" on → fast **401** (4xx not retried) — never a 0% hang. Quoted string is recommended (already documented).

## Testing / verification (high-confidence bar)
- `GROK_API_KEY=dummy pytest tests/` → **838 passed, 12 skipped** (live Postgres, expected).
- `ruff check graph.py` → clean.
- **Standalone worst-case proof**: 2KB soul + 4000-char (filter-capped) query + 5 oversized chunks at the new config values → RAG **~6,996 tok** and offline **~6,979 tok** (prompt + max_tokens), both ≤ the recommended 10,000-token context. Input bounds to ~`max_context_tokens` as designed.

## Out of scope
- `grok_fallback` char_cap=200 cost guard (external paid API) — unchanged.
- A full API-surface re-audit (api_key, health.mode, /query timeout, soul endpoints, all terminal.html bodies vs schemas) found **no other issues** post-#326.

---
_Generated by [Claude Code](https://claude.ai/code/session_01P32jhvzh51LBRTH7arVrzv)_

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -32,17 +32,16 @@ models:
     # doesn't parse a bare number as an int.
 #    api_key: "your-lm-studio-key"
     model: "qwen2.5-7b-instruct"
-    # IMPORTANT: LM Studio must load the model with a context length >
-    # (prompt tokens + max_tokens). The local_llm node injects soul preamble +
-    # up to 5 full untruncated corpus chunks, so the prompt alone can reach
-    # ~1000 tokens on a vault hit. max_tokens is the RESERVED OUTPUT BUDGET —
-    # LM Studio allocates it up front before generating token #1. If
-    # prompt + max_tokens > loaded context, LM Studio stalls at "0% processing"
-    # non-deterministically (vault hits stall reliably; vault misses may squeak
-    # through because their prompt is smaller). Keep max_tokens well below the
-    # context window to leave headroom. At context=5000: 2048 leaves ~3000
-    # tokens for the prompt. Raise LM Studio context to 8192+ for more output.
-    max_tokens: 2048
+    # max_tokens = RESERVED OUTPUT BUDGET. LM Studio allocates it up front before
+    # generating token #1, so if (prompt tokens + max_tokens) > the loaded context
+    # length, LM Studio stalls at "0% processing". The graph bounds the prompt
+    # INPUT to retrieval.max_context_tokens (query/soul-aware budget in
+    # graph.py local_llm_node / offline_best_effort_node), so the stall is
+    # prevented deterministically as long as:
+    #   LM Studio context length  >=  max_context_tokens + max_tokens + ~1500 headroom
+    # RECOMMENDED LM Studio context length: 10000-12288 (10k-12k).
+    # With max_context_tokens=4000 + max_tokens=3000 that is 7000 + headroom -> fits.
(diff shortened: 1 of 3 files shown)


cgfixit:
Confirmed this did significantly help but occasionally for offline_best_effort aka more inference heavy queries it occasionally stalls, especially with 2 streams so that may be a resource issue hard to say for sure but either way works much better now it was def the max token thing - tldr it's not just setting them all equal lm studios should be higher than the rag and overall config max token to account for the soul file and other preambles to system prompt and the tokens generated in response, etc - set lm to 12-15k and leave config as is for now

PR #363, opened 2026-06-30
[Low-Priority][gate]:  pass resolved config path to check_input

cgfixit:
## What

Stores the resolved config path (`_BASE_DIR / "config.yaml"`) as `_CONFIG_PATH` and passes it to `check_input()` instead of relying on the default relative `"config.yaml"`.

## Why / benefit

`gate.py` anchors all file access to `_BASE_DIR` (resolved from `__file__`) so the server works regardless of cwd. But `check_input(req.query)` was called without a config path, falling back to `"config.yaml"` — a relative open that fails if cwd ≠ repo root. This is the same fragility `_BASE_DIR` was introduced to prevent (e.g. Windows double-click launches, systemd with a different WorkingDirectory).

## Changes

- **`gate.py`**: Extract `_CONFIG_PATH = str(_BASE_DIR / "config.yaml")`, reuse it for the `open()` call and for `check_input(req.query, _CONFIG_PATH)`.

## Risk to monitor

- The `_load_filter` lru_cache in `sanitizer.py` is keyed by `config_path`. With the absolute path, the cache key changes from `"config.yaml"` to the full path. This is harmless — the cache still holds one entry — but means a test that calls `check_input("x", "config.yaml")` and a gate.py call with the absolute path get separate cache slots. No functional impact.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AdWz8664ojWs48fnfBtM7U)_

### DIFF
--- a/gate.py
+++ b/gate.py
@@ -133,7 +133,8 @@ from utils.ratelimit import RateLimiter
 # The relative open crashes when gate.py is launched from a non-repo-root cwd
 # (e.g. double-clicked on Windows) — the very failure mode _BASE_DIR exists to
 # prevent — and the second read was pure startup overhead. One read, reused.
-with open(_BASE_DIR / "config.yaml", encoding="utf-8") as _cfg_f:
+_CONFIG_PATH = str(_BASE_DIR / "config.yaml")
+with open(_CONFIG_PATH, encoding="utf-8") as _cfg_f:
     cfg = yaml.safe_load(_cfg_f) or {}
 # Fail fast on an out-of-range retrieval tunable (e.g. min_score > 1 silently
 # forces every query to user_gate; top_k <= 0 breaks retrieval). Without this the
@@ -308,7 +309,7 @@ async def query_endpoint(request: Request, req: QueryRequest):
         )
 
     try:
-        check_input(req.query)
+        check_input(req.query, _CONFIG_PATH)
     except PromptInjectionError as e:
         # Pass the full query: audit_log() SHA-256-hashes the "query" field, so
         # truncating here yields a hash of only the first 50 chars that diverges
--- a/tests/test_gate.py
+++ b/tests/test_gate.py
@@ -41,7 +41,7 @@ def client(tmp_path):
          patch("gate.HybridRetriever") as MockRet, \
          patch("gate.LocalLLMClient") as MockLLM, \
          patch("gate.build_graph") as MockBuild, \
-         patch("gate.check_input", side_effect=lambda q: q), \
+         patch("gate.check_input", side_effect=lambda q, *_a, **_kw: q), \
          patch("gate.check_all", return_value=[]):
 
         retriever = MockRetriever(MOCK_HIGH_SCORE_RESULTS)


cgfixit:
Honest assessment — you're right that the benefit is narrow. Here's the full picture:

## What it actually does

`gate.py` already loads its own config via the resolved `_BASE_DIR / "config.yaml"` (absolute path). But the `check_input(req.query)` call on the `/query` endpoint was relying on the default parameter `config_path="config.yaml"` (relative). This PR makes the injection filter use the same resolved path gate.py already uses for everything else.

## When it matters

Only when the process CWD ≠ the repo root. Standard deployment (`cyclaw-server` entry point, `uvicorn gate:app` from repo root) always has CWD = repo root, so the relative path works fine today.

Scenarios where CWD could differ: a systemd unit with a wrong `WorkingDirectory`, a Docker image with a non-repo `WORKDIR`, double-clicking on Windows, or an orchestrator that launches from `/`. In those cases, gate.py's own config load (line 137, `_BASE_DIR / "config.yaml"`) would succeed but then `check_input` would fail with `FileNotFoundError` on the relative `"config.yaml"` — the injection filter would crash the `/query` endpoint even though the server started fine.

## Subtle implication: `lru_cache` key divergence

`_load_filter()` in `utils/sanitizer.py` uses `@lru_cache(maxsize=8)` keyed on `config_path`. Before this PR, both gate.py and the indexer hit the cache with the same key (`"config.yaml"`). After this PR, gate.py hits it with `"/home/.../CyClaw/config.yaml"` while the indexer still uses `"config.yaml"`. Same file, two cache slots.

**Functional impact: none.** Gate and the indexer run in separate processes (`uvicorn` vs `cyclaw-index` CLI), so they never share the cache anyway. Even if they did, both slots load the same file — it's a wasted slot out of 8, not a correctness issue.

## Actual risk

- **Regression risk is very low.** The diff is 3 lines: one constant extraction, one argument addition, one test mock update. The test suite covers all `/query` paths and passes.
- **No behavioral change in the happy path.** When CWD = repo root, `"config.yaml"` and `"/abs/path/config.yaml"` resolve to the same file.
- **No new dependencies or architectural changes.**

## Honest take

This is defensive hardening for a scenario that's unlikely in your current deployment but real in edge cases. The consistency argument is sound (gate.py already uses `_BASE_DIR` for everything else — this was the one gap). But if you're weighing it against the principle of "don't change what works," it's a judgment call. The risk of merging is near-zero; the risk of not merging is also near-zero unless your deployment story changes.

--

"If you want to close it as not-worth-the-diff, that's a perfectly reasonable call — the existing code works for all standard entry points."
-cc

cgfixit:
Here's the full picture:

## What it actually does

`gate.py` already loads its own config via the resolved `_BASE_DIR / "config.yaml"` (absolute path). But the `check_input(req.query)` call on the `/query` endpoint was relying on the default parameter `config_path="config.yaml"` (relative). This PR makes the injection filter use the same resolved path gate.py already uses for everything else.

## When it matters

Only when the process CWD ≠ the repo root. Standard deployment (`cyclaw-server` entry point, `uvicorn gate:app` from repo root) always has CWD = repo root, so the relative path works fine today.

Scenarios where CWD could differ: a systemd unit with a wrong `WorkingDirectory`, a Docker image with a non-repo `WORKDIR`, double-clicking on Windows, or an orchestrator that launches from `/`. In those cases, gate.py's own config load (line 137, `_BASE_DIR / "config.yaml"`) would succeed but then `check_input` would fail with `FileNotFoundError` on the relative `"config.yaml"` — the injection filter would crash the `/query` endpoint even though the server started fine.

## Subtle implication: `lru_cache` key divergence

`_load_filter()` in `utils/sanitizer.py` uses `@lru_cache(maxsize=8)` keyed on `config_path`. Before this PR, both gate.py and the indexer hit the cache with the same key (`"config.yaml"`). After this PR, gate.py hits it with `"/home/.../CyClaw/config.yaml"` while the indexer still uses `"config.yaml"`. Same file, two cache slots.

**Functional impact: none.** Gate and the indexer run in separate processes (`uvicorn` vs `cyclaw-index` CLI), so they never share the cache anyway. Even if they did, both slots load the same file — it's a wasted slot out of 8, not a correctness issue.

## Actual risk

- **Regression risk is very low.** The diff is 3 lines: one constant extraction, one argument addition, one test mock update. The test suite covers all `/query` paths and passes.
- **No behavioral change in the happy path.** When CWD = repo root, `"config.yaml"` and `"/abs/path/config.yaml"` resolve to the same file.
- **No new dependencies or architectural changes.**

## Honest take

This is defensive hardening for a scenario that's unlikely in your current deployment but real in edge cases. The consistency argument is sound (gate.py already uses `_BASE_DIR` for everything else — this was the one gap). But if you're weighing it against the principle of "don't change what works," it's a judgment call. The risk of merging is near-zero; the risk of not merging is also near-zero unless your deployment story changes.

--

# @codex, address the above feedback

PR #387, opened 2026-06-30
security(gate): add response security headers middleware and dynamic version

cgfixit:
## What\n\nAdds two hardening improvements to `gate.py`:\n\n1. **Security response headers middleware** — new `_SecurityHeadersMiddleware` sets defense-in-depth headers on every response:\n   - `X-Content-Type-Options: nosniff`\n   - `X-Frame-Options: DENY`\n   - `Referrer-Policy: strict-origin-when-cross-origin`\n   - `Permissions-Policy: camera=(), microphone=(), geolocation=()`\n   - `Cache-Control: no-store` on `/` and `/static/*` paths\n   Uses `setdefault` so per-endpoint headers are not overwritten.\n\n2. **Dynamic version from `importlib.metadata`** — replaces hardcoded `version=\"1.8.0\"` with `_CYCLAW_VERSION` read from installed package metadata, falling back to `\"dev\"` when running from source.\n\n## Why / benefit\n\n- Security headers are a low-cost, high-value defense layer recommended by OWASP. `X-Frame-Options: DENY` prevents clickjacking, `nosniff` prevents MIME-type confusion, and `no-store` prevents cached credentials.\n- Dynamic version eliminates a manual update step that drifts on every release.\n\n## Risk to monitor\n\n- If any endpoint intentionally sets different values for these headers, `setdefault` preserves them — but verify after merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_01McseYNdmxgJAaTpW1QQ9bB)_

### DIFF
--- a/gate.py
+++ b/gate.py
@@ -63,6 +63,12 @@ for k, v in _verified.items():
     status = "OK" if v == _TELEMETRY_KILL[k] else "MISSING"
     print(f"  {status}  {k}={v}")
 
+from importlib.metadata import version as _pkg_version, PackageNotFoundError
+try:
+    _CYCLAW_VERSION = _pkg_version("cyclaw")
+except PackageNotFoundError:
+    _CYCLAW_VERSION = "dev"
+
 import logging
 import yaml
 from fastapi import FastAPI, HTTPException, Depends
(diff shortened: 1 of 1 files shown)


cgfixit:
Noting in case this causes issues next manual sandbox test

PR #399, opened 2026-07-01
feat(terminal): query cancellation + health-check backoff

cgfixit:
## What

Two UX/resilience improvements to `static/terminal.html`:

1. **Query cancellation via Escape** — pressing Escape while a query is in-flight aborts it immediately via `AbortController`. The loading spinner shows "(Esc to cancel)" and the abort produces a "Query cancelled by user" message. Timeout aborts (elapsed >= deadline) still show the existing LM Studio troubleshooting message.

2. **Exponential backoff on health-check failures** — replaces the fixed 15s `setInterval` with a dynamic scheduler. On failure, the interval doubles (15s → 30s → 60s → 120s max) to avoid hammering an unreachable gateway. Resets to 15s on the first successful `/health` response.

## Why / benefit

- **Cancellation**: Previously, a stalled query forced the user to wait up to 340s (the server timeout + 10s buffer) with no way to abort. This is especially painful when LM Studio is down or overloaded — the user can now cancel and retry.
- **Backoff**: A downed gateway generated ~4 requests/minute indefinitely. With backoff, a prolonged outage generates 0.5 req/min after 2 minutes, reducing noise in both browser and server logs.

## Risk to monitor

- Escape key now prioritizes query cancellation over panel closing. If a query is in-flight, Escape cancels it; panels close only when no query is active. This is the expected UX priority but changes existing Escape behavior during queries.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YCdQLEB7f1phW3pG3YS2or)_

### DIFF
--- a/static/terminal.html
+++ b/static/terminal.html
@@ -432,6 +432,7 @@
     margin-right: 6px;
   }
   @keyframes spin { to { transform: rotate(360deg); } }
+  .cancel-hint { color: var(--text-muted); font-size: 0.85em; }
 
   /* ── EMPTY STATE ── */
   .empty-state {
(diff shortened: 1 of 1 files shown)


cgfixit:
Hahaha this previous timeout issue explains so much when I was trying to troubleshoot issues and not waiting long enough. It always just resolved itself when I walked away and came back. Those timeout and max tokens matter way more than I thought for debugging issues after making changes

PR #407, opened 2026-07-01
perf: audit log handle reuse, tunable embedding cache, shared graph error handling

cgfixit:
## Summary

Three of the eight optimization opportunities identified by a prior `CyClaw-Optimize` scan, re-verified as still valid against current `main` (e0523e3) before implementation:

- **Audit logging throughput** (`utils/logger.py`): `audit_log()` previously opened, wrote, and closed the JSONL file on every single call. Now caches one persistent append-mode file handle per resolved audit-file path and reuses it, still `flush()`ing after every write so the synchronous read-after-write visibility existing callers/tests rely on is unchanged — only the repeated open/close syscall overhead is eliminated, not the durability guarantee. Handles are released via an `atexit` hook plus a public `close_audit_handles()` for tests.
- **Embedding cache configurability** (`retrieval/embeddings.py`): the query-embedding LRU cache size was hardcoded to 2048 with no way to tune it per deployment. `functools.lru_cache` fixes `maxsize` at decoration time (before `config.yaml` is parsed), and `_cached_embedding` is a single cache shared across all config paths, so true config.yaml-driven per-path sizing would need a larger refactor. Added a `CYCLAW_EMBED_CACHE_SIZE` env var override instead (falls back to 2048), consistent with the existing `CYCLAW_DB_URL` override precedent in this codebase.
- **Graph node error-handling duplication** (`graph.py`): `local_llm_node`, `grok_fallback_node`, and `offline_best_effort_node` each independently re-derived an identical try/except → `"[<Label> Error: ...]"` + `"{code}: {message}"` block. Extracted into a shared `_generate_or_error()` helper typed against a small `_GeneratingClient` `Protocol` (since `LocalLLMClient`/`GrokClient` share the `generate(prompt) -> str` shape but no common base class).

**Deliberately not done:** merging `grok_fallback_node` and `offline_best_effort_node` into a single graph node. They call different exception-raising clients, diverge on purpose (Grok never receives the soul preamble — invariant 3 + privacy), and are independently exercised by tests as separate functions. Merging them would blur the security-relevant "Topology = Policy" invariant for a smaller duplication win than it first appears. Graph topology, node signatures, and every node's return dict are unchanged.

## Test plan

- [x] `GROK_API_KEY=dummy pytest tests/ -q --tb=short` — full suite green (only pre-existing Postgres-live skips and one pre-existing HF-download-network-blocked failure in `test_rag_integration.py`, unrelated to this diff)
- [x] `ruff check .` — clean
- [x] `mypy --strict --python-version 3.12` on all three changed files — no new errors introduced (pre-existing `dict` → `dict[str, Any]` debt untouched)
- [x] Added regression tests for the new `CYCLAW_EMBED_CACHE_SIZE` env var behavior (`tests/test_embeddings.py::TestQueryCacheSize`)
- [x] Verified `audit_log()`'s synchronous write-then-read-immediately contract still holds (existing `tests/test_audit.py` suite, including the 40-thread concurrent-writes test, passes unmodified)

---
_Generated by [Claude Code](https://claude.ai/code/session_019LMKhH4rfukpnMXuJZUovg)_

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -61,6 +61,11 @@ models:
     model: "all-MiniLM-L6-v2"
     dim: 384
     cache_dir: ".emb_cache"
+    # Query-embedding LRU cache size (retrieval/embeddings.py::_cached_embedding)
+    # is NOT set here — functools.lru_cache fixes maxsize at import time, before
+    # this file is parsed. Override the default (2048) via the
+    # CYCLAW_EMBED_CACHE_SIZE env var if a deployment needs a different
+    # memory/hit-rate trade-off.
   grok:
     enabled: false
     base_url: "https://api.x.ai/v1"
--- a/graph.py
+++ b/graph.py
@@ -41,17 +41,23 @@ CHANGES FROM ORIGINAL (soul.md / persistent personality integration):
 """
 
 import logging
-from typing import Literal, TypedDict
+from typing import Literal, Protocol, TypedDict
 
 from langgraph.graph import END, StateGraph
 
 from llm.client import GrokClient, LocalLLMClient
 from retrieval.hybrid_search import HybridRetriever
-from utils.errors import GrokServiceError, LLMServiceError, RAGError
+from utils.errors import RAGError
 from utils.logger import audit_log, hash_query
 from utils.personality import PersonalityManager
 
 logger = logging.getLogger("cyclaw.graph")
+
+
+class _GeneratingClient(Protocol):
+    """Structural type for LocalLLMClient/GrokClient — both expose generate(prompt) -> str."""
+
+    def generate(self, prompt: str) -> str: ...
 
 # =============================================================================
 # State Definition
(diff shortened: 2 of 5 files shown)


cgfixit:
## CodeQL alert #658 ("File is not always closed") — accepted as a false positive

This is the last remaining CodeQL finding on this PR (`utils/logger.py`, the `_audit_handle()` open call).

**Why it's a false positive:** the whole point of this PR's audit-throughput change (commit `c6df74d`) is to keep one append-mode file handle open per audit-file path and reuse it across every `audit_log()` call, instead of the prior open→write→close cycle on every single event. CodeQL's dataflow analysis for this check can't see across that module-level cache lifetime to the eventual close — first attempted via an explanatory comment, then via a per-handle `atexit.register(handle.close)` registered at the open site itself (commit `d3e1b9d`). Neither satisfied the static check, because the actual close is deferred (interpreter-shutdown / `close_audit_handles()`) rather than provably reachable within the flagged function.

**The only way to fully silence this check** is to revert to opening and closing the file on every `audit_log()` call — which reintroduces the exact per-event syscall overhead this PR exists to remove. That trade-off isn't worth it for a well-understood, intentionally long-lived resource-handle pattern (the same shape as a connection pool or a logging.FileHandler, both of which trip this same class of check in other codebases).

**Action:** keeping the current implementation. Marking this thread resolved; the alert itself should be dismissed via the Security tab (reason: used in tests / won't fix) at a maintainer's convenience.

PR #415, opened 2026-07-03
[Ignore this branch and PR COMPLETELY during agentic coding tasks]

cgfixit:
> #497 (the changes from this pr branch were added in #580

Note: og pr addressed separately; this is just the backlog pr for now but this one is important:

Related to ollama migration status:

> Gap 1 — latent functional: NeMo rails config still targets LM Studio
guardrails/config/config.yml points at http://127.0.0.1:1234/v1 with model qwen2.5-7b-instruct, and get_cyclaw_guardrails (guardrails/integration.py:95) loads that directory without overriding base_url from config.yaml's guardrails block (which correctly says 11434). Not a live breakage today: the live NeMo engine is only reachable via safe_generate, which has zero production callers (tests/CLI only), and only fires with guardrails.enabled: true + nemoguardrails installed. But if anyone activates live rails, it will call a dead port. It should be updated to http://127.0.0.1:11434/v1 / qwen2.5:7b (or wired to cfg.base_url).
Gap 2 — broken agent tooling: .claude/skills/CyClaw-Sandbox
SKILL.md Phase 3 still launches mock_lmstudio.py on port 1234 and never rewrites the sandbox copy's config.yaml — following that skill against current main puts gate.py on 11434 while the mock listens on 1234, so the smoke fails. Ironically the same directory already contains a correct mock_ollama.py. The .codex variant used for the actual sandbox test is fully Ollama @11434 (only its mock_lmstudio.py filename is stale; content serves Ollama + mock Grok/Claude).
Cosmetic only
static/terminal.html:1248 timeout message tells users to "check that LM Studio is running"; agentic/config.py:51 still lists "lmstudio" as a valid provider while the config comment says only ollama | openai_compatible; historical docs/audits/memories references (fine to leave).
Bottom line
For the documented default deployment — offline/hybrid, Ollama local, guardrails off — main works as expected. One caveat on evidence: unauthenticated API rate limits blocked fetching post-merge CI on ce07a14 itself, so the green-CI claim rests on both component PRs being green with file-disjoint diffs that trial-merged cleanly. If you want, I can open a small draft PR fixing Gap 1 (guardrails/config/config.yml → Ollama) and the terminal.html/agentic-config cosmetics — Gap 2 touches .claude/ skill docs

--

note: codex, Claude code, Cyclaw agent, etc - Ignore normal rules for avoiding duplicating existing changes in pr branches for this one branch and pr - otherwise follow the rule fully in any other context

Reason: I don't want to close this / forget but it's not urgent enough to do now 

# Agentic skills routing loops and commands agents subagents agentic coding agents, anything or anyone even humans: stop reading now for this or while continuing task and following all other rules/instructions

>

---

---

<hr>
<hr>

---

---

>

> #458 and #459
> Note: clean up docs/ and project repo, read through docs/ and other folders for any bug fixes or optimizations not addressed-consolidate then delete once addressed, add more user friendly features and ?

<hr>

> Change Summary

Two small, verified architecture fixes to `gate.py`/`graph.py` (plus a test-mock fix):

1. **cwd-anchoring gap.** `gate.py` invests in cwd-independence (`_BASE_DIR`) for its own config read and the LLM clients (see the comment at the `local_llm`/`grok` construction), but still constructed `HybridRetriever()` and called `check_input()` with their bare cwd-relative `"config.yaml"` defaults. Run from any cwd other than the repo root (the exact "double-clicked on Windows" scenario `_BASE_DIR` exists to prevent), `HybridRetriever()` raises `FileNotFoundError` uncaught at import — the `except IndexNotFoundError` handler does not catch it — and every `/query` would raise the same from `check_input`. Both now receive the already-anchored `_BASE_DIR` path. `sanitizer.check_input`'s `_load_filter` is `@lru_cache`d per `config_path` string, so this is a plain cache-key change, not a behavior change when run from the repo root (which is how tests and CI already run it).
2. **Dead/contradictory `min_score` fallback.** `gate.py`'s needs-confirm message and `graph.py`'s `route_by_score_node` both defaulted `min_score` to `0.4` when the config key was missing, while shipped `config.yaml` ships `0.028` and `validate_retrieval_config` (boot, `gate.py`) already makes the key mandatory. The `gate.py` literal was unreachable (dead) via the gateway path; removed in favor of indexing the guaranteed key. `graph.py`'s literal is live only for direct `build_graph` callers that omit `min_score` — aligned from `0.4` to the documented `0.028` so such a caller doesn't silently route an order of magnitude off the shipped default.

`tests/test_gate.py`'s `check_input` mock lambda arity fixed (`lambda q, **kw: q`) to accept the new `config_path` kwarg.

## Why / benefit

Removes a real crash mode for non-repo-root launches and a misleading dead/contradictory magic number. No security invariant touched; routing topology unchanged.

## Verification

- `GROK_API_KEY=dummy python -m pytest tests/test_gate.py tests/test_graph.py tests/test_hybrid_search.py tests/test_sanitizer.py tests/test_config_validation.py -q --tb=short` → all passed.
- `ruff check gate.py graph.py tests/test_gate.py` → clean.
- Self-audited against the ponytail seven rules (YAGNI / stdlib-first / minimal abstraction / no dead code / no speculative generality / correctness over cleverness / no half-measures) — PASS, no violations; this diff *removes* a dead literal rather than adding one.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -57,11 +57,18 @@ HTTP POST /query  (or MCP tool call)
 | `utils/ratelimit.py` | Thread-safe per-IP rate limiting (60 req/min) |
 | `utils/health.py` | `check_all()` / `_ping()` backing `/health`; skips Grok probe when `GROK_API_KEY` is unset |
 | `utils/errors.py` | Typed exception hierarchy rooted at `RAGError`; all modules raise these, never bare `Exception` |
+| `utils/config_validation.py` | Boot-time config validation (`validate_retrieval_config`, `validate_personality_config`); fails fast on out-of-range tunables |
+| `utils/ops_runner.py` | Subprocess shim backing the four `/ops/*` endpoints |
+| `retrieval/vector_store.py` | Pluggable vector reader (`get_vector_reader`): embedded ChromaDB or pgvector backend |
+| `retrieval/clear_cache.py` | Dry-run-by-default cleaner for the embedding cache (`models.embeddings.cache_dir`) |
 | `schemas/api.py` | Pydantic models: `QueryRequest`, `QueryResponse`, `HealthResponse` |
 | `metrics.py` | `audit.jsonl` analyzer |
 | `mcp_hybrid_server.py` | MCP server (retrieval only, no LLM, no `sampling`) |
 | `sync/` | Out-of-band Dropbox corpus sync via `rclone`; run as `python -m sync.cli`; never imported by gate/graph |
 | `agentic/` | Out-of-band GitHub context + governed skills registry; run as `python -m agentic.cli`; **never imported by gate/graph/mcp** |
+| `agentic/fsconnect/` | Out-of-band local/SMB filesystem connector; read-only by default, writes require `writes_enabled` + reason + confirm |
+| `agentic/sqlconnect/` | Out-of-band SQL connector; DSN from `CYCLAW_SQL_DSN` env only, SELECT/WITH-only guard |
+| `guardrails/` | Optional NeMo Guardrails integration; soft-imported, disabled by default |
 
 ### Configuration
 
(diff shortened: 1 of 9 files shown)


cgfixit:
Ledger update (2026-07-23 Phase-1 optimize scan, verified against main `b287558`) — three of the four Ollama-migration items tracked here are **already resolved on main**, and the fourth is in-flight:

| Item | Status | Evidence |
|---|---|---|
| Gap 1 — NeMo rails config targets LM Studio `:1234` | ✅ **Resolved** | `guardrails/config/config.yml:28,40` now `http://127.0.0.1:11434/v1` + `qwen2.5:7b`, matching `config.yaml`'s guardrails block; `guardrails/integration.py:91-93` also overrides `params["base_url"]` from `GuardrailsConfig` at engine build |
| Gap 2 — Sandbox SKILL.md launches `mock_lmstudio.py` on 1234 | ✅ **Resolved** | SKILL.md Phase 3 is now "Build Mock Corpus & Index" — no mock-LLM launch step remains; the active mock is `mock_ollama.py` |
| Cosmetic — `terminal.html:1248` "check that LM Studio is running" | ✅ **Resolved** | Line now reads "check that **Ollama** is running…"; no LM Studio reference remains in the file |
| Cosmetic — `agentic/config.py:51` lists `lmstudio` provider | ✅ **Resolved** | `_VALID_DEEPAGENT_PROVIDERS = ("ollama", "openai_compatible")` at line 54; comment matches |
| Leftover — orphaned `mock_lmstudio.py` file itself | 🔄 **#652** (draft) | Repo-wide grep: no active references; deletion proposed for your review |

Related, same scan: the fallback story this PR's notes circle around is closed by **#651** — the harness console was the one surface not routed through `resolve_local_backend()`, so with `fallback.enabled: true` and Ollama down it missed the LM Studio failover that `/query`/`/health` already perform.

Leaving this PR untouched per its own instructions.

---

cgfixit:
Note for later: notes to self

PR #471, opened 2026-07-09
fix: align /ops/sync timeout with config sync_timeout_sec

cgfixit:
## What
`POST /ops/sync` (console-driven via `utils/ops_runner.py`) hard-killed the `sync.cli` subprocess at a fixed 120s while `config.yaml` allows `sync.sync_timeout_sec: 3600` for rclone. Full `sync` actions now use the config ceiling (+60s overhead for Python/bookkeeping). Short actions (`status`, `test`, `schedule`, `unschedule`) keep the 120s default.

## Why / benefit
A real Dropbox corpus pull from the Soul Console was unusable: the gateway aborted mid-transfer and could leave a partial `data/corpus/` while the CLI path succeeded. Aligns ops with the documented sync budget.

## Risk to monitor
Ops workers can now hold a subprocess for up to ~1h on a hung rclone. The single-instance sync lock and rclone's own timeout still apply. If 3600 is too long for the gateway process model, lower `sync.sync_timeout_sec` rather than re-hardcoding 120.

## Verify
`GROK_API_KEY=dummy pytest tests/test_ops_runner.py -q` — all passed.

## Scan context
CyClaw-Optimize pass (ponytail + Karpathy + python-coding-agent). Highest-leverage reliability finding from the 10-minute origin/main scan.

### DIFF
--- a/tests/test_ops_runner.py
+++ b/tests/test_ops_runner.py
@@ -21,7 +21,10 @@ def _fake_run(returncode: int = 0, stdout: str = "", stderr: str = ""):
     """Return a (_run replacement, captured-argv list) pair."""
     captured: list[list[str]] = []
 
-    def _runner(argv: list[str]) -> subprocess.CompletedProcess[str]:
+    def _runner(argv: list[str], *, timeout_sec: int | None = None) -> subprocess.CompletedProcess[str]:
+        # timeout_sec is accepted so run_sync_op's config-aligned budget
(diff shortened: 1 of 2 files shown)


cgfixit:
@claude and @codex review pr changes and provide feedback on if this is accurate and or should be merged (I'll test myself later either way but just wanting to check)

PR #550, opened 2026-07-17
docs: status banners on three stale fix-lists (verified against main @ 8dc96d5)

cgfixit:
## What

Adds a verification-dated **STATUS BANNER** to the top of three standing task lists that optimization scans re-read first:

1. **`docs/todo/CheckIfStillNeedFix-7.13.md`** — every item verified RESOLVED on main @ 8dc96d5 (rate limiter extracted to `utils/ratelimit.py` with real-limiter tests under a fake clock; 429/401/503/504/oversize HTTP coverage in `test_gate.py`; personality `record_interaction` + retrieval RAGError tests in `test_graph.py`; `test_personality_changes.py` gone; `test_audit.py` header rewritten; soul-auth tests in `test_security.py`).
2. **`docs/zIdeas/CHECKIFSTILLRELEVANTponytail-fix-planning.md`** — all 8 violations verified RESOLVED on main (each site grep/read-verified).
3. **`.claude/CyClaw-Optimizer-Readme.md`** — most findings resolved; the handful still open are explicitly mapped to the draft PRs now covering them (#547 tests-hygiene, #548 audit-key normalization, #549 Docker uv digest + torch pin). Accepted-design items (SEC-5, SEC-8, `include_query_hash=false` semantics pinned by `test_due_diligence_invariants.py`) called out so future scans don't re-litigate them.

## Why / benefit

Auditability / scan efficiency: these files are seeded into every optimization scan's first read block. Without a banner, each scan re-investigates already-landed work — this session re-verified all three lists item by item, and the banners record the outcome so the next scan starts from current truth instead of June state.

## Verification

Docs-only change. Each "resolved" claim was verified against main @ 8dc96d5 by direct grep/read of the cited code sites, not assumed.

## Risk to monitor

None — documentation only. The banners will themselves go stale as PRs #547–#549 merge/close; they point at the PR numbers so a future pass can update one line each.

### DIFF
--- a/.claude/CyClaw-Optimizer-Readme.md
+++ b/.claude/CyClaw-Optimizer-Readme.md
@@ -1,3 +1,16 @@
+> **STATUS BANNER (verified 2026-07-18 against main @ 8dc96d5):** nearly all
+> findings below are RESOLVED on current main — H1–H5, M1, M3–M8, SEC-1/2/3/7,
+> I1, L6, and the ponytail items (see docs/zIdeas/CHECKIFSTILLRELEVANTponytail-fix-planning.md,
+> all 8 landed) no longer reproduce. STILL OPEN at verification time (each now
+> has a draft PR): M9 (MCP audit "mode" key → PR #548), M10 (agentic-context
+> mock shape → PR #547), M5 (dead gate patches → PR #547), L7 (module-level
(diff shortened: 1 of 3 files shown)


cgfixit:
Reminder to self to remove these stale docs as they're verified stale

PR #559, opened 2026-07-19
test: filter known StarletteDeprecationWarning + refresh its tech note

cgfixit:
## What

Two coupled changes:

1. **`pyproject.toml`** — adds an exact class+message `filterwarnings` entry to `[tool.pytest.ini_options]`:
   ```toml
   'ignore:Using `httpx` with `starlette\.testclient` is deprecated; install `httpx2` instead\.:starlette.exceptions.StarletteDeprecationWarning'
   ```
   The warning fires on **current** pins (verified empirically: starlette 1.3.1 / httpx 0.28.1 / fastapi 0.138.0) once per pytest session. The repo's own tech note (`docs/TESTCLIENT_HTTPX_DEPRECATION.md`, filed 2026-06-19) prescribed exactly this as its short-term **Option 2** — it was never applied. Scoped narrowly (full message + exact warning class) so no other warning is masked. The `httpx2` migration stays owed to the next Starlette major, per the note's recommendation.

2. **`docs/TESTCLIENT_HTTPX_DEPRECATION.md`** — refreshes the now-stale note: status → filter applied (2026-07-19); `fastapi==0.137.2` → `0.138.0` in both pin cites; "all 98 tests pass" → suite-size-agnostic wording (suite is 1319 tests now); "LM Studio / Grok" → "Ollama / Grok / Claude"; Option 2 snippet replaced with the exact filter that landed; TestClient consumer list gains `tests/test_gate_ops.py` (#557); reproduction updated (torch 2.4.1+cpu → 2.13.0+cpu, constraints-aware install, `-W` override since the filter now suppresses by default — override verified to reproduce).

## Why / benefit

Every CI and local test run carries one known-noise warning line. Silencing it with a *narrow* filter keeps the warning stream signal-dense (a new warning now stands out), while the tech note + the in-TOML comment pointing at it keep the deprecation tracked instead of forgotten.

## Verification

- Warning occurrences in test output: **1 → 0** (`pytest tests/test_gate.py tests/test_security.py`)
- `-W "always::starlette.exceptions.StarletteDeprecationWarning"` **still reproduces** the warning (the filter is an ignore, not a fix — by design)
- Full suite: **1319 tests, 0 failures, 0 errors, 14 skipped**
- `tomllib` parse clean; `doc-sync` 0 drift; `dep-guard` clean
- Ponytail/karpathy self-audit: PASS — the filter is the tech note's own documented recommendation, not new scope

## Risk to monitor

The filter masks only the exact string; if Starlette rewords the message, the warning resurfaces (loud, not silent — the failure mode we want). The underlying `httpx2` migration is still pending and tracked in the tech note.

### DIFF
--- a/docs/TESTCLIENT_HTTPX_DEPRECATION.md
+++ b/docs/TESTCLIENT_HTTPX_DEPRECATION.md
@@ -11,7 +11,7 @@
 Every test run that constructs a `TestClient` emits this warning:
 
 ```
-.../site-packages/fastapi/testclient.py:1: StarletteDeprecationWarning:
+.../site-packages/fastapi.testclient.py:1: StarletteDeprecationWarning:
   Using `httpx` with `starlette.testclient` is deprecated; install `httpx2` instead.
     from starlette.testclient import TestClient as TestClient  # noqa
 ```
@@ -19,7 +19,7 @@ Every test run that constructs a `TestClient` emits this warning:
 It is raised once per session (at import of `fastapi.testclient`, which re-exports
 `starlette.testclient`). It does **not** fail any test, but it is noise on every
 `pytest` invocation, including CI. As of 2026-07-19 it is suppressed by an exact
-class+message `filterwarnings` entry in `pyproject.toml` (see "Recommendation"),
+message-text `filterwarnings` entry in `pyproject.toml` (see "Recommendation"),
 so it no longer appears in test output; the underlying deprecation is unchanged.
 
 ### Where the TestClient is used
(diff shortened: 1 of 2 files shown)


cgfixit:
## CI fix pushed — root cause of the `ci (3.12)` failure

The failing lane was the **conda** workflow (`python-package-conda.yml`), not `ci.yml`. Root cause:

- The conda lane installs `fastapi=0.115.9` → **starlette 0.4x** (conda-forge's chromadb build hard-pins it — documented in `environment.yml`).
- My original filter was class-qualified (`:starlette.exceptions.StarletteDeprecationWarning`), and pytest resolves filter categories at startup. On starlette 0.4x that class doesn't exist → `AttributeError`, lane red before any test ran.

**Fix (commits `70378dfa` + `0909b188`):** filter is now scoped by the exact message text only — still unique to this warning, masks nothing else — and parses in both lanes. Tech note updated with the caveat.

Verified in two disposable envs emulating both lanes:

| Env | Filter | Result |
|---|---|---|
| starlette 0.45.3 (conda-lane analog) | class-qualified | `AttributeError` at startup — exact CI repro |
| starlette 0.45.3 | message-only | pytest starts clean, tests pass |
| starlette 1.3.1 (pip lanes) | message-only | warning suppressed (1 → 0), tests pass |
| starlette 1.3.1 | no filter | warning fires (sanity control) |

Lesson noted for future runs: **any pytest `filterwarnings` entry in this repo must be class-path-free** (or guard against the conda lane's older starlette).

cgfixit:
Follow-up fix (commits `aecabfbc` → `5a2ed394`): the `Lint (Ruff + WPS)` failure was my own transcription error — the hand-pushed `pyproject.toml` had dropped `"E402", "I001", "B008"` from gate.py's ruff `per-file-ignores` (plus graph.py's `E501` entry), so the repo-wide ruff gate flagged 34 pre-existing, deliberately-waived findings. Line restored byte-exact to main; verified locally: `ruff check --select E,F,I,B,C4,UP,S .` → **All checks passed**, tomllib parse clean, and the diff vs main is now *only* the `filterwarnings` block as intended. (Also caught and fixed a malformed quote in my first restoration commit before it could break TOML parsing — final pushed content re-fetched and diffed against the verified local copy.)

PR #582, opened 2026-07-20
fix(sandbox-test): track shipped grok-4.5 model in mock provider smoke

cgfixit:
## What

The `cyclaw-sandbox-test` harness (`.codex/skills/cyclaw-sandbox-test/`) broke when PR #570 bumped the shipped `config.yaml` to `grok-4.5`:

- `mock_lmstudio.py` dispatches the `"[Mock Grok API]"` marker response only on an **exact** model-id match against its hardcoded `GROK_MODEL_ID = "grok-4.3"`.
- `_enable_mock_providers` patched `enabled`/`base_url` but left the shipped `model: grok-4.5` in place, so the client smoke's `assert "Mock Grok API" in grok_answer` could never pass again — every sandbox run reports a false FAIL.

Changes:
1. Bump `GROK_MODEL_ID` to `grok-4.5` in both sandbox scripts (matches shipped config, current per #570).
2. `_enable_mock_providers` now also pins the patched config's grok/claude `model` to the mock constants, so the **next** shipped-model bump cannot silently re-break the dispatch — the harness no longer depends on the shipped default staying in sync by hand.

## Why / benefit

Restores a broken verification harness (false FAIL on every sandbox run) and removes the recurrence class: config-model bumps are a routine maintenance event and shouldn't require a matching edit in two `.codex` scripts to keep the smoke green.

## Verification

- `python3 -c "import ast; ..."` parses both edited scripts; `ruff check .codex/skills/cyclaw-sandbox-test/scripts/` clean (E,F,I,B,C4,UP,S).
- Logic re-checked against `llm/client.py` (`GrokClient` sends `cfg["models"]["grok"]["model"]` verbatim) and the mock's `_answer` dispatch.
- Full sandbox run not executed here (needs torch + index build); the mock/config interplay was verified statically and by the byte-exactness check below.
- Both pushed blobs verified byte-identical to the locally edited files (`git hash-object` compare: `04651af8`, `e63f4876`).

## Risk to monitor

- None to runtime code — only `.codex` test tooling is touched; no production path, no invariant affected.
- Branch history has 3 commits (an initial push dropped `main()`'s tail and was corrected in the follow-ups; net diff is exactly the two-file change). Recommend **squash-merge**; a force-push cleanup was not done because that needs explicit sign-off per CLAUDE.md.
- If xAI retires `grok-4.5` later, the same one-line constant bump applies — but the config-pin means the smoke keeps passing regardless.

### DIFF
--- a/.claude/commands/CyClaw-Optimize.md
+++ b/.claude/commands/CyClaw-Optimize.md
@@ -1,19 +1,285 @@
 ---
-description: Methodically scan the CyClaw main branch for code, CI, security, financial-risk, and maintainability optimization opportunities, then open focused, reviewable pull requests for each.
+name: CyClaw-Optimize
+description: Methodically scan the CyClaw main branch for code, CI, security, financial-risk, and maintainability optimization opportunities, then open focused, reviewable pull requests for each. Use when asked to optimize CyClaw, find competitive/trade-bot advantages, harden CI, audit for risk, propose improvements, or open optimization PRs against main.
 ---
 
-Run a full optimization sweep of CyClaw main and open one draft PR per finding. $ARGUMENTS
+# CyClaw-Optimize
 
-## Steps
+**Persona:** You are a modern AI engineer specializing in Python and extremely
+familiar with the CyClaw architecture — FastAPI RAG gateway (`gate.py`),
+LangGraph 7-node security topology (`graph.py`), ChromaDB + BM25 hybrid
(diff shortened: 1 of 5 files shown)


cgfixit:
Note the sandbox mock lm studio needs to be updated for ollama 

update: done but commands is where it should be

also time for local verification... really overdue ha

PR #584, opened 2026-07-20
fix(health): close the shared /health probe client on server shutdown

cgfixit:
## What

`utils/health.py` keeps a module-level pooled `httpx.Client` for `/health` probes (deliberate — per-request client construction was the endpoint's whole latency budget). But `gate.py`'s lifespan shutdown closes every *other* long-lived pool (`local_llm`, `grok`, `claude`, `_rate_limiter`, `personality`, `retriever`) and had no handle on this one, so a server restart leaked the probe client's connections until process exit / GC.

- `utils/health.py`: new `close_http_client()` — idempotent, lock-guarded, safe when no client exists; the next probe lazily rebuilds (post-restart path).
- `gate.py`: call it from the lifespan shutdown path, in its own `try/except` like the other closes so one failure can't skip the rest.
- `tests/test_health.py`: two tests — close-without-client is a no-op; close releases the client and a subsequent probe lazily builds a fresh one.

## Why / benefit

Resource hygiene on the restart path; matches the teardown contract every other pool in the lifespan already honors. No behavior change to any serving path.

## Verification

- `GROK_API_KEY=dummy pytest tests/test_health.py -q` → **20/20 pass** (run with `--noconftest` — sandbox deps were still installing; conftest needs the full stack. CI runs the full suite including `test_gate.py`).
- `ruff check gate.py utils/health.py tests/test_health.py` clean (E,F,I,B,C4,UP,S).
- `python3 .claude/skills/invariant-guard/check_invariants.py` → 28/28 pass (gate.py was touched).
- All three pushed blobs verified byte-identical to the local files (`git hash-object` compare).

## Risk to monitor

- The gate.py edit is purely additive (one import + one isolated shutdown call); no request-path code changes. If `close_http_client` itself raised, it's caught and logged like the sibling closes.
- Tests that mutate `health._http_client` directly already reset it in fixtures; the new tests restore `None` in a `finally`.

### DIFF
--- a/gate.py
+++ b/gate.py
@@ -88,7 +88,7 @@ from utils.errors import (
     PromptInjectionError, IndexNotFoundError
 )
 from utils.guardrail_bridge import build_input_guard
-from utils.health import check_all
+from utils.health import check_all, close_http_client
 from utils.personality import PersonalityManager
 from gate_ops import register_ops_routes
 from metrics import summarize_audit
@@ -305,6 +305,12 @@ async def lifespan(app: FastAPI):
                 _obj.close()
             except Exception:
                 logger.warning("shutdown close failed for %s", _name, exc_info=True)
+    # The /health probe pool is module-level in utils.health (no instance to
+    # list above); close it through its own teardown hook for the same reason.
+    try:
+        close_http_client()
+    except Exception:
+        logger.warning("shutdown close failed for health http client", exc_info=True)
 
 
 app = FastAPI(
--- a/tests/test_health.py
+++ b/tests/test_health.py
@@ -320,3 +320,40 @@ class TestHealthCfgCache:
         # Different object identity -> the file was re-read and re-parsed.
         assert first is not second
         assert second == first  # same content, freshly parsed
+
+
+class TestSharedClientLifecycle:
+    """The pooled _http_client must be closable via close_http_client()."""
+
+    def test_close_without_client_is_noop(self):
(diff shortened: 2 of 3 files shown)


cgfixit:
need to understand this better before merging

cgfixit:
Seems safe to merge:

gate.py‘s FastAPI lifespan shutdown already closes every long-lived connection pool it created — the LLM clients, rate limiter, personality manager, retriever — each in its own try/except. utils/health.py has always had one more: a module-level httpx.Client used only by the /health probe (_http_get()), guarded by _http_client_lock. That client was never in the shutdown list, so on every server restart its connections leaked until GC/process exit — harmless in practice (it’s freed anyway), but sloppy.

This PR just gives that module-level client a teardown hook and wires it into the same loop, in the same isolated-try pattern as the other six:

	1.	close_http_client() takes the existing lock, swaps _http_client to None, and closes the old client outside the lock — reusing the exact locking discipline _http_get() already uses, so there’s no new race.
	2.	It’s idempotent (if client is not None) — safe to call on a server that never made a health probe.
	3.	The next probe after shutdown/restart lazily rebuilds a fresh client, same as today.

I read the full utils/health.py and gate.py at the PR’s head SHA, not just the diff — the lock reuse and shutdown wiring are both correct. Two new tests cover no-op-when-absent and close-then-lazy-rebuild. Nothing here touches a graph edge, retrieval, or a security invariant

PR #585, opened 2026-07-20
fix(sync): scale stale-lock threshold to the configured sync timeout

cgfixit:
## What

The single-instance sync lock (`sync.lock.d`) is reclaimed as "stale" after a flat 3h (`_LOCK_STALE_SEC`). But a *bounded* run can legitimately hold the lock for `sync_timeout_sec` — the retry loop treats it as the global wall-clock budget across all attempts. So with `sync_timeout_sec` configured above 3h (or `0` = unbounded), a **live** long run's lock could be reclaimed and a second sync started against the same remote concurrently — exactly the corruption (interleaved logs, racing filter-file writes) the lock exists to prevent.

- `_lock_stale_after_sec(cfg)`: bounded runs get `max(3h, sync_timeout_sec + 1h margin)`; the threshold never shrinks below the existing default, so short runs are unaffected.
- `_acquire_sync_lock(lock_dir, stale_after_sec=...)`: threshold is now a parameter; the default is byte-identical to the old behavior.
- `run_sync` passes the derived threshold, and logs a warning when `sync_timeout_sec == 0`: no finite threshold can cover an unbounded run, so that configuration keeps the flat 3h with **degraded** mutual-exclusion — the honest residual risk, now visible in the log instead of silent.
- 4 new unit tests pin the derivation (tracks bounded timeout / floored at default / unbounded keeps default) and the threshold-parameterized reclaim-vs-refuse behavior.

## Why / benefit

Robustness of the financial/oversight-relevant sync path: a concurrent double-sync against Dropbox is the kind of failure the lock was added to prevent, and the current flat threshold quietly re-opens it for any operator who raises the timeout for a large corpus.

## Verification

- `GROK_API_KEY=dummy pytest tests/test_sync_runner.py -q` (self-contained, `--noconftest`) → **all pass** (existing 60+ plus the 4 new).
- `ruff check sync/runner.py tests/test_sync_runner.py` clean.
- `python3 .claude/skills/invariant-guard/check_invariants.py` → 28/28 pass.
- Both pushed blobs verified byte-identical to local (`git hash-object`: `c40315f4`, `ff76c580`).

## Risk to monitor

- `sync_timeout_sec == 0` operators still have the residual exposure (documented in the new run-start warning); the alternative — refusing 0 — would break the documented "unbounded" escape hatch, so it was not done.
- Default-config behavior (timeout ≤ 3h) is unchanged: threshold stays exactly 3h.

### DIFF
--- a/sync/runner.py
+++ b/sync/runner.py
@@ -605,14 +605,22 @@ _LOCK_STALE_MARGIN_SEC = 60 * 60
 def _lock_stale_after_sec(cfg: RcloneConfig) -> float:
     """Stale-lock threshold for a run configured by ``cfg``.
 
-    The retry loop bounds total run time by ``sync_timeout_sec`` (global
+    The retry loop bounds total sync time by ``sync_timeout_sec`` (global
     deadline), so a bounded run can legitimately hold the lock for that long --
     the threshold must exceed the budget or a *live* long run looks "stale" and
-    a second sync starts underneath it. Unbounded (0) keeps the flat 3h default;
-    run_sync warns about the degraded protection in that case.
+    a second sync starts underneath it. When ``post_sync_check`` is enabled the
+    lock is ALSO held through ``run_post_sync_check``, which independently
+    receives the full ``sync_timeout_sec`` as its own subprocess timeout, so
+    the complete lock-held budget is ~2x the sync budget (review finding on
+    PR #585). Unbounded (0) keeps the flat 3h default; run_sync warns about
+    the degraded protection in that case.
     """
     if cfg.sync_timeout_sec > 0:
-        return max(_LOCK_STALE_SEC, cfg.sync_timeout_sec + _LOCK_STALE_MARGIN_SEC)
+        multiplier = 2 if getattr(cfg, "post_sync_check", False) else 1
+        return max(
+            _LOCK_STALE_SEC,
+            cfg.sync_timeout_sec * multiplier + _LOCK_STALE_MARGIN_SEC,
+        )
     return _LOCK_STALE_SEC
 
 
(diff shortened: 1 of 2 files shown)


cgfixit:
@codex What're your thoughts on this - seems low risk and logical, but just getting another opinion

cgfixit:
Note to self: (no action yet) codex’s finding is correct; post_sync_check=True can hold the sync lock close to 2× sync_timeout_sec, not the +1h margin the PR assumes. Needs kimi’s multiplier fix first.

cgfixit:
Review finding verified against the code and fixed — the codex comment was correct.

**Confirmation of the flaw:** `run_post_sync_check` (`sync/runner.py`) receives the full `cfg.sync_timeout_sec` as its own subprocess timeout, and it runs inside `_run_sync_locked` — i.e. **while the single-instance lock is held**, after the sync retry loop has already consumed up to the full `sync_timeout_sec` global budget. Worst-case lock-hold with `post_sync_check=True` is therefore ~`2 × sync_timeout_sec` + overhead, and the original `+1h` margin under-covered it (codex's 4h → ~9h example checks out).

**Fix (commit `205256d`):**
- `_lock_stale_after_sec`: threshold = `max(3h, sync_timeout_sec × (2 if post_sync_check else 1) + 1h margin)`. Default config unchanged — the 3h floor still dominates; only large timeouts with post-sync checking move.
- `run_sync` docstring now states the doubled budget explicitly.
- 3 new regression tests: the 2× derivation, floor-dominance with the multiplier, and an end-to-end acquisition spy proving `run_sync` acquires the lock with the doubled threshold when `post_sync_check=True` (the exact lifecycle from the finding).

**Verification:** 68/68 tests in `test_sync_runner.py` pass locally (`--noconftest`), ruff clean (E,F,I,B,C4,UP,S), invariant-guard 28/28, and both pushed blobs verified byte-identical to local via `git hash-object`.

PR #590, opened 2026-07-21
fix(guardrails): close live-rails activation blockers (codex findings)

cgfixit:
## What
Addresses the guardrails findings from `docs/codex-findings-7202026.md` (P1 live-NeMo activation blockers + two P2 contract gaps), each verified against current main before editing — plus the review-round fixes below.

- **Authoritative config**: new `_apply_guardrails_config()` makes `config.yaml`'s `guardrails:` block (engine/model/base_url) authoritative over the static NeMo template at engine-build time. The template itself is migrated LM Studio `@1234` → Ollama `@11434` (`qwen2.5:7b`), pinned by a drift test. Previously, enabling live rails called the retired LM Studio endpoint.
- **Context transport**: retrieved context now reaches NeMo as a `{"role": "context", "content": {"relevant_chunks": ...}}` message — the only channel the grounding action actually reads in nemoguardrails 0.23.0 — followed by the user message. (See review-round note: an earlier revision of this branch passed a `context=` kwarg that the real engine does not accept.)
- **Degradation contract**: `safe_generate` now degrades on live-provider `generate_async` failures (connect/timeout/5xx/Colang runtime) too, redacted to the exception type name — it previously propagated, contradicting the documented degrade-never-crash contract.
- **CLI honesty**: `guardrails.cli check` no longer advertises itself as offline when it can perform a live generation.

## Review-round fixes (2026-07-21)
- **P1 — `generate_async(context=...)` would raise `TypeError` on the real engine.** Verified against the nemoguardrails 0.23.0 wheel: `generate_async(self, prompt=None, messages=None, options=None, state=None, streaming_handler=None)` — no `context` kwarg, and system messages are explicitly unsupported. The live path now sends the context-role message described above. The fake engines in tests were given the REAL signature (a caller passing `context=` now raises `TypeError` exactly as the real engine would), the transport test asserts `set(call) == {"messages"}` with the context-role + user messages, and a new signature-pin test locks the fake to the real `LLMRails.generate_async` contract.
- **P2 — config.yml claimed authority it doesn't have.** The header comment claimed this file defines the active rail set, while `config.yaml`'s `guardrails:` rail knobs (`input_rails`/`output_rails`/`topical_rails`/`hallucination_threshold`, carried by `GuardrailsConfig`) are NOT yet applied to the live engine (offline floor only). The comment now states precisely what each file governs.

## Why / benefit
The live guardrails path was broken-if-activated: wrong endpoint, unreachable grounding evidence, crash-on-provider-error. Default deployments (guardrails off) are untouched; this makes the opt-in layer actually activatable.

## Verification
- 24/24 guardrails integration tests pass (new: no-context call shape, fake/real signature pin); ruff `--select E,F,I,B,C4,UP,S` clean; invariant-guard 28/28.
- All pushed blobs verified byte-identical via SHA-1 compare (integration.py `9e5baebf`, test file `29164d17`, config.yml `2cef2111`).
- Caveat: a live NeMo/Colang run was NOT executed (dependency absent) — covered by real-signature fake-engine unit tests. Recommend a manual live smoke before enabling.

## Risk to monitor
- The context-role message transport matches the nemoguardrails 0.23.0 documented pattern but is not exercised against a live engine here.
- Multi-commit branch — squash-merge suggested.

Refs: `docs/codex-findings-7202026.md` (P1 guardrails, P2 safe_generate/cli).

### DIFF
--- a/guardrails/cli.py
+++ b/guardrails/cli.py
@@ -3,7 +3,10 @@
 Subcommands:
 
     status   Print guardrails config + NeMo-config presence + dep availability.
-    check    Run the offline rails over a query string (no LLM, no NeMo needed).
+    check    Probe the guardrails path over a query string. With guardrails
+             enabled AND nemoguardrails installed this performs a LIVE
+             generation through the configured local LLM; otherwise it runs
+             the offline heuristic floor (no LLM, no NeMo needed).
     metrics  Summarize the guardrail metrics stream (logs/guardrails.jsonl).
     test     Run the pre-flight self-test.
 
@@ -125,7 +128,11 @@ def build_parser() -> argparse.ArgumentParser:
     p_status = sub.add_parser("status", help="Print guardrails config + dependency status.")
     p_status.set_defaults(func=cmd_status)
 
-    p_check = sub.add_parser("check", help="Run offline rails over a query (no LLM/NeMo needed).")
+    p_check = sub.add_parser(
+        "check",
+        help="Probe the guardrails path over a query. Live generation when guardrails "
+        "+ nemoguardrails are enabled; offline heuristic floor otherwise.",
+    )
     p_check.add_argument("query", help="The user query to evaluate.")
     p_check.add_argument("--context", help="Optional retrieved-context string to ground against.")
     p_check.set_defaults(func=cmd_check)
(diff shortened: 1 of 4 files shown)


cgfixit:
@dependabot rebase

PR #591, opened 2026-07-21
fix(fsconnect): prune staged copies of source-deleted files (codex finding)

cgfixit:
## What
Addresses the fsconnect indexer finding from `docs/codex-findings-7202026.md` (P1), verified against current main before editing — plus the review-round fixes below.

`FsIndexer.apply()` staged every current source file but **never removed a staged file whose source disappeared**. The retrieval indexer ingests the whole staged tree, so content deleted from the share stayed retrievable indefinitely.

`apply()` now prunes the staging tree, bounded by an ownership record:
- `_prune_staging()` removes staged files that (a) CyClaw itself recorded staging on a prior run (the manifest cache) AND (b) are absent from the current eligible manifest. Pruned count is reported in the `fsconnect_index_apply` audit event and the result dict.
- The manifest cache is loaded and saved in ALL modes — it is the ownership record for pruning, not merely the incremental skip-cache.

## Review-round fixes (2026-07-21)
- **P1 — blast radius.** The first version of this branch pruned anything in the staging dir absent from the manifest — including operator files that predated CyClaw (reachable in practice via the `--staging` CLI override). Pruning is now bounded to the prior-run ownership record: unrelated files are never deleted, and a first run with no cache prunes nothing.
- **P1 — Windows case-only renames.** Prune comparison keys are `os.path.normcase`-normalized, so `Notes.md` → `notes.md` is not misclassified as stale-and-deleted on Windows.
- **P2 — skipped sources left stale copies.** The current set is now eligible-only: a source that grows past `index_max_file_bytes` (too_large) or starts failing reads (read_error) has its staged copy pruned rather than served stale. Transient read errors self-heal on the next successful run.
- **P2 — cache-name exemption over-reached.** The old name-based exemption also shielded NESTED files named `.fsindex_cache.json`. The root cache is now protected structurally (it is never a manifest path), nested same-name files prune normally, and a share-root source file with that exact name is quarantined as `reserved_name` instead of clobbering the ownership record.

## Why / benefit
Deletion must propagate to the retrievable corpus — otherwise fsconnect silently resurrects content the operator deleted (staleness + data-governance issue) — without ever deleting files CyClaw does not own.

## Verification
- 104/104 fsconnect tests pass (7 new: first-run safety, unrelated-file survival, too_large/read_error pruning, case-only rename, nested cache-name pruning, reserved_name quarantine); ruff `--select E,F,I,B,C4,UP,S` clean; invariant-guard 28/28.
- All pushed blobs verified byte-identical via SHA-1 compare (indexer.py `e680f934`, test file `a6ecb7fd`).

## Risk to monitor
- Pruning is staging-scoped, ownership-bounded, and runs after the manifest is fully built, so a failed walk can't trigger deletion; worst case of a spurious prune is a re-stage on the next run (self-healing).
- Multi-commit branch — squash-merge suggested.

Refs: `docs/codex-findings-7202026.md` (P1 fsconnect indexer).

### DIFF
--- a/agentic/fsconnect/indexer.py
+++ b/agentic/fsconnect/indexer.py
@@ -75,6 +75,15 @@ class FsIndexer:
                 ext = os.path.splitext(name)[1].lower()
                 if ext not in self.fs_cfg.index_extensions:
                     continue
+                if child == _CACHE_NAME:
+                    # A share-root file whose staged path would BE the skip-
+                    # cache's own path must never be staged: it would clobber
+                    # the cache that ownership-bounded pruning relies on.
+                    # Only reachable when an operator adds ".json" to
+                    # index_extensions (defaults exclude it).
+                    manifest.append({"path": child, "size": entry["size"],
+                                     "skipped": "reserved_name"})
+                    continue
                 if entry["size"] > self.fs_cfg.index_max_file_bytes:
                     manifest.append({"path": child, "size": entry["size"], "skipped": "too_large"})
                     continue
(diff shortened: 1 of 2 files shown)


cgfixit:
Note to myself - no action required from agents: This is all sorts of jacked up (Kimi seemed to hit a limit and guess or simplify code files before committing.... if it doesn't correct itself by end of that session note the idea but close this pr

Fixed for these 2 and don't use kimi without harness again ha..

PR #593, opened 2026-07-21
fix(fsconnect): hard-refuse Windows writes + strict-type safety booleans (codex P1/P2)

cgfixit:
## What

Mined from `docs/codex-findings-7202026.md` (landed in #589) — fixes two documented fsconnect findings, verified still present on `main @19ba973`:

**P1 — Windows write TOCTOU containment gap → hard-refuse.**
The Windows fallback in `pathsafe` validates and reparse-checks a path by NAME, then opens/writes/moves/deletes by NAME. A junction swapped into an in-root component between validation and use redirects the write outside the configured root. POSIX holds directory fds and descends `O_NOFOLLOW`; the Windows path has no equivalent containment and its branches are excluded from test coverage. The security review checklist already documented Windows write-enablement as refused — `writer.py:_executable` now **enforces** it: when `os.name == "nt"`, every write op is hard-refused with `failed_gate="platform"` (audited like any other gate) instead of relying on operator discipline. POSIX behavior is unchanged.

**P2 — quoted YAML booleans fail execution gates OPEN.**
`bool("false") is True`, so `writes_enabled: "false"` (quoted) in config.yaml would silently OPEN the write gate it was meant to close. `config.py` now strict-types all ten boolean gate fields (`_BOOL_FIELDS` + `_validate_booleans()`, run first in `__post_init__`) and the `enabled` loader toggle: a non-bool value raises `FsConnectConfigError` instead of being truthiness-coerced.

## Tests

- `tests/test_fsconnect_writer.py`: +`test_windows_writes_hard_refused` (monkeypatches `os.name` to `"nt"` after fixture roots open; asserts `failed_gate == "platform"`, no file written, refusal audited), +`test_posix_writes_still_execute_with_gates_satisfied` (over-correction guard).
- `tests/test_fsconnect_config.py`: +`test_quoted_bool_gates_fail_closed` (quoted-`"false"` matrix over `enabled`, `writes_enabled`, `allow_hard_delete`, `require_confirm_destructive`, `strict_roots`, `block_on_injection_flags`), +`test_real_bools_still_accepted`.
- Local gates: 47 fsconnect writer/config tests pass; `ruff check --select E,F,I,B,C4,UP,S` clean; invariant-guard 28/28.

## Notes

- Base: `main @19ba973`. All four blobs on this branch were byte-verified (git blob SHA) against the locally tested commit after pushing.
- Recommend squash-merge to keep history tidy (branch includes fix-forward commits restoring full file content).
- Docs-mining run: source finding doc is `docs/codex-findings-7202026.md`; sibling PRs cover the sync, stale-staging, and guardrails findings.


cgfixit:
#595

PR #616, opened 2026-07-22
feat(llm): opt-in Ollama→LM Studio local backend auto-fallback

cgfixit:
## What

Option **C** from the local-backend design: opt-in auto-probe failover for the **local** OpenAI-compatible stack only (not hybrid Grok/Claude).

- `models.local_llm.fallback` in `config.yaml` (`enabled: false` by default)
- `resolve_local_backend()` probes primary with short timeout; on failure probes secondary (LM Studio defaults)
- Requires `fallback.model` when enabled; secondary **must** be loopback
- Process-level cache shared by `LocalLLMClient` and `/health`
- Provider-aware error labels (`Ollama` / `LM Studio`)
- Audit event `local_llm_backend_selected` when fallback wins

## Why

Operators who bounce between Ollama and LM Studio should not re-edit `base_url` by hand. Failover stays offline/loopback and does not touch triple-gated cloud paths.

## Risk to monitor

- Wrong LM Studio model id still fails generate (must set `fallback.model` to the UI/API id)
- Both probes fail → uses primary (boot does not crash; first query errors as today)
- Guardrails/deepagent `base_url` are **not** auto-rewired in v1

## Verification

```text
GROK_API_KEY=dummy pytest tests/test_client.py tests/test_health.py -q
→ 79 passed
ruff check → clean
```

## Enable

```yaml
models:
  local_llm:
    fallback:
      enabled: true
      base_url: http://127.0.0.1:1234/v1
      model: <your-lm-studio-model-id>
```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,8 +1,12 @@
 ## Note: when initiated pr draft from grok build, kimi/kimi code, or codex use preferred branch naming conventions:
 > claude code: claude/{feature}
+> 
 > codex: cx/{feature}
+> 
 > grok build: grok/{feature}
+> 
 > kimi or kimi code: kimi/{feature}
+> 
 > CyClaw (directly or via mcp connector in claude code or similar): CyClaw/{wow!}{feature}-{date} 
 
 ## Title
@@ -78,6 +82,7 @@ If this is a relatively large, complex, or core-path change, kick off the discus
 - Explicit before/after invariant matrix
 - Sandbox validation diff or key evidence
 - Any compensating controls or observability added
+- a mature technical and eli5 version description of whats changed and whats at risk and where to monitor
 
 **Examples of what good "Further comments" look like for core changes:**
 - "No change to graph topology or entry points. RAG-first and audit convergence remain enforced by edges only."
--- a/config.yaml
+++ b/config.yaml
@@ -57,6 +57,17 @@ models:
       # llm/client.py). It means Ollama is stalled ("0% processing"); a retry
       # would just wait another full timeout_sec and, since api.graph_timeout_sec
       # cuts the request first, only orphans a worker thread on a 2nd stalled call.
+    # Optional local backend failover (opt-in). When enabled, a short HTTP probe
+    # prefers primary (above); if unreachable, LM Studio (or any OpenAI-compat
+    # loopback server) is used for LocalLLMClient + /health. Not hybrid/cloud —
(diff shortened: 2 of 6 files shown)


cgfixit:
## Review: approve — no regressions, LM Studio fallback verified

Verified locally on the PR branch (`claude/local-llm-fallback-lmstudio`, based on current main `c9f4e72`):

**Regression checks — all green**
- Full test suite: ran to 100%, zero FAILED/ERROR, pytest exit 0. Only expected skips (POSIX-only fsconnect fixtures, Postgres-DSN tests — environment, not this PR).
- Targeted: `test_client.py` + `test_health.py` (79 passed), `test_graph.py` + `test_gate.py` all pass — the two paths consuming `LocalLLMClient`.
- ruff clean on all touched files.

**Fallback functionality — verified with live probes, not just mocks**
- Both backends down → resolves primary, no crash (~2 × probe timeout), boot never dies; first query errors exactly as today. Fail-open to existing behavior.
- Primary up → primary wins.
- Non-loopback fallback URL → hard-rejected (`LLMServiceError`) — can't be pointed at a LAN/cloud endpoint. Consistent with the offline/loopback invariant.
- `enabled` without `model` → rejected at resolution; `/health` degrades to an unhealthy `local_llm` status instead of 500ing.
- **Disabled (default) → zero probes, zero behavior change** — existing Ollama-only installs are byte-for-byte unaffected, which is why regression risk is genuinely nil.
- Failover emits `local_llm_backend_selected`; triple-gated Grok/Claude paths untouched; error labels stay redacted and provider-aware.

**Non-blocking nits (fine to merge as-is)**
1. If the fallback has no `api_key`, the primary's key is sent to the fallback server (`resolved.api_key or llm_cfg...`) — harmless on loopback.
2. Backend selection is process-cached — Ollama recovering mid-run isn't noticed until restart. Documented, but worth an operator note.
3. `/health` renames the status from `ollama` to `lmstudio` when fallback wins — anything scraping by name sees a new key.

Safe to undraft and merge.

— Kimi Code CLI (local review)

cgfixit:
@codex review and provide feedback on if anything was missed? I only want to add the lm studio (no config change required) failback if ollama not installed or online if it doesn't negatively impact ollama functionality nor bloat the code overall. Ollama is def the finished product choice but for messin around honestly lm studio is sometimes easier. Plus for whatever reason I don't like the way ollama has the model marketplace it makes me feel like I don't know where I'm downloading the Qwen 7B model from whereas with lm studio it's only 1 and not uploaded by a rando. I might have done something wrong though but the 7B instruct had random letters after and 2-3 different uploaders..

PR #619, opened 2026-07-22
fix(fsconnect): atomic trash sidecar ordering + collision-proof trash names

cgfixit:
## What

Two small trash-integrity bugs in fsconnect, one concept (a trashed payload must always stay discoverable and movable):

1. **Sidecar written after the payload move** (`agentic/fsconnect/writer.py` `_to_trash`): the payload was moved into `.cyclaw-trash/` before its `.meta.json` sidecar was written. A crash or sidecar-write failure in between left a payload with no metadata — and `trash.list_entries` only reads `*.meta.json`, so the payload was invisible to `trash-list`/`trash-empty` forever. Now the sidecar is written first, and if the move then fails the sidecar is rolled back. Worst case (crash between sidecar-write and move) is an orphan sidecar — visible to `trash-empty` — instead of an undiscoverable payload.
2. **Trash names still collided for same-path, same-second deletes** (`agentic/fsconnect/trash.py` `entry_name`): the "disambiguator" was `sha256(original + second-stamp)[:8]` — both inputs identical for two deletes of the same path within one second, so the second delete's `overwrite=False` move failed with an already-exists error (the docstring claimed the opposite). Added a `secrets.token_hex(4)` random suffix; docstring updated to match.

Also: `pathsafe._mkdir_win` now wraps `FileExistsError` in `FsConnectRuntimeError`, mirroring the POSIX branch, so `writer._ensure_trash`'s `suppress(FsConnectRuntimeError)` actually works on the Windows fallback (second delete previously died on a raw `FileExistsError`).

## Why

Data-recoverability: trash is the safety net for every `fs_delete`. A payload stranded in `.cyclaw-trash/` with no sidecar can never be listed, restored, or emptied — silent data loss disguised as successful deletion. And a disambiguator that disambiguates nothing turns a routine "delete, recreate, delete again" flow into a confusing runtime error.

## Risk to monitor

- Entry-name format changed (16-hex suffix instead of 8): anything parsing trash entry names by fixed digest length would notice. Only the dry-run plan preview and the sidecar filenames are affected; `trash.list_entries` keys off the `.meta.json` suffix, not the name shape.
- Sidecar-first means a crash mid-sequence can leave an orphan sidecar whose payload never moved; `trash_empty` already tolerates missing payloads and removes the sidecar, so this self-heals.
- New tests monkeypatch `_writes_refused_platform` off inside a `tmp_path` root so they exercise the Windows fallback in tests only — production still hard-refuses writes on Windows.

## Tests

- `tests/test_fsconnect_trash_integrity.py` (new, cross-platform): sidecar-write failure leaves no untracked payload; two same-path same-second deletes both succeed and both list.
- `tests/test_fsconnect_trash.py`: same-path/same-second names no longer collide; suffix length updated.
- `GROK_API_KEY=dummy python -m pytest tests/ -q --tb=short -k fsconnect` → exit 0 (50 passed, 134 POSIX-only skips); full suite → exit 0 (1269 passed, 150 skipped, 0 failures).

### DIFF
--- a/agentic/fsconnect/pathsafe.py
+++ b/agentic/fsconnect/pathsafe.py
@@ -684,7 +684,14 @@ class ScopedRoots:
 
     def _mkdir_win(self, sr: SafeRoot, comps: list[str]) -> dict:  # pragma: no cover
         real = self._win_resolve(sr, comps, must_exist=False)
-        real.mkdir()
+        try:
+            real.mkdir()
+        except FileExistsError as exc:
+            # Mirror the POSIX branch: callers (e.g. writer._ensure_trash) suppress
+            # FsConnectRuntimeError for the already-exists case, not raw OSError.
+            raise FsConnectRuntimeError(
+                f"{real} already exists", details={"path": str(real)}
+            ) from exc
         return {"created": str(real)}
 
     def _move_win(  # pragma: no cover
--- a/agentic/fsconnect/trash.py
+++ b/agentic/fsconnect/trash.py
@@ -1,7 +1,7 @@
 """Trash-first governed-deletion helpers for the filesystem connector (Phase 2).
 
 This module is pure policy/formatting glue on top of the ``pathsafe`` mechanism: it
-computes deterministic, collision-resistant trash entry names, serializes/parses the
+computes collision-proof, time-sortable trash entry names, serializes/parses the
 ``.meta.json`` sidecars, and lists/expires entries. It performs NO privileged syscall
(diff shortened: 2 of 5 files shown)


cgfixit:
Review follow-up (bb49ce3): orphan trash sidecars are now actually reclaimed by `trash_empty` — a missing payload is treated as already-removed and the sidecar is unlinked, so the entry is purged instead of skipped forever. Added a regression test in `tests/test_fsconnect_trash_integrity.py` (orphan sidecar with no payload is purged and deleted); scoped and full test suites pass.

PR #620, opened 2026-07-22
fix(sqlconnect): fail-closed MSSQL read-only + reject lock-taking SELECT hints

cgfixit:
## What

Two hardening fixes to the sqlconnect read-only MSSQL connector (`agentic/sqlconnect/client.py`):

1. **Fail-closed read-only enforcement** (`_execute`): previously `conn.read_only = True` was wrapped in `suppress_attr_error()`. pyodbc connections have no settable `read_only` property, so on MSSQL the read-only session was **silently never applied** and queries ran on a read-write connection. Now a new `_enforce_read_only()`:
   - tries `conn.read_only = True` (psycopg path, unchanged),
   - falls back to pyodbc's own mechanism `conn.set_attr(SQL_ATTR_ACCESS_MODE, SQL_MODE_READ_ONLY)`,
   - raises `SqlConnectRuntimeError` if neither mechanism exists or the driver rejects the attribute — the query never executes.
2. **Reject write-grade lock hints**: `_FORBIDDEN_RE` now includes `UPDLOCK`/`HOLDLOCK`/`XLOCK`, so `SELECT * FROM t WITH (UPDLOCK)` no longer passes the SELECT-only guard.

Tests added in `tests/test_sqlconnect_client.py` (all mock-based, no live DB):
- lock-hint SELECTs are rejected (`WITH (UPDLOCK)` / `(HOLDLOCK)` / `(XLOCK)`, incl. combined/lowercase),
- pyodbc-style connections get `SQL_ATTR_ACCESS_MODE=SQL_MODE_READ_ONLY` via `set_attr`,
- fail-closed when the driver supports neither mechanism, and when `set_attr` raises — user query never reaches the connection,
- psycopg-style path still sets `read_only = True`.

## Why

Security/availability:
- The connector's contract is read-only; silently losing the read-only session on pyodbc was a **fail-open** — a read-only tool executing on a read-write connection.
- `WITH (UPDLOCK)` takes update-grade locks from inside an otherwise valid SELECT — an availability/DoS risk on a connector whose only purpose is safe read previews.

## Risk to monitor

**Behavior change:** setups on drivers that cannot set a read-only session (no `read_only` property and no `set_attr`/ODBC access-mode constants) previously ran fine (but *not* actually read-only) and will now raise `SqlConnectRuntimeError` instead of executing. That is the intended fail-closed behavior, but previously-working deployments may start erroring. Real pyodbc supports `set_attr` + `SQL_ATTR_ACCESS_MODE`, so genuine MSSQL deployments are unaffected; psycopg (Postgres) is byte-for-byte the same code path as before.

## Tests

- `GROK_API_KEY=dummy python -m pytest tests/ -q --tb=short -k sqlconnect` → pass (122 tests; 19 new, all ran — none skipped)
- `GROK_API_KEY=dummy python -m pytest tests/ -q --tb=short` → 1425 tests, 0 failures, 0 errors, 150 skipped (pre-existing skips only: POSIX fixtures, live-Postgres, optional deps), exit 0

### DIFF
--- a/agentic/sqlconnect/client.py
+++ b/agentic/sqlconnect/client.py
@@ -35,15 +35,16 @@ _IDENT_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*(\.[A-Za-z_][A-Za-z0-9_]*)?$")
 # rejected legitimate read queries like ``SELECT replace(name,'a','b') FROM t``.
 # Even if a REPLACE write statement existed, the leading-keyword gate (must start
 # with SELECT/WITH) plus the single-statement check would already stop it.
-# ``updlock``/``holdlock``/``xlock`` are MSSQL table hints that take
-# write-grade locks inside an otherwise valid SELECT (``SELECT * FROM t WITH
+# ``updlock``/``holdlock``/``xlock``/``tablock``/``tablockx``/``paglock``/
+# ``serializable`` are MSSQL table hints that take write-grade or aggressively
+# escalated locks inside an otherwise valid SELECT (``SELECT * FROM t WITH
 # (UPDLOCK)``). On a connector whose contract is read-only that is an
 # availability risk (blocking other writers/readers), so the hints are
 # forbidden even though the statement itself only reads.
 _FORBIDDEN_RE = re.compile(
     r"\b(insert|update|delete|drop|alter|create|truncate|grant|revoke|merge|call|"
     r"exec|execute|into|copy|vacuum|attach|begin|commit|rollback|"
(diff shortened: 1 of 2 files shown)


cgfixit:
Security review follow-up: extended `_FORBIDDEN_RE` to also reject the TABLOCKX and SERIALIZABLE (= HOLDLOCK synonym) MSSQL table hints, plus TABLOCK and PAGLOCK per the same lock-escalation/availability rationale. Added the new hints to the parametrized lock-hint rejection test (10 cases, all pass; full suite green).

PR #622, opened 2026-07-22
test: harden guard/fusion/injection/personality/indexer-symlink coverage

cgfixit:
## What

Test-only hardening batch — five coverage gaps where a guard existed but nothing proved it would actually fire. No runtime code changes.

1. **`tests/test_agentic_isolation.py`** — negative self-tests for the I6 isolation guard: planted `import agentic` / `from agentic import ...` sources and a planted agentic-side `import gate`/`from graph import ...` must trip `_imports()` and the forbidden set. A broken `_imports()` previously greened the whole suite.
2. **`tests/test_hybrid_search.py`** — `TestRRFFusion` previously recomputed RRF arithmetic on bare Python literals and stayed green no matter what the real fusion formula did. New `test_real_fusion_path_combines_both_legs` drives the REAL `HybridRetriever.hybrid_search` with stubbed vector/BM25 legs (existing SimpleNamespace pattern) and asserts dual-leg fusion (two RRF terms, `hybrid` mode, contribs, ordering) end-to-end.
3. **`tests/test_gate.py`** — `/soul/apply`'s `PromptInjectionError → 400` branch was never triggered (only `/query`'s `check_input` patch was). New test patches `personality.apply_evolution` to raise and asserts HTTP 400 + the `soul_apply_injection_blocked` audit event.
4. **`tests/test_personality.py`** — (a) parametrized empty/whitespace `reason` rejection (I5 human gate, no write/DB/version change); (b) `restore_from_backup()` with absent `.bak` raises `FileNotFoundError` leaving state untouched.
5. **`tests/test_indexer.py`** — corpus symlink-escape guard (`retrieval/indexer.py:71-76`, test-spec tier 1.2): a symlink resolving outside the corpus is skipped with a warning; one resolving inside is still indexed. Real-symlink tests `pytest.skip` only when `os.symlink` itself is denied (WinError 1314); a deterministic `resolve()`-stubbed test exercises the guard branch on every platform.

## Why

Guards that can't prove they'd catch a violation are false confidence. Each of these branches could silently rot — a weakened `_imports()`, a changed fusion formula, a dropped 400 mapping, a removed reason check, or a deleted symlink guard would all have stayed green before this batch.

## Risk to monitor

- The two real-symlink tests skip on Windows hosts without the symlink privilege (expected; they run on Linux CI). The stubbed guard-branch test covers the logic regardless — if CI ever gains/loses symlink privileges, watch for the skip counts changing.
- `test_soul_apply_injection_blocked` depends on `gate.personality` being enabled by the repo's shipped `config.yaml` (same assumption as the existing `/soul` 200 test); if the shipped config disables personality, both need revisiting.

## Verification

- Targeted: `pytest tests/test_agentic_isolation.py tests/test_hybrid_search.py tests/test_gate.py tests/test_personality.py tests/test_indexer.py -q` — all pass (2 sanctioned symlink-privilege skips); every new test confirmed PASSED via `--junitxml` (not silently skipped).
- Full suite: `pytest tests/ -q` — 1429 tests, 0 failures, 0 errors, 152 skipped (pre-existing Postgres-DSN skips + the 2 symlink skips).

### DIFF
--- a/tests/test_agentic_isolation.py
+++ b/tests/test_agentic_isolation.py
@@ -41,6 +41,36 @@ def test_request_path_does_not_import_agentic(module_file):
     )
 
 
+@pytest.mark.parametrize("planted_source", [
+    "import agentic\n",
+    "from agentic import fsconnect\n",
+    "import os, agentic.fsconnect\n",
+])
+def test_guard_flags_planted_agentic_import(tmp_path, planted_source):
+    # Negative self-test: if _imports() ever stopped detecting an `import agentic`
+    # (or was weakened to a no-op), test_request_path_does_not_import_agentic
+    # would green on a real violation. Feed the guard a planted temp file and
+    # assert the flag actually fires.
+    planted = tmp_path / "gate.py"
+    planted.write_text(planted_source, encoding="utf-8")
+    imported = _imports(planted.read_text(encoding="utf-8"))
+    assert "agentic" in imported, (
+        f"guard is blind: a planted {planted_source.strip()!r} must be flagged"
+    )
+
+
+def test_reverse_guard_flags_planted_request_path_import(tmp_path):
+    # Symmetric negative self-test for test_agentic_does_not_import_request_path:
+    # a planted agentic-side module importing gate/graph must trip the forbidden set.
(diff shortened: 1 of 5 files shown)


cgfixit:
*

PR #637, opened 2026-07-22
fix(config: KEEP BUT LEAVE COMMENT audit_fields knobs (silent no-ops on audit log)

cgfixit:
## What

Removes four `logging.audit_fields` keys from `config.yaml` that no code reads:

- `include_top_score`
- `include_retrieval_mode`
- `include_online_escalated`
- `include_model_used`

**Proof:** repo-wide grep for each key name across *.py / *.yaml / *.md / tests/ / docs/ finds them only in `config.yaml` (the definitions), the `tests/conftest.py` and `tests/test_audit.py` fixtures (which set but never assert them), and one historical audit doc. `utils/logger.py` (`audit_log()`, ~line 242) reads exactly one key from `audit_fields`: `include_query_hash`. The audit record's other fields (`top_score`, `retrieval_mode`, `online_escalated`, `model_used`, written by `graph.py`'s audit node) are emitted unconditionally — there are no per-field include toggles in the code.

Also in this PR:
- Fixture lines setting the dead keys removed from `tests/conftest.py` and `tests/test_audit.py` (no assertions touched — none existed).
- Short comment left in `config.yaml` stating only `include_query_hash` is read, so the knobs don't grow back.
- One-line update note in `docs/audits/PR_TEST_AUDIT_DETAILED.md` where the finding enumerated the removed keys (historical finding text otherwise preserved).
- Validators checked, no update needed: `.claude/skills/config-guard/check_config.py` (C1–C12) and `utils/config_validation.py` never reference `audit_fields`.

## Why / benefit

The audit log is the compliance surface of this project. Four exposed config knobs that silently do nothing are worse than no knob: an operator can flip `include_model_used: false` believing they have tightened what the audit log records, while the field keeps being written. Deleting them makes the config honest — what you see under `audit_fields` is everything you can actually toggle.

## Risk to monitor

Out-of-repo tooling (an operator's local scripts, dashboards, or config linters) could reference these key names — they'd now be absent from the shipped `config.yaml`. In-repo impact is nil (grep-verified), and any external reader treating them as optional keys is unaffected since nothing ever acted on them.

## Verification

- `GROK_API_KEY=dummy python -m pytest tests/test_audit.py tests/test_conftest_fixtures.py -q --tb=short` — green (32 passed)
- `GROK_API_KEY=dummy python -m pytest tests/ -q --tb=short` — exit 0 (only the usual Postgres-DSN skips)
- `python -c "import yaml; yaml.safe_load(...)"` on `config.yaml` — parses, `audit_fields == {'include_query_hash': True}`
- `ruff check --select E,F,I,B,C4,UP,S tests/conftest.py tests/test_audit.py` — clean
- `.claude/skills/config-guard/check_config.py` — 0 failures, 0 warnings

Found by automated cyclaw-optimize ponytail-lens scan.


cgfixit:
@codex @claude review

cgfixit:
Honestly I'd prefer this to be a comment stating it does nothing rn than removing it cuz it should be config held

PR #648, opened 2026-07-22
[fix] - harness runs listing keeps real runs; port override bounded; dead constant dropped

cgfixit:
## Proposed changes

Three small harness server/config fixes from the optimize scan, each verified against the code:

1. **`/api/harness/runs` slice-before-filter** (`harness/server.py`): the endpoint took the newest `_MAX_RUNS` directory *entries* and only then filtered to directories — a stray file (index, `.lock`) inside the window silently displaced a real run from the response. Now filters `is_dir()` first, then slices. The regression test **fails on the old code** (proved by reverting `server.py` and re-running) and passes on the fix.
2. **`CYCLAW_HARNESS_PORT` bounds** (`harness/server.py` `main()`): the env override accepted any digit string — `"0"` bound an ephemeral port (the console link silently breaks), `"99999"` crashed uvicorn at bind. It now applies the same `1024–65535` bounds `config.py` already enforces on the stored port, exiting loudly like the adjacent loopback-host check. *Slight behavior change:* out-of-range values previously slipped through to uvicorn; they now exit with a clear message.
3. **Dead constant** (`harness/config.py`): `_SKILLS_SRC` was defined and never referenced anywhere (grep-verified repo-wide; `_seed_skills` recomputes the identical path inline). Deleted.

**Invariant / Governance Impact:** None — out-of-band `harness/` only; I6 isolation untouched. `invariant-guard`: 28 passed / 0 failed. The loopback-only bind gate is unchanged (the port check sits after it and only narrows accepted input).

## Types of changes

**Scope note:** Out-of-band harness layer.

## Benefits / why

- The runs pane reports every real run even when tooling drops stray files into the runs directory.
- An explicit-but-invalid port override fails loudly at startup instead of silently binding somewhere else or stack-tracing in uvicorn — consistent with the adjacent host-check behavior.
- One less dead symbol for a reader to chase.

## Risks to monitor

- The port check also (theoretically) covers `cfg.port`, but stored config is already bounds-checked at load and the default is 8790, so only env values can trigger the exit. If a future caller wants ports &lt;1024, that's a deliberate config.py decision, not an env-path one.

## Further comments

Verified: `ruff` + `flake8`/WPS (pinned 1.6.2 against `setup.cfg`, the exact changed-files CI leg) clean on all three files; `pytest tests/test_harness.py` → **25 passed**; regression proof done by revert; `invariant-guard` 28/0. **Trial-merged with #647** (both PRs append tests to `tests/test_harness.py` in non-adjacent regions): merges clean in sequence, zero conflict markers, both tests present, combined tree **26 passed** — safe to merge in either order.

https://claude.ai/code/session_0138hmScu3tFU4v1jzvFUwVT

---
_Generated by [Claude Code](https://claude.ai/code/session_0138hmScu3tFU4v1jzvFUwVT)_

### DIFF
--- a/harness/config.py
+++ b/harness/config.py
@@ -35,7 +35,6 @@ from utils.errors import AgenticError
 logger = logging.getLogger("cyclaw.harness.config")
 
 _REPO_ROOT = Path(__file__).resolve().parent.parent
-_SKILLS_SRC = _REPO_ROOT / ".claude" / "skills"
 
 DEFAULT_HOST = "127.0.0.1"
 DEFAULT_PORT = 8790
--- a/harness/server.py
+++ b/harness/server.py
@@ -22,7 +22,7 @@ from pathlib import Path
 from fastapi import FastAPI, HTTPException
 from fastapi.responses import FileResponse
 
-from harness.config import HarnessConfig
+from harness.config import _MAX_PORT, _MIN_USER_PORT, HarnessConfig
 from harness.ollama import HarnessChatClient, HarnessLLMError
 from harness.prompts import compose_system_prompt
 from harness.registry_view import full_registry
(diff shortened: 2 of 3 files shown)


cgfixit:
Note config edge case distinction that you may forget later

cgfixit:
Noted for the record — the config edge-case distinction this PR introduced:

| Path | Invalid / out-of-range port | Behavior |
|---|---|---|
| **Stored config** (`~/.CyClaw/config.json` → `config.py` `_apply_stored`) | e.g. `"port": 99999` or wrong type | **Silently ignored** — keeps the running value (default 8790). Ambient state degrades quietly, matching `_apply_stored`'s type-checked field-by-field style. |
| **Env override** (`CYCLAW_HARNESS_PORT` → `server.py` `main()`) | e.g. `0` or `99999` | **Loud `sys.exit`** with the offending value. An explicit per-run operator intent fails loudly, matching the adjacent `CYCLAW_HARNESS_HOST` loopback check. |

Both use the same bounds (`_MIN_USER_PORT`/`_MAX_PORT` = 1024–65535, defined in `harness/config.py`). Only env values can ever trigger the exit — stored config is bounds-checked at load. If sub-1024 ports are ever wanted, that's a deliberate `config.py` bounds decision, never an env-path change.

(Also captured in the machine-local memory store — `docs/memories/` is gitignored by design, so this comment is the durable copy.)

---

PR #652, opened 2026-07-23
[fix] - retire LM-Studio-era leftovers: orphaned sandbox mock + config host consistency

cgfixit:
## Proposed changes

Two small Ollama-migration leftovers from the Phase 1 scan:

1. **Delete `.claude/skills/CyClaw-Sandbox/mock_lmstudio.py`** (port 1234, simulates `qwen2.5-7b-instruct`). Orphaned since the sandbox flow moved to `mock_ollama.py`: repo-wide grep finds no active reference — only self-references and dated audit docs — while the active `mock_ollama.py` is referenced by the sandbox skill, the `.codex` variant, and `tests/`. ⚠️ **Deletion of an operator-authored file — flagging explicitly for your review** (that's why this is its own clearly-labeled draft; nothing is lost until you merge, and git history retains it regardless). Note: #415's Gap 2 (SKILL.md launching the mock on 1234) is already resolved on main — Phase 3 no longer launches any mock LLM server; this file is the残 leftover.
2. **`config.yaml` `agentic.deepagent_github.base_url`: `localhost` → `127.0.0.1`** — the lone `localhost` local-endpoint while every sibling (`models.local_llm`, `fallback`, `guardrails`) uses `127.0.0.1`. Both pass `_is_loopback`; consistency only. Carries the same `DevSkim: ignore DS162092,DS137138` marker as the sibling loopback URLs so scanner treatment stays uniform.

**Invariant / Governance Impact:** None — a dead file under `.claude/` (ruff-excluded, not production code) and a host-string form change inside a disabled-by-default block. `config-guard`: 0 failures / 0 warnings.

## Types of changes

**Scope note:** Out-of-band tooling + config hygiene.

## Benefits / why

- Removes the last piece of dead LM-Studio-era scaffolding that could mislead an agent or operator following the sandbox skill (the file's own docstring instructs launching it on 1234 against a config that now expects 11434).
- One host-string convention across every loopback endpoint in `config.yaml`.

## Risks to monitor

- If any private/local workflow still launches `mock_lmstudio.py` by hand, it disappears — `git checkout <sha> -- .claude/skills/CyClaw-Sandbox/mock_lmstudio.py` restores it, and `mock_ollama.py` in the same directory covers the same OpenAI-compatible surface on the current port.

## Further comments

Verified: `config-guard` 0/0; `config.yaml` parses; remaining `localhost` strings are the intentional `allowed_origins`/`allowed_hosts` lists (untouched); `tests/test_agentic_deepagent_optional.py` unaffected (skips without optional deps, as designed). Companion to #651 (harness fallback parity) — no shared files between the two.

https://claude.ai/code/session_0138hmScu3tFU4v1jzvFUwVT

---
_Generated by [Claude Code](https://claude.ai/code/session_0138hmScu3tFU4v1jzvFUwVT)_


cgfixit:
For later reference: the deleted mock tests for lm studio were truly redundant; current mock_ollama.py does have the option of failback functionality to lm studio

PR #675, opened 2026-07-28
fix(privacy): redact xAI (xai-...) API keys in error sanitizer and audit log

cgfixit:
## What

Adds the xAI Grok API-key shape (`xai-[A-Za-z0-9]{20,}`) to both redaction lists:

- `gate.py` `_SECRET_PATTERNS` (HTTP 500 body sanitizer)
- `config.yaml` `policy.privacy.redact_secrets_like` (audit-log redaction via `utils/logger.redact_sensitive`)

plus one pattern-leg test (`test_xai_style_key_pattern_redacted`) mirroring `test_anthropic_style_key_pattern_redacted`.

## Why

Found during the 2026-07-28 sandbox verification run (anomaly A3): every integrated credential shape was covered by both lists — `sk-ant`, `Bearer`, `api_key=`, `AKIA`, `xox`, `ghp_`, OpenAI `sk-` — **except** the xAI Grok key format, which has no `sk-` prefix and no `Bearer`/`api_key` anchor. A Grok key surfacing in a traceback would reach HTTP 500 response bodies and `logs/audit.jsonl` verbatim. CyClaw's only external paid provider is the one whose key shape was missing.

Adding a pattern is safe (per CLAUDE.md pattern guidance); no coverage removed. Both lists stay in sync — `redact_secrets_like` is the documented single source of truth for the audit path and `_SECRET_PATTERNS` covers the HTTP path.

## Verification (run locally before push)

- Full suite: **1554 passed, 14 skipped, 0 failed** (1553 on main + 1 new test)
- `ruff check --select E,F,I,B,C4,UP,S gate.py tests/test_gate.py` — clean
- `config-guard --strict` — 0 errors / 0 warnings
- `invariant-guard` — 28 passed / 0 failed
- New test confirmed to fail against main (shape uncovered) and pass with the pattern
- All three files pushed via API and re-fetched: blob SHAs byte-exact vs local

## Risk to monitor

Over-redaction false positives: the pattern requires the literal `xai-` prefix followed by 20+ alphanumerics, so prose mentioning "xai" is untouched; a string like `xai-<20+ chars>` in a log line that isn't a key would be redacted — acceptable collateral for a secret-shaped token. No behavior change for any existing pattern.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -285,6 +285,7 @@ policy:
       # Anthropic keys (sk-ant-api03-...) contain hyphens inside the token body,
       # so the OpenAI-style pattern above (no hyphens allowed) never matches them.
       - "sk-ant-[a-zA-Z0-9_-]{20,}"  # Anthropic (Claude) API keys
+      - "xai-[a-zA-Z0-9]{20,}"       # xAI (Grok) API keys
 
 api:
   host: "127.0.0.1"  # DevSkim: ignore DS162092 - loopback-only binding by design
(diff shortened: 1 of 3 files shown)


cgfixit:
@codex review (seems like it needs an audit-path regression test beyond what kimi did locally before committing)

PR #676, opened 2026-07-28
fix(security): keep soul persistence consistent

cgfixit:
## Summary
- compensate `soul.md`, its prior `.bak`, and the version transaction when any publication step fails
- create all soul temporary files owner-only before writing, then atomically replace the destination
- create and repair the live soul and SQLite history database as owner-only on POSIX
- audit failed updates and raise a typed consistency error if compensation itself cannot complete

## Security impact
Prevents a failed `/soul/apply` from activating an unversioned soul or destroying the prior recovery copy. It also closes the `0644` creation/temp-file exposure for soul content and SQLite history on shared POSIX hosts.

## Validation
- `python -m pytest tests/test_personality.py -q --tb=short` (pass; 5 POSIX mode tests skipped on Windows)
- `python -m pytest tests/test_due_diligence_invariants.py tests/test_gate.py tests/test_runtime_errors.py -q --tb=short` (pass)
- `python .claude/skills/invariant-guard/check_invariants.py` (28 passed)
- `python -m ruff check utils/errors.py utils/personality.py utils/personality_db.py tests/test_personality.py` (pass)
- `python -m py_compile utils/errors.py utils/personality.py utils/personality_db.py` (pass)
- full `tests/` run reached 100%; its only failure was `test_windows_missing_schtasks_raises`, reproduced unchanged on clean `main` because this host has a file at `~/.config/rclone/logs`

## Unverified locally
- live PostgreSQL integration (no PostgreSQL service)
- POSIX permission assertions (Windows host; CI should exercise them)

### DIFF
--- a/tests/test_personality.py
+++ b/tests/test_personality.py
@@ -5,8 +5,10 @@ Run: pytest tests/test_personality.py -v
 
 import os
 import threading
-import pytest
+from pathlib import Path
 from unittest.mock import patch
+
+import pytest
 
 @pytest.fixture
 def tmp_paths(tmp_path):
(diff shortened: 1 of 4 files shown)


cgfixit:
Verified the compensation logic (imports, `audit_log(cfg=)`, `self.cfg`, the `fd = -1` double-close guard) — all sound.

Found and fixed one real defect during review: `_load_soul()`'s `os.chmod(self.soul_path, 0o600)` and `personality_db.connect()`'s `os.chmod(db_path, 0o600)` ran unconditionally on every load/connect. `os.chmod()` requires the calling process to own the file (or be root); a root-installed service running as a lower-privileged user, or an ops deployment that intentionally shipped `soul.md` or the SQLite history DB read-only, would hit `PermissionError` on the very first load and CyClaw would fail to start — a regression from the pre-PR behavior, which just read the file.

Pushed `30d60cf`: stat the current mode first and only `chmod` on mismatch, wrapped in `try/except OSError` with a warning — same pattern the PR already uses for the atomic-write temp files (which are always freshly created, hence always owned by this process, hence never at risk). Added regression coverage for both the graceful-degradation path (`chmod` raises `PermissionError`, construction still succeeds) and the skip-when-already-correct path (`chmod` is never called when the mode already matches).

Also re-confirms the scope caveat already called out in the PR body: this is exception safety, not crash safety — a `SIGKILL`/power-loss between the file and DB compensation steps can still leave an unversioned soul active. Worth keeping in the PR description as-is.

Validated under Python 3.12: `tests/test_personality.py` (incl. the 4 new tests) green, `tests/test_due_diligence_invariants.py` + `tests/test_gate.py` + `tests/test_runtime_errors.py` green, full `tests/` suite green (exit 0), `ruff check` clean, `invariant-guard` 28/28, no new `mypy --strict` findings on the touched lines.

---

PR #678, opened 2026-07-28
security: repair NeMo guardrails integration

cgfixit:
## Summary
- replace the nonexistent `nemoguardrails==0.18.2` extra with published `0.23.0` and constrain it consistently
- read NeMo's documented `last_user_message` action context so injection and soul-mutation rails inspect the actual current input
- force NeMo anonymous usage-stat opt-out before both gateway and standalone guardrails import paths
- add regression coverage for current/legacy contexts and both telemetry entry points

## Security impact
Before this change, the opt-in NeMo extra could not resolve, and a real 0.23.0 engine passed the current message under `last_user_message` while CyClaw read `user_message`; those defense-in-depth actions therefore allowed tested injection and soul-mutation payloads. The standalone guardrails path also bypassed `gate.py`'s telemetry kill. This change closes all three paths without changing the graph topology or making guardrails a core dependency.

References:
- https://docs.nvidia.com/nemo/guardrails/configure-guardrails/actions/action-parameters
- https://docs.nvidia.com/nemo/guardrails/latest/resources/telemetry
- https://pypi.org/project/nemoguardrails/0.23.0/

## Validation
- `python -m pytest` across the complete guardrails and telemetry family: 108 passed
- isolated real-wheel NeMo 0.23.0 probe: `LLMRails` constructed; telemetry opt-out asserted; injection and soul mutation blocked
- focused tests with the isolated 0.23.0 install active: 55 passed
- `python .claude/skills/dep-guard/check_deps.py --strict`: 0 failures, 0 warnings
- documented constrained pip resolver dry-run with the PyTorch CPU index: passed
- `python .claude/skills/invariant-guard/check_invariants.py`: 28 passed
- Ruff, `py_compile`, and `git diff --check`: passed
- independent agentic-AI specialist review: GO, no merge blockers

## Local full-suite note
The full Windows suite completed with one unrelated host-state failure: `tests/test_sync_scheduler.py::test_windows_missing_schtasks_raises` cannot create `C:\Users\cgrady\.config\rclone\logs` because that path is a file. The same failure was reproduced on an untouched `origin/main` clone; changed tests and CI-targeted suites are green.


cgfixit:
Independently verified the three technical claims against the real PyPI index (not just the wheel already inspected in the PR body):

- `nemoguardrails==0.18.2` genuinely does not exist — confirmed via `pip download` against the live index. Available versions jump `0.18.0 → 0.19.0 → ... → 0.23.0`; nothing in between was ever published (not yanked, never existed). `0.23.0` resolves and downloads cleanly.
- `guardrails/rails.py`'s `_context_user_message()` fallback (`last_user_message` → `user_message`) matches the diff exactly.
- `NEMO_GUARDRAILS_NO_USAGE_STATS=1` is set in both `gate.py`'s telemetry-kill block and `guardrails/__init__.py` (for the standalone `python -m guardrails.cli` path that never passes through `gate.py`).

Also independently confirmed the offline-first tension: unzipped the `nemoguardrails-0.23.0-py3-none-any.whl` METADATA and checked `Requires-Dist` with no `extra ==` marker (i.e., base install, not an optional NeMo extra) —

```
fastembed (>=0.2.2) ; python_version >= "3.10" and python_version < "3.14"
onnxruntime (>=1.17.0) ; python_version >= "3.11"
pandas (>=1.4.0,<3)
protobuf (>=5.29.5)
aiohttp (>=3.10.11)
```

All five are unconditional base dependencies of `nemoguardrails` itself, not gated behind any extras — so `pip install .[guardrails]` pulls them in regardless of how narrowly CyClaw's own rails are configured. `fastembed`'s documented default behavior is to fetch ONNX embedding models from a remote CDN on first use. That's real friction against `docs/THREAT_MODEL.md`'s offline-first, no-egress posture, even though the extra itself stays opt-in and `guardrails.enabled` stays `false` by default.

This doesn't change the verdict on the code fix — it's correctly diagnosed and necessary (the extra literally couldn't resolve before, and the rails were reading a key NeMo never populated, i.e. silent no-ops). But whether to ship the extra at all before that network-fetch behavior is documented or constrained is a product call, not something I'm making unilaterally. Flagging for your decision rather than editing the PR.

---

cgfixit:
Per your call above: pushed `7f95835`, a comment-only caveat next to the `guardrails` extra in `pyproject.toml` (and mirrored in `constraints.txt`) documenting that installing it pulls in `fastembed`'s network-fetch-by-default behavior. No code/behavior change — `dep-guard --strict`, `ruff`, and the guardrails/telemetry test family stay green. This doesn't change the merge verdict; it just makes sure a future "why is this hitting the network" doesn't have to be re-discovered.

---

PR #679, opened 2026-07-28
security: repair Falco detection rules + 725 !!

cgfixit:
## Summary
- load Falco's shipped default rules before CyClaw's custom rules when using explicit `-r` arguments
- require both the loopback destination and port 11434 for the Ollama egress allowance
- document the paired host/port requirement and add static regression tests for both detection properties

## Security impact
The monitoring profile explicitly supplied only the custom rules file, so Falco ignored its configured default rules even though CyClaw referenced default macros (`spawned_process`, `open_write`, and `outbound`). That could leave the optional monitor in a startup failure/restart loop. Separately, `or fd.sport = 11434` treated every external destination on that port as expected, suppressing the unexpected-egress alert.

References:
- https://falco.org/docs/concepts/rules/default-custom/
- https://falco.org/docs/concepts/rules/overriding/
- Falco 0.39.2 configuration: https://github.com/falcosecurity/falco/blob/0.39.2/falco.yaml

## Validation
- `python -m pytest tests/test_falco_detection.py -q`: 2 passed
- YAML parsing for `docker-compose.yml` and `deploy/falco/falco_rules.yaml`: passed
- Ruff, `py_compile`, and `git diff --check`: passed
- `python .claude/skills/invariant-guard/check_invariants.py`: 28 passed
- independent detection-engineering review: GO, no merge blockers

## Runtime validation still required
Docker/Falco is unavailable on this Windows host. Before merge, run Falco 0.39.2's native validation on a Docker-capable Linux host against the two ordered rule files, then start the `monitoring` profile and confirm `cyclaw-falco` remains healthy.

## Local full-suite note
The full Windows suite completed with one unrelated host-state failure: `tests/test_sync_scheduler.py::test_windows_missing_schtasks_raises` cannot create `C:\Users\cgrady\.config\rclone\logs` because that path is a file. The same failure was reproduced on untouched `origin/main`; the new detection tests are green.

### DIFF
--- a/deploy/falco/README.md
+++ b/deploy/falco/README.md
@@ -46,7 +46,9 @@ deployment:
 
 1. `cyclaw_container` — the app container name (default `cyclaw-prod`).
 2. `cyclaw_expected_outbound` — your Ollama host/port if the model is not on
-   the shipped default (`127.0.0.1:11434`).
+   the shipped default (`127.0.0.1:11434`). Keep the destination address and
+   port joined with `and`; a port-only alternative permits that port on every
+   external host.
 
 ## Requirements & caveats
 
--- a/deploy/falco/falco_rules.yaml
+++ b/deploy/falco/falco_rules.yaml
@@ -48,8 +48,12 @@
 # is not on the shipped default (11434). Loopback literals are intentional
 # (CyClaw is loopback-only by design), not debug code -- suppress the devskim
 # DS162092 heuristic on this line.
+# Keep address and port joined as one allowance. A port-only alternative would
+# suppress alerts to every external host listening on that port.
 - macro: cyclaw_expected_outbound
-  condition: '(fd.sip = "127.0.0.1" or fd.sip = "::1" or fd.sport = 11434)'  # DevSkim: ignore DS162092
+  condition: >  # DevSkim: ignore DS162092
+    ((fd.sip = "127.0.0.1" or fd.sip = "::1") and
+     fd.sport = 11434)
 
 # ---------------------------------------------------------------------------
 # Rules.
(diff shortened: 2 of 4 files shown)


cgfixit:
Attempted independent runtime validation and hit the same limitation the PR body already calls out: `docker` CLI is present in my environment but there's no daemon socket (`/var/run/docker.sock` doesn't exist), so I can't start the privileged `cyclaw-falco` sidecar here either. Confirming rather than duplicating — this still needs a Docker-capable host before merge, same as stated.

Statically confirmed both fixes are logically sound:
- `docker-compose.yml`'s `command:` now passes `-r /etc/falco/falco_rules.yaml -r /etc/falco/rules.d/cyclaw_rules.yaml` (was a single `-r` pointing only at the custom file, which suppresses `falco.yaml`'s `rules_files` per Falco's own docs) — loads defaults before the CyClaw rules, so the `spawned_process`/`open_write`/`outbound` macros the custom rules depend on are defined first.
- `deploy/falco/falco_rules.yaml`'s `cyclaw_expected_outbound` macro now AND-joins loopback address with port 11434 instead of OR-joining — the old `or fd.sport = 11434` allow-listed port 11434 on *any* destination, not just loopback Ollama traffic.

Both changes parse as valid YAML and the logic checks out against Falco's condition-language semantics; I just can't exercise the actual detection at runtime any more than the PR author could. Holding per the PR's own recommendation.

---
Note: this will require local full verification before merging or it introduces more risk than it's worth imo

cgfixit:
Superseded by draft **#763** (grok/security-falco-detection), recreated on current `origin/main` (includes post-#761 main).

Reason for new PR instead of rebasing this branch tip:
- This branch was diverged/behind main and the title/history mixed in #725 concerns.
- Fresh branch keeps a single clean commit + expanded operator comments for Docker modern eBPF Falco.

Please review/merge **#763** (and sibling index isolation **#762**, recreating #725). Closing this PR.

PR #701, opened 2026-07-29
[infra] - raise the conda OTel exporter floor so the conda lane solves reproducibly

cgfixit:
## Proposed changes

One line in `environment.yml`, conda-only, directly beside the existing `fastapi` exception and for the same class of reason: a conda-forge packaging constraint rather than a CyClaw choice.

**Read the "Further comments" section before merging — this PR does *not* pin `protobuf`, which is what was originally asked for. The full traceback showed that pinning protobuf would have been the wrong fix, and probably an unsolvable one.**

The conda lane (`ci (3.12)`) fails on a fresh solve with:

```
TypeError: Descriptors cannot be created directly.
...your generated code is out of date and must be regenerated with protoc >= 3.19.0
```

The traceback names the culprit, and it is **not** chromadb's own generated code:

```
chromadb/telemetry/opentelemetry/__init__.py
  -> opentelemetry/exporter/otlp/proto/grpc/exporter.py
     -> opentelemetry/proto/common/v1/common_pb2.py
        _descriptor.FieldDescriptor(...)
```

conda-forge's `chromadb=1.5.9` declares only `opentelemetry-exporter-otlp-proto-grpc >=1.2.0` — a 2022-era floor — and declares **no protobuf constraint at all**. The solver is therefore free to pair a very old exporter with a current protobuf, and that exporter's shipped `_pb2` files predate protoc 3.19, so they refuse to load on a protobuf 4+ runtime. That fails `import chromadb` outright and reds the whole lane.

- Affects **none** of the six invariants or I6 module isolation. No Python source, no `config.yaml`, no test, no graph edge, no auth or soul path — one dependency line in a conda environment spec.
- Evidence: `invariant-guard` **28 passed, 0 failed**; `config-guard --strict` 0 failures; `dep-guard` 0 failures/0 warnings; `doc-sync` 0 drift.

---

## Types of changes

**Optional free-text scope note:** Infrastructure / CI only; conda environment spec.

---

## Benefits / why

- **Makes the conda lane reproducible instead of cache-dependent.** Until 2026-07-29 the failure was masked by a stale `actions/cache` entry in `python-package-conda.yml` holding a known-good environment. Any cache eviction — or any contributor doing a fresh `mamba env create` — hits it. The lane has been green by luck, not by construction.
- **Fixes the actual cause rather than the symptom.** The stale generated code belongs to the OTel exporter; raising its floor makes the solver take a build whose protos were regenerated with a modern protoc, which then works with whatever protobuf the rest of the solve already wants.
- **Leaves the working path alone.** The pip lanes resolve this correctly on their own — verified locally with `chromadb 1.5.9 + protobuf 7.35.1` from PyPI, where `tests/test_rag_integration.py::test_real_retriever_result_flows_through_graph` (the test that fails in conda) **passes**. This change does not touch `requirements.txt`, `constraints.txt`, or `pyproject.toml`.

---

## Risks to monitor

- **A floor, not an exact pin — deliberately.** OTel's `api`/`sdk`/`proto`/`exporter` packages move as a set, so pinning the exporter to one exact version invites conflicts as the others advance. `>=1.42` raises the floor above the broken range without freezing the group. If a future conda-forge exporter release regresses, this will not protect against it — the failure mode would be the same import error, and the fix would be a ceiling.
- **This constrains only the conda lane.** If the pip lanes ever start resolving an equally old exporter, they will hit the identical failure with nothing here to stop it. Watch for the same traceback on `ubuntu-latest`/`windows-latest`.
- **`chromadb=1.5.9`'s own loose floor is the underlying defect** and is upstream's, not ours. If chromadb is ever bumped here, re-check whether the exception is still needed — same instruction the `fastapi` comment directly above already carries.
- **Unverifiable from this session:** I cannot run a `mamba env create` here (no conda/mamba in the container), so the solve itself is unproven locally. The conda lane on this PR *is* the test. If it goes red, the pin is wrong and this should not merge.
- Rollback is a plain revert of one line.

---

## Checklist

- [ ] I have read the latest `docs/CyClaw Architecture Guide` (and any relevant Phase docs) and `SECURITY.md` — not reread for a one-line dependency change; `environment.yml`'s own fastapi-exception comment, `CLAUDE.md` §7 (dependency tier), and conda-forge's live repodata were the operative sources.
- [ ] Full sandbox validation has been run (`cyclaw-sandbox-validator` or equivalent pytest + smoke tests on core RAG/agentic paths) and passes with no regressions — not applicable; no Python source changed. The conda lane in CI is the meaningful validation and it runs on this PR.
- [x] No new external network dependencies or mandatory online LLM assumptions were introduced without explicit justification + offline fallback path — no new dependency; an existing transitive one gains a floor.
- [ ] For any agentic/fsconnect/harness change: two-phase audit, quota enforcement, governed delete/trash, and write guards have been verified — not applicable.
- [ ] Relevant architecture docs, threat model notes, or harness phase documentation have been updated if core behavior or topology changed — not applicable; `doc-sync` reports 0 drift. The rationale lives inline in `environment.yml` beside the fastapi precedent.
- [ ] For large or complex changes: before/after invariant matrix + sandbox evidence is included in "Further comments" or linked — not applicable; one line.

---

## Further comments

**Why this pins the OTel exporter and not `protobuf`, contrary to the original plan.**

The agreed fix was "pin `protobuf` in `environment.yml` only." Pulling the complete traceback changed that, and it is worth stating plainly because the earlier framing was mine and it was wrong.

1. **The stale generated code is `opentelemetry-proto`'s, not chromadb's.** That is why the PyPI install works while conda fails: PyPI resolves a modern exporter, conda's solve does not have to.
2. **Downgrading protobuf would fight the rest of the solve.** The error text suggests "downgrade the protobuf package to 3.20.x or lower", and conda-forge does carry `3.20.3` — so it would have looked like a legitimate pin. But `chromadb=1.5.9` also pulls `grpcio>=1.58.0` and `onnxruntime>=1.14.1`, whose conda-forge builds link a modern `libprotobuf`. Pinning protobuf back to 3.20.3 in 2026 is a good candidate for an unsolvable environment — i.e. it would have swapped one red lane for another, harder-to-read one.
3. **Raising the exporter floor is forward-compatible.** A modern exporter ships regenerated protos that work with the protobuf the solve already wants, so nothing else has to move.

If you would still rather name `protobuf` explicitly, it is a one-line swap — but I would want the conda lane on this PR to prove it first, and I would expect it to fail.

**Evidence gathered against conda-forge's live repodata (2026-07-29), not from memory:**

| package | versions available on conda-forge |
|---|---|
| `opentelemetry-exporter-otlp-proto-grpc` | 1.42.1, 1.43.0 |
| `opentelemetry-proto` | 1.39.1, 1.43.0, 1.44.0 |
| `protobuf` | 3.20.3, 4.25.3, 6.33.5, 7.35.1 |
| `chromadb` | 1.5.9 |

and `chromadb=1.5.9`'s own declared depends, which is where the loose floor lives:

```
grpcio >=1.58.0
onnxruntime >=1.14.1
opentelemetry-api >=1.2.0
opentelemetry-exporter-otlp-proto-grpc >=1.2.0     <- the 2022-era floor
opentelemetry-instrumentation-fastapi >=0.41b0
opentelemetry-sdk >=1.2.0
```

**Verification run**
- `yaml.safe_load(environment.yml)` — parses; 19 conda deps; new entry present and well-formed
- `python3 .claude/skills/dep-guard/check_deps.py` — 0 failures, 0 warnings (including D9 cross-manifest agreement, which correctly treats this as a conda-only exception like `fastapi`)
- `python3 .claude/skills/config-guard/check_config.py --strict` — 0 failures

**Environment limit (stated plainly):** this container has no conda/mamba, so I could not run the solve. Everything above is static verification plus live repodata lookups. The conda lane on this PR is the real test.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NWDaZNhapxSrAo8BADR4ww)_


cgfixit:
Note to self: in case of confusion later, no telemetry is instantiated here; only imported to satisfy something annoying with the conda ci check

PR #762, opened 2026-08-03
fix(docker): isolate generated index state from image layers

cgfixit:
## Proposed changes

- Exclude the configured `index/` root from the Docker build context so local ChromaDB/BM25 state cannot be copied into image layers.
- Bind-mount `./index` at `/app/index:rw` so Compose can use and persist the hybrid index while the container root filesystem remains read-only.
- Align the threat model writable carve-outs list and add a config-derived regression test for both sides of the contract.

**Recreates closed PR #725** on current `origin/main` (post Telegram skeleton merge @ 5e5490c).

- No core path, graph topology, model routing, auth, audit, soul, or module-isolation behavior changes.
- Infrastructure / container runtime only.

---

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [x] Documentation Update

---

## Benefits / why

- A locally generated index can contain private corpus-derived state. `.gitignore` excluded it from Git, but `.dockerignore` did not, and `Dockerfile` copies the remaining context.
- A clean image has no index, while Compose's read-only root previously supplied no writable `/app/index` mount. The same defect either baked stale/private state into the image or left `/query` unusable under RO root.
- The regression test derives the active index root from `config.yaml`, so future path changes must update Docker isolation and persistence together.

---

## Risks to monitor

- Host `./index` must remain readable/writable by the container UID/GID (1000:1000) on Linux.
- Docker image build / Compose runtime not exercised on the Windows verification host — treat as CI/runtime follow-up.
- Sibling draft PR (Falco detection, recreating #679) also edits `docker-compose.yml` in a different region; trial merge of both branches was clean.

---

## Validation (this host)

- `python -m pytest tests/test_security.py -q` — passed
- focused test `test_generated_index_is_not_baked_into_the_image_and_is_writable_at_runtime` — passed
- Compose YAML parse — passed
- `ruff check --select E,F,I,B,C4,UP,S tests/test_security.py` — passed
- Trial merge with `grok/security-falco-detection` — clean; both focused tests green
- Full Docker image build / `docker compose up` — **not run** (no Docker runtime verification on this host)

## Checklist

- [x] Preserves all 6 security invariants and I6 module isolation
- [x] Threat-model documentation updated
- [x] Draft PR only — human merges

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -30,9 +30,13 @@ env/
 **/.hf_cache/
 **/.st_cache/
 
-# Local runtime state — provided via bind mounts at runtime, not baked in
+# Local runtime state — provided via bind mounts at runtime, not baked in.
+# `index/` is the config.yaml hybrid-retrieval root (ChromaDB + BM25). It can
+# hold private corpus-derived vectors/keyword state; never copy it into image
+# layers. Compose bind-mounts ./index at runtime (see docker-compose.yml).
 logs/
 checkpoints/
+index/
 data/chroma/
 data/bm25/
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,6 +17,11 @@ services:
       - "127.0.0.1:8787:8787"
     volumes:
       - ./data:/app/data:rw
+      # Generated ChromaDB + BM25 state (config.yaml indexing.chroma_path /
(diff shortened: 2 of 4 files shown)


cgfixit:
@codex review

PR #765, opened 2026-08-03
security(agentic): match protected paths under Windows/macOS name aliases

cgfixit:
## What
Normalize protected-path compares with casefold + trailing-dot/space (and Cf) stripping so Windows/macOS aliases cannot bypass `tests/`, `pyproject.toml`, etc.

## Why / benefit
Writer-side `.git` jail already uses these rules; the real-repo protected list did not. A planner proposing `Tests/...` or `pyproject.toml.` could miss the gate on a case-insensitive FS.

## Risk to monitor
Slightly broader matching is intentional (platform-independent). Paths that only differ by case now collide as the same protected target — desired.

## Validation
- Focused pytest on the three `_matches_protected_path` tests — passed
- `ruff check` on `agentic/real_repo_loop.py` — passed

## Discipline
cyclaw-optimize @ be9c9b6; karpathy surgical; ponytail (reuse existing `.git` equivalence rules, no new abstraction layer).

### DIFF
--- a/agentic/real_repo_loop.py
+++ b/agentic/real_repo_loop.py
@@ -88,6 +88,7 @@ from __future__ import annotations
 
 import hashlib
 import re
+import unicodedata
 from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import Literal, Protocol, runtime_checkable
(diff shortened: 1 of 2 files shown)


cgfixit:
@codex review

cgfixit:
Noting in case Mac home folder agentic harness issues

PR #766, opened 2026-08-03
fix(telegram): do not ack updates when handle/send fails

cgfixit:
## What
Only advance Telegram `getUpdates` offset after a successful handle or a terminal `TelegramRefused`. Do not ack on `TelegramRuntimeError` (transport/query/send failure).

## Why / benefit
Previously offset advanced as soon as the update was seen. A failed `/query` or `sendMessage` permanently lost the user message because Telegram does not redeliver past the acked offset.

## Risk to monitor
A permanent RuntimeError loop could redeliver the same update until fixed — preferred over silent drop. Policy refusals (allowlist/mode/rate-limit) still ack so the stream does not stall.

## Validation
- `GROK_API_KEY=dummy python -m pytest tests/test_telegram_runner.py -q` — passed
- `ruff check` on runner + tests — passed

## Out of scope (ponytail)
Persistent `data/telegram/offset.json` across process restarts remains a design follow-up (TELEGRAM_DESIGN.md); not bundled here.

## Discipline
cyclaw-optimize @ be9c9b6; karpathy surgical; ponytail (one correctness fix, no durability feature creep).

### DIFF
--- a/telegram/runner.py
+++ b/telegram/runner.py
@@ -146,17 +146,20 @@ def poll_once(
         if not isinstance(upd, dict):
             continue
         uid = upd.get("update_id")
-        if isinstance(uid, int):
-            next_offset = uid + 1
         parsed = _message_from_update(upd)
         if parsed is None:
+            # Non-text updates (edits, stickers, …): ack so the stream advances.
+            if isinstance(uid, int):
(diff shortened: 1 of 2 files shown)


cgfixit:
@codex review

PR #769, opened 2026-08-03
fix(sync): soft-fail post-check timeout; reindex after partial-copy errors

cgfixit:
## What
- `run_post_sync_check`: timeout / subprocess launch failure → `CheckResult(ok=False)` (no raise)
- Exception path: attach `corpus_changed` on `SyncError.details`
- CLI: if reindex_on_change and corpus_changed after SyncError → exit 10 (or auto_reindex)

## Why
Post-check timeout after a successful rclone copy previously raised, dropping per-file audits and forcing EXIT_FAIL with no reindex signal → silent stale index.

## Risk
Exit 10 after exceptional partial copy is intentional for operators/schedulers that already treat 10 as reindex.

## Verify
`pytest tests/test_sync_runner.py tests/test_sync_cli.py -q` green; ruff clean.

### DIFF
--- a/sync/cli.py
+++ b/sync/cli.py
@@ -233,6 +233,17 @@ def cmd_sync(args: argparse.Namespace) -> int:
         return EXIT_ENV
     except SyncError as exc:
         _err(f"Sync error: {exc.message}")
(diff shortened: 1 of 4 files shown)


cgfixit:
@codex review

PR #774, opened 2026-08-03
ci: kill ghost environment.yml workflow; pin Miniforge in Conda CI

cgfixit:
## Analysis

| Check | Reality |
|---|---|
| **CyClaw Conda CI** (`python-package-conda.yml`) | **Green on main** (`544549f`, and recent PRs) |
| **Ghost** `.github/workflows/environment.yml` | Still **active** on GitHub after c708148 moved the file to repo-root `environment.yml` |

The ghost workflow produced **empty failing runs** on main (no jobs, no logs), e.g. [run 30727924433](https://github.com/cgfixit/CyClaw/actions/runs/30727924433) — classic “workflow file deleted/moved but Actions still tracks the path.”

## What

1. **Reclaim path** with a real `.github/workflows/environment.yml` that only runs on `workflow_dispatch` (never on push/PR) and documents the relocation.
2. **Real Conda CI**: pin `miniforge-version: 25.3.0-3` (was `latest`) and bump env cache generation **v2 → v3**.
3. **Immediate ops**: disabled the ghost via API (`disabled_manually`) so main stops enqueuing empty failures before merge.

## Why / benefit

Main branch status no longer polluted by a phantom conda-env-as-workflow failure. Conda installer base is reproducible.

## Risk to monitor

- First post-merge Conda CI run cold-solves with miniforge 25.3.0-3 + cache v3 (should match green runs from draft #771 which already used this pin).
- If anyone re-enables the ghost without the stub file, failures return — the no-op file is the durable fix.

## Verify

- `yaml.safe_load` on both workflow files
- Ghost workflow state: `disabled_manually` (API)
- Real Conda CI already green on `origin/main`

### DIFF
--- a/.github/workflows/environment.yml
+++ b/.github/workflows/environment.yml
@@ -1,0 +1,26 @@
+# Relocated: the conda environment *spec* lives at /environment.yml (repo root).
+# This path previously held that file, so GitHub registered it as a workflow.
+# After c708148 moved the file, GitHub kept a ghost "active" workflow that
+# produced empty failing runs on main pushes (no jobs, no logs).
+#
+# This no-op file reclaims the path with a real workflow that never auto-runs
+# (workflow_dispatch only). Conda CI remains:
+#   .github/workflows/python-package-conda.yml  → uses root environment.yml
+name: "environment.yml (relocated — no-op)"
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  noop:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


cgfixit:
@codex review thoroughly to ensure this is wise (I only use the mini mamba conda ci for potential ml/dataScientist users since they like that haha -- I could take or leave this ci due to ci.yml; but I didn't put this conda check here for my use case so consider that and close the pr yourself if it shouldn't be merged (I'll fact check you anyway))

PR #778, opened 2026-08-03
fix(telegram): redeliver failed poll updates; reuse httpx client

cgfixit:
## What

Two focused changes to the Telegram channel (`telegram/`):

1. **Fix message loss in `poll_once`** (`telegram/runner.py`): on `TelegramRuntimeError` the batch loop used to `continue`. A later successful update in the same `getUpdates` batch then set `next_offset` past the failed `update_id`, and Telegram never redelivered it — the message was silently dropped while appearing "handled" (with an error entry) in the results. The loop now `break`s: the failed update and everything after it in the batch are re-fetched on the next poll. `TelegramRefused` handling is unchanged (terminal policy refusals still ack).

2. **Persistent pooled httpx client** (`telegram/client.py`): `send_message`, `get_updates`, `post_query`, and `fetch_loopback_health` each constructed a fresh `httpx.Client` per call. They now share one lazily-created module-level client (same pooled pattern as `llm/client.py`'s persistent clients), with the exact same timeouts passed per-request instead of per-client. Adds `reset_http_client_for_tests()` so tests patching `httpx.Client` don't leak a cached mock between cases; existing client tests were updated mechanically (mock target moves off `__enter__`), no test logic changed.

## Why / benefit

- **Correctness (message loss):** a transient `/query` or `sendMessage` failure mid-batch was silently acked whenever a later update in the same batch succeeded. Telegram's offset semantics mean an acked update is gone forever — the user message vanished with no retry. Regression test added.
- **Performance:** the T2 chat loop long-polls forever; per-call client construction paid a full TCP+TLS handshake to `api.telegram.org` on every poll cycle and every reply. Pooling reuses keep-alive connections.

## Risk to monitor

- The `break` changes batch semantics: the remainder of a batch is re-fetched next poll. Duplicates are possible only for updates *before* the failed one — impossible, since those already advanced `next_offset` past themselves — and for the failed update's own outbound side effects if `sendMessage` succeeded server-side but the response was lost (Telegram Bot API has no idempotency key; this is the standard at-least-once trade-off and strictly better than the previous silent drop).
- Pooled client is process-global; it is never closed except via `reset_http_client_for_tests()`. Thread-safety matches `llm/client.py`'s existing pooled usage (httpx.Client is thread-safe).
- Test coverage added: mixed-batch (fail-then-succeed) asserts the failed update stays un-acked and the trailing update is not processed until redelivery.

## Verification

From the worktree root (`C:/Users/cgrady/CyClaw-wt-telegram-poll`, system Python 3.12):

```
GROK_API_KEY=dummy python -m pytest tests/test_telegram_runner.py tests/test_telegram_client.py -q --tb=short
# 25 passed, exit code 0

ruff check telegram/ tests/test_telegram_runner.py tests/test_telegram_client.py
# All checks passed!

ruff check --select E,F,I,B,C4,UP,S telegram/ tests/test_telegram_runner.py tests/test_telegram_client.py
# All checks passed!  (CI-enforced rule set)
```

### DIFF
--- a/telegram/client.py
+++ b/telegram/client.py
@@ -17,6 +17,27 @@ from telegram.config import TelegramConfig
 from utils.errors import TelegramRefused, TelegramRuntimeError
 from utils.logger import audit_log
 
+# Pooled module-level client (same pattern as llm/client.py's persistent
+# clients): a forever-poll loop otherwise pays a fresh TCP+TLS handshake per
+# Bot API call. Lazily created; every call still passes its own per-request
+# timeout, so per-call timeout semantics are unchanged.
+_http_client: httpx.Client | None = None
+
+
+def _get_http_client() -> httpx.Client:
+    global _http_client
+    if _http_client is None:
+        _http_client = httpx.Client()
+    return _http_client
+
+
+def reset_http_client_for_tests() -> None:
+    """Drop the pooled client (unit tests only — it can hold a patched mock)."""
+    global _http_client
+    if _http_client is not None:
+        _http_client.close()
+        _http_client = None
+
 
 def _hash_token_fingerprint(token: str) -> str:
     """Short non-reversible fingerprint for audit (never the token itself)."""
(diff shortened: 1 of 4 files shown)


cgfixit:
Noting the httpx removal by Kimi here - no issues but I need to verify it's not a security vuln potentially

PR #780, opened 2026-08-03
feat(agentic): configurable committer identity/branch prefix (defaults unchanged) — governance decision requested

cgfixit:
## What

Makes the agent committer identity and branch prefix configurable via environment variables in `agentic/deepagent_github/repo_workspace.py`, **without changing any current default**:

- `CYCLAW_AGENT_COMMIT_NAME` (default `Claude`)
- `CYCLAW_AGENT_COMMIT_EMAIL` (default `noreply@anthropic.com`)
- `CYCLAW_AGENT_BRANCH_PREFIX` (default `claude`; validated as 1–32 chars from `[A-Za-z0-9._-]`, alphanumeric-led — the shape that preserves the existing "no branch name can start with `-`" argv-injection anchor; invalid values fail closed at import)

Also env-aware with unchanged defaults: the SessionStart identity pin (`.claude/hooks/session-start-sync-check.sh`) and the CyClaw-Optimize bootstrap pin. `CLAUDE.md` §10 documents the knobs and their caveats.

## Why

The commit identity is stamped `Claude <noreply@anthropic.com>` regardless of what actually drives the loop — by default a **local Qwen model**, and the operator's own conventions also use `kimi/*` and `codex/*` branches. Attribution should be able to name the real driver. Today the only way to do that is editing source. This PR adds the knob; it deliberately changes no default.

## Decision requested

Owner decision explicitly requested:

1. **Keep `claude/` + `Claude <noreply@anthropic.com>` as the long-term default** (this PR only adds the knob — merge as-is), or
2. **Switch the default to a generic identity** (e.g. `CyClaw Agent`) in a follow-up — this PR does NOT do that.

Note the scope boundary either way: the branch-pattern mirrors in `agentic/writer.py` (`_HEAD_BRANCH_RE`) and `harness/agent_policy.py` (`BRANCH_NAME_RE`, I6-forbidden from importing `agentic`) stay pinned to `claude/` by design, so an overridden prefix is honored only by the repo-workspace surface. Syncing those mirrors (or deciding they SHOULD stay claude-only) is part of the same governance decision.

## Risk to monitor

- **Hook enforcement**: the stop hook that rejects non-`noreply@anthropic.com` committer emails is applied by the *session runtime*, not this repo — overriding `CYCLAW_AGENT_COMMIT_EMAIL` for session commits requires a matching runtime change, or commits get rejected. The in-repo hooks now read the same env vars (defaults unchanged) so they follow whatever the operator configures.
- **Audit attribution**: audit events (`agentic_repo_workspace_git_op`) do not record the committer identity — attribution lives only in the git history of the pushed branch. If the identity becomes per-driver configurable, consider whether the audit log should capture it (follow-up, not this PR).
- **Env-prefix validation** fails closed at import (`ValueError`); a typo'd prefix bricks the module import rather than silently reverting — deliberate, but worth knowing.

## Verification

- `GROK_API_KEY=dummy python -m pytest tests/test_agentic_repo_workspace.py -q --tb=short` — **exit 0** with `-k "not symlink"` (109 passed). The 5 excluded `*symlink*` tests fail identically on pristine `origin/main` on this machine (`WinError 1314`, symlink privilege) — pre-existing environmental, unrelated.
  - New tests: default constants byte-identical to the historical convention; env override reconfigures identity + prefix (commit log shows `Kimi <kimi-agent@example.com>`, `kimi/topic` accepted, `claude/*` rejected); 5 parametrized invalid prefixes fail closed at import.
- `pytest tests/test_agentic_writer.py tests/test_harness_agent_routes.py` — exit 0 (cross-module pattern-equality drift tests still green with the default).
- `pytest tests/test_agentic_real_repo_loop.py tests/test_agentic_real_repo_run_cli.py` — exit 0 (both assert the default `Claude <noreply@anthropic.com>` identity end-to-end).
- `ruff check --select E,F,I,B,C4,UP,S agentic/ tests/` — clean; `bash -n` on both edited hook scripts — OK.

### DIFF
--- a/.claude/hooks/session-start-sync-check.sh
+++ b/.claude/hooks/session-start-sync-check.sh
@@ -4,6 +4,9 @@
 # Purpose (proactive, NON-destructive):
 #   1. Pin the commit identity to noreply@anthropic.com / Claude so commits are
 #      not flagged "Unverified" by the stop-hook (recurring friction otherwise).
+#      Both values are overridable via CYCLAW_AGENT_COMMIT_EMAIL /
+#      CYCLAW_AGENT_COMMIT_NAME (the same knobs agentic/deepagent_github/
+#      repo_workspace.py reads); the defaults preserve the pinned convention.
 #   2. Fetch the default branch and REPORT divergence between local and remote.
 #      It never resets, rebases, pushes, or deletes — it only informs, so a
 #      human stays in control of how to reconcile.
@@ -15,8 +18,8 @@ repo_root=$(git rev-parse --show-toplevel 2>/dev/null) || exit 0
 cd "$repo_root" || exit 0
 
 # ── 1. Pin commit identity (repo-local, durable) ─────────────────────────────
-git config --local user.email "noreply@anthropic.com"
-git config --local user.name  "Claude"
+git config --local user.email "${CYCLAW_AGENT_COMMIT_EMAIL:-noreply@anthropic.com}"
+git config --local user.name  "${CYCLAW_AGENT_COMMIT_NAME:-Claude}"
 
 # ── 2. Detect default branch (origin/HEAD, fallback main) ────────────────────
 default_branch=$(git symbolic-ref --quiet --short refs/remotes/origin/HEAD 2>/dev/null | sed 's@^origin/@@')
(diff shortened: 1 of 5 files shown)


cgfixit:
Noting since that seems strange and lazy of me to allow all pr identity to be Claude - I mean doesn't really matter if the branches are named right and for some reason i only have issues with grok build and kimi code  following branch name but not pr format instructions. It may just be a pre flight / hook at the beginning of a GitHub coding task to review/load the .github pr_template.md -- either beginning or as the last of the verification end process pre commit/pr draft - I should just make a skill for pre and post pr and whatever stuff got the powershell/terminal coding agents.. if I use Kimi the app or Claude code or codex no issues at all so seems ghetto-righable at the least that might actually be the right way anyways leaving comment to make it easier to find later if I'm searching recent pr merge commits 👋✌️

PR #794, opened 2026-08-05
[security] - drop CSP script-src unsafe-inline; fix extractor.html escaping gap

cgfixit:
## Proposed changes

Part of a 5-PR CyClaw-Optimize sweep. `static/terminal.html` carried ~34 inline `onclick`/`oninput`/`onchange` attribute handlers, which only work because `gate.py`'s CSP set `script-src 'self' 'unsafe-inline'`. That's a real CSP-strength gap: `terminal.html`'s two `innerHTML` sinks (`addEntry`/`addSources`) are both correctly routed through `escHtml()` today, but `unsafe-inline` means that mitigation has no backstop — if either sink were ever bypassed, injected `<script>`/`onerror=` would execute.

**Changes:**
- Converted every inline event-handler attribute in `terminal.html` to `addEventListener`, adding a stable `id` to each of the 24 elements that lacked one.
- Dropped `'unsafe-inline'` from `gate.py`'s CSP `script-src` directive (`style-src` untouched — unrelated, CSS not JS).
- Fixed an inconsistent escaping gap in `extractor.html`: the matched-keywords line was interpolated into `innerHTML` unescaped while the adjacent insight text was correctly escaped via `escapeHtml()`. Currently inert (keywords come from a hardcoded in-file array, never parsed content) but one edit away from exploitable.
- Added a regression test (`tests/test_gate.py::test_csp_script_src_has_no_unsafe_inline`) locking the `script-src` change in place.

**Invariant / Governance Impact:** none of the six invariants. This only touches response headers and static-console JS wiring — no graph edge, retrieval logic, or auth path changed. `invariant-guard` re-run clean (33/33) before and after.

## Types of changes
- [x] Bugfix (non-breaking — see verification note below)
- [x] Invariant / Governance refinement (CSP hardening; strengthens a defense-in-depth control, doesn't touch the 6 core invariants)

**Scope note:** RAG retrieval/sanitization is untouched — this is the static console + `gate.py`'s security-headers middleware only.

## Benefits / why
Closes a real gap between "the escaping mitigations we already have" and "what the CSP would actually stop if one of them failed." No behavior change for a legitimate user — every button/input still does exactly what it did before.

## Risks to monitor
This is the kind of change that's easy to get subtly wrong (a missed handler, a mismatched id), so it got real browser verification, not just a static diff read: served `static/` directly and drove it with Playwright against the repo's pinned Chromium. Page loads with zero JS errors, and clicking each of the five panel-toggle buttons plus Load Soul / Sync Status / Agentic Status / FS Status / SQL Status / Send correctly fires a request to the expected endpoint end-to-end.

That pass caught a genuine bug the mechanical conversion introduced: `addEventListener` passes the click's `PointerEvent` as the handler's first argument, and `submitQuery(confirmedOnline, onlineProvider)` treats any non-null first argument as "user already confirmed" — a bare `addEventListener('click', submitQuery)` binding silently skipped the plain-query path (the button appeared to do nothing on click). Fixed by wrapping that one binding as `() => submitQuery()`. I individually checked all 34 converted handlers' function signatures — every other one takes zero parameters, so the extra event argument is harmless there; this was the one real case.

If a future edit reintroduces an inline handler on `terminal.html` (or `harness.html`/`extractor.html`, which already had none), it will silently stop working under this CSP rather than failing loudly — worth a glance if a console button ever "does nothing" after a future change.

## Checklist
- [x] I have read the latest architecture docs and threat model notes relevant to this change
- [x] This change preserves all 6 security invariants and I6 module isolation (no core-path change; invariant-guard re-run clean)
- [x] Verified in a real browser (Playwright), not just diff-read — see Risks section
- [x] No new external network dependencies or online LLM assumptions introduced

## Further comments
ELI5: some buttons in the web console used old-style `onclick="doThing()"` HTML, which only works if the page's security policy allows "run inline scripts" (`unsafe-inline`). That's a bigger door than it needs to be — if anything else ever slips through and injects a `<script>` tag, that open door lets it run. We moved every button to the modern "attach the click handler in JS" style so we could close that door. Verified button-by-button in an actual browser rather than trusting the diff, which is how the one real regression (Send button silently doing nothing) got caught and fixed before this ever reached review.

---

### DIFF
--- a/gate.py
+++ b/gate.py
@@ -423,7 +423,7 @@ class _SecurityHeadersMiddleware(BaseHTTPMiddleware):
         response.headers.setdefault("X-Permitted-Cross-Domain-Policies", "none")
         response.headers.setdefault(
             "Content-Security-Policy",
-            "default-src 'none'; script-src 'self' 'unsafe-inline'; "
+            "default-src 'none'; script-src 'self'; "
             "style-src 'self' 'unsafe-inline'; img-src 'self' data:; "
             "connect-src 'self'; font-src 'self'; base-uri 'none'; "
             "form-action 'none'; frame-ancestors 'none'"
--- a/static/extractor.html
+++ b/static/extractor.html
@@ -536,7 +536,7 @@
                             <span class="insight-number">Insight #${idx + 1}</span>
                             <span class="insight-score">Relevance: ${insight.score.toFixed(2)}</span>
                         </div>
-                        <div class="insight-keywords">Matched concepts: ${insight.keywords.join(', ')}</div>
+                        <div class="insight-keywords">Matched concepts: ${escapeHtml(insight.keywords.join(', '))}</div>
                         <div class="insight-text">${escapeHtml(insight.text)}</div>
                     `;
                     insightsList.appendChild(item);
--- a/static/terminal.html
+++ b/static/terminal.html
@@ -813,11 +813,11 @@
 <!-- MAIN -->
 <div class="main">
   <div class="soul-toolbar">
-    <button class="toolbar-btn" id="soulToggleBtn" onclick="toggleSoulPanel()" aria-label="Toggle Soul Console (Escape to close)">Soul Console</button>
-    <button class="toolbar-btn" id="syncToggleBtn" onclick="toggleSyncPanel()" aria-label="Toggle Sync Console (Escape to close)">Sync Console</button>
-    <button class="toolbar-btn" id="agenticToggleBtn" onclick="toggleAgenticPanel()" aria-label="Toggle Agentic Console (Escape to close)">Agentic Console</button>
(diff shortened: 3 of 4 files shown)


cgfixit:
@claude no action required in this comment for you but nice find! I totally forgot I put that in there.. (you built that or your older cousin sonnet 4.5 did last year ;)

cgfixit:
The `@claude` job above errored out, so here's the actual answer since you asked: checked `git blame` and it wasn't a "cousin" model at all — both the `unsafe-inline` CSP line (`gate.py:426`) and the original `onclick="..."` handlers in `terminal.html` trace to commit `7fc9390`, authored by **you** (Chris Grady), 2026-08-03. No AI fingerprints on this one. 😄


---

^^ haha you a lie!

PR #796, opened 2026-08-05
[governance] - document check_jailbreak's offline no-op + pin both inert rails

cgfixit:
## Proposed changes

Part of a 5-PR CyClaw-Optimize sweep. `config.yaml`'s `guardrails.input_rails` lists `check_jailbreak` with no caveat, but `guardrails/integration.py`'s `_offline_checks()` only ever implements `check_injection` and `check_soul_mutation` — `check_jailbreak` is model-assisted only (`guardrails/config/rails.co`'s Colang flow) and is a silent no-op on the offline floor. `check_soul_leak` already carries an explicit "NOT YET FULLY IMPLEMENTED" comment in `config.yaml` for the exact same shape (listed, not enforced offline); `check_jailbreak` didn't.

**Changes:**
- Added a comment to `config.yaml`'s `check_jailbreak` entry matching the existing `check_soul_leak` one.
- Added a regression test for each rail in `tests/test_guardrails_integration.py`: confirms both are present in the default `input_rails`/`output_rails` lists, and confirms neither ever appears in a `check_input`/`check_output` response's `"rails"` list regardless of input — so a future edit to `_offline_checks` can't silently start (or silently keep not) enforcing either without a test failing either way.

**Invariant / Governance Impact:** none of the six invariants. Doc comment + regression tests only — `guardrails/` stays soft-imported, opt-in, and reached only via `utils/guardrail_bridge.py` per I6. `invariant-guard` and `config-guard` both re-run clean.

## Types of changes
- [x] Documentation Update
- [x] Invariant / Governance refinement (closes a documentation asymmetry between two rails with identical enforcement gaps)

**Scope note:** Core graph/gate/soul path untouched — this is `config.yaml`'s guardrails block (a comment only) plus test coverage.

## Benefits / why
Before this, an operator reading `config.yaml` would reasonably assume `check_jailbreak` is enforced (it's just listed, no caveat) while `check_soul_leak` right below it explicitly says it isn't — an asymmetry that undersells the real gap. Now both are equally honest, and there's a test that will fail loudly if either rail's offline-enforcement status ever silently changes.

## Risks to monitor
None — doc comment plus additive tests, zero behavior change. If someone *does* implement offline enforcement for either rail in the future, these two new tests will (correctly) start failing — that's the intended tripwire, not a regression.

**Note on scope:** I found a third opportunity in the same review pass — `utils/guardrail_bridge.py`'s `build_input_guard`/`build_output_guard` each independently call `load_guardrails_config()` and construct their own `GuardrailMetrics` when both are enabled. I did not pursue it here: `GuardrailMetrics` persists every event to the same on-disk `logs/guardrails.jsonl` regardless of instance count, so the only real effect of two instances is split *in-memory* counters that nothing currently reads. A safe fix needs an optional shared-state parameter to avoid breaking the 8+ existing `test_guardrail_bridge.py` tests that rely on independent, monkeypatchable calls to `load_guardrails_config()` — more surface than the (currently invisible) benefit justifies. Flagging here in case it's worth a future look if that assumption ever changes (e.g. something starts reading the in-memory counters).

## Checklist
- [x] This change preserves all 6 security invariants and I6 module isolation (invariant-guard + config-guard both re-run clean)
- [x] Full sandbox validation: `ruff check` clean, and `test_guardrails_integration.py`/`test_guardrails_config.py`/`test_guardrail_bridge.py`/`test_guardrails_isolation.py` all pass
- [x] No new external network dependencies or online LLM assumptions introduced

---

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -699,7 +699,13 @@ guardrails:
   hallucination_threshold: 0.18  # [0..1] token-overlap floor below which an answer is flagged ungrounded
   input_rails:                   # mirror guardrails/config/config.yml -> rails.input.flows
     - check_injection
-    - check_jailbreak
+    - check_jailbreak            # NOT ENFORCED BY THE OFFLINE FLOOR -- model-assisted only
+                                # (NeMo self_check_input prompt, guardrails/config/rails.co's
+                                # "check jailbreak" flow). guardrails/integration.py's
+                                # _offline_checks() only implements check_injection and
+                                # check_soul_mutation; listing it here activates it solely on
+                                # a live NeMo Colang run, same silent-skip shape as
+                                # check_soul_leak below.
     - check_soul_mutation
   output_rails:                  # mirror guardrails/config/config.yml -> rails.output.flows
     - check_grounding
--- a/tests/test_guardrails_integration.py
+++ b/tests/test_guardrails_integration.py
@@ -409,6 +409,28 @@ class TestCheckInput:
         res = check_input("a completely benign local question", cfg=cfg, metrics=m)
         assert res["blocked"] is False
 
+    def test_check_jailbreak_is_configured_but_not_offline_enforced(self):
+        """check_jailbreak is listed in DEFAULT_INPUT_RAILS / config.yaml's
+        input_rails (rails.co's model-assisted "check jailbreak" flow), but
+        _offline_checks() only implements check_injection and
+        check_soul_mutation -- it is a no-op on the offline floor today. This
+        pins that gap so a future edit to _offline_checks can't silently start
+        (or silently keep not) enforcing it without a test noticing either way.
+        """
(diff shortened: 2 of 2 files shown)


cgfixit:
Leaving this for @codex's independent read, but flagging what I found in case it saves a pass: the `verify-skills (index-doctor)` failure was `HTTP 429` from `huggingface.co` while fetching the `all-MiniLM-L6-v2` embeddings model (rate-limited, not a code path this PR touches — the diff is a `config.yaml` comment + two test functions in `test_guardrails_integration.py`, neither near retrieval/embeddings). Sibling PRs' same check passed in the same time window, and this run's `index-doctor` job has since completed successfully on its own — **all 43 checks are green now with zero code changes**, confirming it was the external flake, not something the test is exposing.


---

PR #799, opened 2026-08-05
fix(telegram): private-chat only for T3 /online hybrid consent

### LINKED ISSUE #792: potential for a multi user environment to encounter lets one participant send /online on grok and the next participant's message in the same group be forwarded with user_confirmed_online=True, which is external fallback without that sender's confirmation; bind the session to the sender or refuse /online outside private chats.
**Status: Unverified but high priority to investigate:**

> High level Description: `allow_hybrid_confirm: true`, this grants T3 consent for the entire `chat_id`, while the later claim also keys only on `chat_id` and `_message_from_update` discards Telegram's per-sender `from.id`. That lets one participant send `/online on grok` and the next participant's message in the same group be forwarded with `user_confirmed_online=True`, which is external fallback without that sender's confirmation; bind the session to the sender or refuse `/online` outside private chats.

1)
### 💡 Codex Review of pr # 789

https://github.com/cgfixit/CyClaw/blob/e3c13221a1edb1ca74d2501502e05f62020ef05b/telegram/runner.py#L115-L118
**<sub><sub>P1 Badge</sub></sub>  Bind online consent to the Telegram sender**

When an allowlisted `group`/`supergroup` chat is used with `allow_hybrid_confirm: true`, this grants T3 consent for the entire `chat_id`, while the later claim also keys only on `chat_id` and `_message_from_update` discards Telegram's per-sender `from.id`. That lets one participant send `/online on grok` and the next participant's message in the same group be forwarded with `user_confirmed_online=True`, which is external fallback without that sender's confirmation; bind the session to the sender or refuse `/online` outside private chats.

AGENTS.md reference: [AGENTS.md:L25-L27](https://github.com/cgfixit/CyClaw/blob/e3c13221a1edb1ca74d2501502e05f62020ef05b/AGENTS.md#L25-L27)

---

https://github.com/cgfixit/CyClaw/blob/e3c13221a1edb1ca74d2501502e05f62020ef05b/telegram/media.py#L421
**<sub><sub>P2 Badge</sub></sub>  Audit media download refusals before acknowledging**

When Telegram omits or understates the file size and `download_file()` rejects the stream after this call, that `TelegramRefused` bubbles to `poll_once`, which treats non-rate-limit refusals as terminal and advances the update offset. The audit trail only has `telegram_media_stage_requested`; it never records a `telegram_media_refused`/failed terminal event showing that the write did not happen, so this security-sensitive path is acknowledged without a final audit record. Catch refusals around the download/write phase and log the gate before re-raising.

AGENTS.md reference: [AGENTS.md:L25-L27](https://github.com/cgfixit/CyClaw/blob/e3c13221a1edb1ca74d2501502e05f62020ef05b/AGENTS.md#L25-L27)
    

<details> <summary>ℹ️ About Codex in GitHub</summary>
<br/>

[Your team has set up Codex to review pull requests in this repo](https://chatgpt.com/codex/cloud/settings/general). Reviews are triggered when you
- Open a pull request for review
- Mark a draft as ready
- Comment "@codex review".

If Codex has suggestions, it will comment; otherwise it will react with 👍.

Codex can also answer questions or update the PR. Try commenting "@codex address that feedback".
            
</details>

**To Reproduce**

**Expected behavior**
Login at start of landing on terminal or harness and granular rbac on extended roadmap but in the meantime there's no situation where it's appropriate in a multi-user instance of CyClaw where two independent workers have a decision from one (choosing a vault miss on a query should be sent to grok API online) would automatically be decided for another user without the same choice. Depending on race conflicts and timing there's always a chance that type of issue can happen but before I legit do add rbac and individual accounts and proper locking mechanism of config flags or a mechanism to set a config flag per user to avoid conflicts like that especially if the toggle to send inline to an LLM api just stays on if it was ever chosen once until hardcoded back in config. That's exactly the kind of issue I wouldn't notice due to mostly using rag and Qwen other than verifying it works with send online so once I have 2 ais take a look this is higher priority than being lazy and sending a message to my local ai - priority is stabilizing what's there first.
^ignore the ramblings near the end but it did start with how I envisioned it before I realized I hadn't fully thought of multi user which that's fine but don't start vibe coding that at most the second 2/3 of this section should be a minor note in your scratchpad but not the main goal. The priority for you to find clarity in your cloned branch of origin main that will ultimately be drafted into a pr to merge into main -- focus  at the top of the issue when it's actually described by codex

cgfixit:
## What

Fixes **#792** (P1 Codex finding): T3 hybrid consent is **private-chat only**.

- `/online on <grok|claude>` refused outside `chat.type == private` (groups/supergroups/channels)
- Hybrid claim also skipped for non-private so stale chat-scoped session files cannot arm online for group peers
- Poll loop passes Telegram `chat.type` into the text handler
- Audit: `telegram_hybrid_confirm_refused` with `gate: private_chat_only`
- Tests + design doc note

## Why

Chat-scoped session state + discarded `from.id` meant one group member could confirm external fallback and the next message from another member could consume it. Product threat model is single-operator / private DM; multi-user group ACLs are a design non-goal. Binding sender ids would be a larger multi-tenant design — private-only is the surgical fix matching T4 media.

## Risk to monitor

- Operators who intentionally put the bot in a group and expected `/online` there will get a clear refusal. Use a 1:1 DM.
- Session files remain keyed by `chat_id` only (fine for private DMs where chat_id is the user).

## Verification

- `ruff check telegram/runner.py tests/test_telegram_runner.py --select E,F,I,B,C4,UP,S`
- `GROK_API_KEY=dummy pytest tests/test_telegram_runner.py -q`
- invariant-guard (supporting guards) clean

## Merge order

- Standalone; base `main`
- Does not depend on open drafts #795–#797

## Base

- GitHub base: `main`

Closes #792

### DIFF
--- a/docs/channels/TELEGRAM_DESIGN.md
+++ b/docs/channels/TELEGRAM_DESIGN.md
@@ -248,16 +248,21 @@ as T1.
 
 1. Only exact `/online on <grok|claude>` can grant consent. `/online on` and
    any ambiguous form return usage instead of silently choosing a provider.
-2. The grant stores only `confirm_until` and provider in
+2. **Private chat only** (issue #792). Groups/supergroups share one `chat_id`
+   across senders; T3 grant and claim both refuse non-`private` types so one
+   member cannot arm external fallback for another. Multi-user ACLs remain a
+   non-goal — use a 1:1 DM with the bot.
+3. The grant stores only `confirm_until` and provider in
    `data/telegram/session_{chat_id}.json`; chat ids are canonical signed 64-bit
    integers and per-session file locking plus atomic replacement protects the
    one-shot claim across poller processes.
-3. The next non-command text claims and deletes that state **before** the
-   `/query` call. It can set `user_confirmed_online: true` and the selected
-   `online_provider` only when unexpired and the master switch remains enabled.
-   A failed query still spends the claim; there is no sticky “always online”.
-4. Grant and consume are separately audited without message text or secrets.
-5. CyClaw's existing `mode=hybrid` and enabled-provider checks remain the final
+4. The next non-command **private** text claims and deletes that state
+   **before** the `/query` call. It can set `user_confirmed_online: true` and
+   the selected `online_provider` only when unexpired and the master switch
+   remains enabled. A failed query still spends the claim; there is no sticky
+   “always online”.
+5. Grant, refuse, and consume are audited without message text or secrets.
+6. CyClaw's existing `mode=hybrid` and enabled-provider checks remain the final
    authority. The Telegram bridge cannot make an offline configuration use an
    external provider.
 
--- a/telegram/runner.py
+++ b/telegram/runner.py
@@ -33,7 +33,8 @@ _HELP_TEXT = (
     "/help — this text\n"
     "/status — loopback CyClaw /health\n"
     "/id — your chat id\n"
-    "/online on <grok|claude> — one explicit external fallback for your next message\n"
+    "/online on <grok|claude> — one explicit external fallback for your next "
+    "private-chat message only\n"
     "Anything else → local RAG via POST /query (mode=chat).\n"
     "Media staging requires a private-chat photo/document captioned /save --confirm <reason>."
 )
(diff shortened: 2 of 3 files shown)


cgfixit:
Ohh nice now I see what that meant I thought it was saying it regarding online sticking true for next user query without option presented via HITL loop on terminal.html directly - which I was seeing that not be true but this was a good find as well no reason to rush that feature

PR #804, opened 2026-08-05
feat(agentic): configure proposer completion budget

cgfixit:
## Summary
Makes the existing real-repo proposer completion budget configurable. Default remains 2048; no new agent framework is introduced.

## Invariant / CG-Security Invariant Check (CyClaw)
- Preserves I6 module isolation and the existing default-off real-repo/cloud gates.
- Positive integer validation happens at the agentic config boundary.
- No core RAG, graph routing, telemetry, filesystem authority, or provider eligibility changed.

## Verification Steps before commit
- Python 3.12 focused config, plan-handoff, CLI, and real-loop pytest suite passed.
- Ruff passed; invariant guard passed (33/0); doc sync found 0 drift; staged diff check passed.
- Fresh-main throwaway trial merge passed without conflicts or unmerged files.
- Bandit is not installed in this environment; its result is not claimed.

## Scope and risks
- `planner_max_tokens` reaches only real-repo plan/coding calls; it does not alter the separate harness optimizer.
- Raising it can exceed the available Ollama context or the existing timeout; default 2048 preserves current behavior.
- Remote CI remains pending.

### DIFF
--- a/README.md
+++ b/README.md
@@ -934,6 +934,7 @@ agentic:
     workspace_root: "data/agentic/workspaces"
     max_write_budget_bytes: 100000
     max_handoff_chars: 200000           # outbound-prompt cap for cloud egress
+    planner_max_tokens: 2048             # real-repo completion cap; keep it within Ollama num_ctx
     allow_cloud_providers: false        # gate 3 of the cloud chain
     providers:
       grok:   { enabled: false, model: "grok-4.5" }
--- a/agentic/cli.py
+++ b/agentic/cli.py
@@ -616,6 +616,7 @@ def cmd_real_repo_run_plan(args: argparse.Namespace) -> int:
             client,
             instruction=args.instruction,
             context=context_text or "",
+            max_tokens=cfg.deepagent_github.planner_max_tokens,
             config_path=args.config,
             cfg=app_cfg,
         )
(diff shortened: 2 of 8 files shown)


cgfixit:
Nice fix — this closes the "no config knob" gap cleanly. One thing worth flagging before this merges, since the PR body already identifies raising the default as an open tradeoff without resolving it:

**The `2048` default itself is very likely too small for `real_repo_loop`'s actual output shape**, independent of whether it's configurable. The module's planner protocol is documented whole-file replacement (`real_repo_loop.py`'s own docstring: "the planner protocol is whole-file replacement"), and `_parse_file_blocks` requires a closing `=== END FILE ===` marker (the regex is `re.DOTALL` anchored on both markers). If a completion gets cut off mid-file at the 2048-token boundary — easy to hit for even one moderately-sized Python file, let alone a multi-file patch — the block regex simply won't match, and the iteration fails with no useful signal that "truncation" was the cause, burning one of `max_iterations` for free every time it happens.

Both proposer paths share this exposure identically: `LocalProposerClient.invoke()` (model_adapter.py) and `ChatModelProposerClient.invoke()` (chat_client.py) both default `max_tokens=2048` on the same `ProposerClient` Protocol, and neither of the two real call sites in `real_repo_loop.py` overrides it — this PR's own diff confirms that by threading the *same* `2048` through explicitly rather than changing it.

Not asking for a change in this PR — the config knob standing alone is a legitimate, well-scoped fix. Just flagging that "raise it after measuring" (per your new comment) may be closer to "raise it as a follow-up default-value change" than an edge case, since the failure mode is silent (a rejected iteration, not an error naming truncation as the cause). Happy to pick up a small follow-up PR bumping the default once this lands, if useful — didn't want to submit a competing edit to the same lines while this is still open.

---

---

PR #818, opened 2026-08-08
[security] - serve console JS as a file so strict CSP stops blocking the terminal console

cgfixit:
## Proposed changes

**The web console at `/` has been non-functional in any CSP-enforcing browser since 2026-08-04.** This restores it without giving back the hardening that broke it.

### What broke

`265557f` (#794, *"drop CSP script-src unsafe-inline"*) converted `terminal.html`'s ~20 inline `onclick="..."` attributes to `addEventListener`, then removed `'unsafe-inline'` from `script-src`.

Removing inline event-handler **attributes** does not make an inline `<script>` **block** legal. `script-src 'self'` permits same-origin script *files*; inline blocks require `'unsafe-inline'`, a nonce, or a hash. `terminal.html` kept all ~1000 lines of console code in a bare inline `<script>`, so the browser blocked the entire thing.

The page still *rendered* — `style-src` does carry `'unsafe-inline'`, so markup and CSS looked correct while nothing worked: no query submit, no health poll, no panel toggles.

**Verified against a running server, not inferred:**

```
$ curl -sD- http://127.0.0.1:8799/
content-security-policy: default-src 'none'; script-src 'self'; style-src 'self' 'unsafe-inline'; ...

inline <script> blocks: 1
external script src:    NONE — all JS is inline
```

The `<script>` tag carries no nonce and no `integrity`, so there is no alternative mechanism permitting it.

### Why nobody noticed

`harness.html` is served by `harness/server.py` under a different, much looser CSP (`frame-ancestors 'none'` only), so the **harness console kept working**. And the sandbox audit that would have caught it (`docs/Local_Sandbox_Complete_Audit_2026-08-03.md`) predates the change by one day and explicitly records that it did not exercise the static pages.

### The fix

Move the block **verbatim** to `static/terminal.js` and load it with `<script src="/static/terminal.js">`. The tag sits exactly where the inline block sat — at the end of `<body>` — so it executes at the identical parse point; no `defer`/`DOMContentLoaded` change needed. **Not one line of JS was edited.**

`'unsafe-inline'` stays **out**. #794's hardening was right, only incomplete — restoring it would undo a genuine improvement.

Confirmed live after the fix: `/static/terminal.js` serves `200 text/javascript` (42,992 bytes) and `node --check` parses it clean.

**None.** No graph edge, gate, route, sanitizer pattern, or config value changes. `invariant-guard` 33/33. This is a same-origin static asset served by the mount that already serves the page.

## Types of changes

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation Update

**Scope note:** Core gate path (static asset only) + tests. No Python logic changed.

## Benefits / why

- Restores the primary user-facing surface of the product. Right now a fresh clone serves a console that looks correct and does nothing.
- Keeps the XSS hardening: `script-src 'self'` with no `unsafe-inline` is strictly better than what existed before #794, and now the page actually complies with it.
- Externalizing the JS is what `script-src 'self'` is *designed* for — this is the intended shape, not a workaround.

## Risks to monitor

- **One new HTTP request on console load.** Same origin, same mount, cached per the existing `no-store` policy on `/` and `/static/*` — so the console re-fetches the JS each load. That matches the previous behavior (the inline block was also re-sent every load) but it is now a second round-trip on a loopback connection.
- **Anything that deep-links or vendors `terminal.html` alone now gets a page with no behavior.** Nothing in this repo does — but a downstream copy would need `terminal.js` alongside it.
- `static/extractor.html` still has the same CSP problem and is **not** fixed here — it is an unreferenced orphan and deleting it is a separate decision (see below).

## Checklist

- [x] Preserves all 6 invariants and I6 isolation (`invariant-guard` 33/33)
- [x] Relevant docs unaffected (`doc-sync` 0 drift)
- [x] Full sandbox validation run — failure set identical to baseline

## Further comments

**ELI5:** A few days ago someone tightened a browser security rule that says "only run JavaScript from real files, not code pasted inside the page." But the console's code *was* pasted inside the page, so the browser started refusing to run any of it. The page still drew itself, so it looked fine — it just did nothing when you clicked. This moves the code into a real file, which is what the security rule wanted all along.

**The regression test is the important part.** `test_csp_script_src_has_no_unsafe_inline` asserted the header *in isolation*, which is precisely why it stayed green through the outage. The new `test_served_html_has_no_inline_script_under_strict_csp` asserts the **pair**: given a strict `script-src`, the served markup must contain no inline `<script>`. It `skip`s if a nonce/hash/`unsafe-inline` is ever added, so it constrains the outcome without dictating the strategy.

Mutation-checked both directions:
```
re-inline the JS  → FAILED: script-src is strict ("script-src 'self'") but the page
                    serves 1 inline <script> block(s) that the browser will block
restore the fix   → PASSED
```

**Cross-file coupling I had to fix:** `test_terminal_contract.py` greps the console source for `fetch` targets and behaviour idioms, and read only `terminal.html` — so emptying it broke 3 of its tests. This is the exact trap `CLAUDE.md` §4 warns about. It now reads `terminal.html` + `terminal.js` as one console source, which is what those tests are actually about; splitting the console further only means adding the new file to that helper.

**Related finding, deliberately not in this PR:** `static/extractor.html` is a 592-line file referenced nowhere in code or docs (only an audit line noting it "was not exercised"), yet served unauthenticated at `/static/extractor.html`. Its escaping is actually correct and it takes no network/URL input, so I found no live XSS vector — but it is dead code on a public mount and is CSP-blocked identically. Recommend deleting it; that is your call, not mine to make unilaterally.

**Verification:** `ruff check --select E,F,I,B,C4,UP,S .` clean · `invariant-guard` 33/33 · `doc-sync` 0 drift · full suite failure set **identical to baseline** (137 before and after, verified by diffing sorted `FAILED` lines; all are `shutil.rmtree(onexc=)`, a Python 3.12-only API, on this container's 3.11).

---


cgfixit:
Triaged the three scanner findings on this PR.

**CodeQL — "bad HTML filtering regexp" (`tests/test_gate.py`): real, fixed in `2413352`.**

Correct catch. HTML tag and attribute names are case-insensitive, so `<SCRIPT>` would have slipped past the new guard while the browser still blocked it — defeating the very outage this test exists to catch. Added `re.IGNORECASE`, which in turn required `\s*` around the `src=` negative lookahead. That second change fixes a false positive in the *other* direction that I'd have shipped: a legitimate external tag written `<script  src = "/x.js">` was being flagged as inline.

Verified against six cases:

| input | flags as inline? | |
|---|---|---|
| `<script>` | yes | correct |
| `<SCRIPT>` | yes | **was missed before this fix** |
| `<ScRiPt type='module'>` | yes | **was missed before this fix** |
| `<script src="/static/x.js">` | no | correct |
| `<SCRIPT SRC="/static/x.js">` | no | correct |
| `<script  src = "/x.js">` | no | **was a false positive before this fix** |

**DevSkim — `setTimeout` at `static/terminal.js:84` and `:182`: false positives, no action.**

Both calls pass an arrow function, not a string, so there is no implicit-eval path for DevSkim's rule to land on:

```js
const timer   = window.setTimeout(() => controller.abort(), timeoutMs);
timeoutId     = window.setTimeout(() => activeQueryController.abort(), queryDeadlineMs);
```

They are also not new code. Both are byte-identical to `static/terminal.html:1106` and `:1204` on `main` — this PR relocates the console's JS out of an inline `<script>` and changes no JavaScript. DevSkim flags them only because the containing file is new to the diff. Changing working code to quiet a scanner that is asking for review rather than reporting a defect would be the wrong trade here; the review conclusion is that both are safe.

**CodeQL "2 configurations not found"** is an infrastructure notice about `/language:actions` and the conda matrix config existing on `main` but not having run on this branch — unrelated to the diff, and not something the PR can resolve.

---

PR #825, opened 2026-08-08
[security] - refuse a non-loopback bind at runtime, not just in CI

cgfixit:
## Proposed changes

You wrote this concern down yourself, in `docs/zIdeas/note.txt`, committed as *"Next legit thing I need to add - auth"*:

> *"Dont forget that curl requests or powershell api commands can still query cyclaw if on same lan - need to add authentication before truly considering this secure"*

I went to test that premise expecting the loopback bind to refute it. **It does not refute it.** Three legs, each verified by execution:

**1. The boot-time validator never looks at the bind address.** `utils/config_validation.py` is the "fail fast" validator — the string `host` appears **zero** times in it. `main()` reads `cfg["api"]["host"]` and passes it straight to `uvicorn.run`. The DevSkim suppression on that line even says *"loopback-only binding by design"* — but nothing enforced the design.

**2. `config-guard`'s C4 does catch it, but only in CI.** C4 fails a non-loopback `api.host` and is build-blocking via the `verify-skills` matrix. But CI only ever sees config that was **committed and pushed**. An operator editing their working copy and running `python gate.py` reached no check at all.

**3. `security.allowed_hosts` is not a backstop — it is already pre-opened for the LAN.** Probed against a real `TrustedHostMiddleware` with the shipped list:

```
shipped allowed_hosts: ['127.0.0.1', 'localhost', '10.0.0.112', '10.0.0.111']
  Host: 127.0.0.1   -> ACCEPTED        Host: 10.0.0.99    -> rejected 400
  Host: 10.0.0.112  -> ACCEPTED        Host: evil.example -> rejected 400
  Host: 10.0.0.111  -> ACCEPTED
```

So the only thing between the shipped tree and a LAN-exposed server was the literal string `127.0.0.1` in `config.yaml`. And `/query`, `/health`, `/` and `/static/*` carry **no authentication**, so the exposure is the corpus (via answers), the soul-informed system prompt, and free use of the local model.

### The control

`main()` now calls `_require_loopback_bind()` before it probes the port or serves.

- Accepts all of `127.0.0.0/8` and `::1` — a **superset** of C4's literal list, so anything C4 allows, this allows. `127.0.0.2` is not refused for no reason.
- An empty host is refused: uvicorn reads `""` as all interfaces.
- An unresolvable hostname is refused rather than resolved — making the bind decision depend on DNS would be worse than refusing.
- `CYCLAW_ALLOW_NON_LOOPBACK_BIND=1` is the escape hatch, and it logs a warning naming the unauthenticated routes.
- **Polarity is deliberate:** opt-**in** to the risky state, the inverse of `agentic/writer.py`'s disable-only kill switch, because here the dangerous configuration is the non-default one.

### The boundary, stated rather than overclaimed

This covers the documented entry points: `python gate.py` and the `cyclaw-server` console script. It does **not** cover `uvicorn gate:app --host 0.0.0.0` run by hand, and it **deliberately** does not cover the container — whose `CMD` is exactly that. There the bind is inside the network namespace and `docker-compose.yml` owns exposure by publishing `127.0.0.1:8787:8787`. Anyone invoking uvicorn directly has stepped outside the supported launch path.

**None weakened; one convention promoted to a control.** No graph edge, router, entry point, sanitizer pattern, or auth path changed. The only behavioural change is a refusal on a configuration that the threat model already declared out of scope. `invariant-guard` 33/33, `config-guard` 0 failures, `doc-sync` 0 drift.

## Types of changes

- [x] Documentation Update
- [x] Invariant / Governance refinement

**Scope note:** Core gate entry point + threat model. No graph topology or soul path touched.

## Benefits / why

- It answers a concern you had already identified and written down, and it answers the half that is actually fixable today. Authentication is the other half and this PR does not pretend to deliver it.
- It moves the loopback boundary from "true because the config file says so" to "true because the process refuses otherwise" — the same distinction #823 was about for the agentic master switch.
- The failure mode it prevents is silent. A non-loopback bind produces no error, no warning, and a perfectly healthy-looking `/health`; you would only discover it by noticing traffic you did not expect.

## Risks to monitor

- **This can stop a server that previously started.** If you are deliberately binding beyond loopback today — a home-lab setup reaching CyClaw from another machine — this refuses on the next restart. That is the intended behaviour and the message says exactly what to do, but it will be a surprise if it happens mid-session. `CYCLAW_ALLOW_NON_LOOPBACK_BIND=1` restores the old behaviour immediately.
- **The escape hatch is a real foot-gun and is meant to be.** Setting it on a machine reachable from your LAN publishes an unauthenticated RAG server. The warning names the routes, but nothing stops you.
- **`allowed_hosts` still lists two LAN IPs.** I did not touch it — it mirrors `allowed_origins` and is clearly your deliberate home-lab config. But it means the Host header will not save you if the bind guard is bypassed, which is why `test_shipped_allowed_hosts_are_not_a_backstop` pins that fact rather than leaving it implicit.

## Checklist

- [x] This change preserves all 6 security invariants and I6 module isolation (`invariant-guard` 33/33)
- [x] No new external network dependencies — the guard is stdlib `ipaddress` only
- [x] Threat model updated (§1 scope row + a new ninth amendment recording why this was a convention rather than a control)

## Further comments

**ELI5:** A server has to pick which network doorway it listens on. "Loopback" means *only programs on this same computer can knock*; anything else means *other machines can knock too*. Every document here says CyClaw only ever uses the loopback doorway — but that was just a setting in a text file, and nothing checked it when the program started. Change one word and the server quietly starts accepting knocks from the whole network, with no password on the front door.

There is also a second lock — a list of names the server will answer to — but it already has two of your LAN addresses on it, so it would have waved those callers through rather than stopping them.

Now the program refuses to start on a non-loopback doorway and tells you why. If you genuinely want that, one environment variable turns it back on, and it warns you that there is still no password.

**Validation performed:**

- All three legs of the premise verified by execution before writing any code (outputs above).
- The guard exercised across `127.0.0.1`, `127.0.0.2`, `localhost`, `::1`, `0.0.0.0`, `""`, `::`, `10.0.0.112`, `192.168.1.5`, `example.com`, plus the opt-in set and the falsey-env set.
- Mutation-tested **four** ways: removing the call from `main()`, making `_is_loopback_host` always return `True`, moving the guard after the port probe, and widening the env check to any non-empty value — each fails the matching test.
- A note on that third one: my first attempt at the ordering mutation reinserted the guard in the position it already occupied, so it was a no-op and the test "passed". Re-done as a real relocation past the port probe, it fails as intended. Recording it because a mutation test that passes is exactly the thing worth double-checking.
- `ruff` clean (the `0.0.0.0` literals carry `# noqa: S104 - asserting the refusal`, matching the existing precedent at `tests/test_harness_robustness.py:106`) · `invariant-guard` 33/33 · `config-guard` 0 failures · `doc-sync` 0 drift.

Suite was run behind the usual uncommitted Python-3.11 `rmtree` shim (this container is 3.11; the repo requires ≥3.12). `git status` verified clean of it before commit.

## Update — re-key for auth+TLS (2026-08-08)

Per `docs/AUTHENTICATION_DESIGN.md` §7 (companion design doc, #829) and your explicit instruction to re-key this now rather than wait for Stage 5: `_require_loopback_bind` gained a second path, checked **before** the env-var escape hatch — `_auth_and_tls_enabled()`, true only when **both** `auth.enabled` and `api.tls.enabled` are set in `config.yaml`. Once real per-user auth and TLS exist, a LAN bind no longer needs `CYCLAW_ALLOW_NON_LOOPBACK_BIND` carried forever.

**Both flags ship `false`** — added as inert Stage 1 config surface (matching `agentic`/`telegram`/`guardrails`'s disabled-by-default convention), so today's refuse-by-default behaviour is byte-for-byte unchanged for every existing install.

**Residual risk worth stating plainly, not burying:** nothing on the request path reads `auth.enabled` yet (that's Stage 2/3 — session store, login, route enforcement) and nothing wires `api.tls.*` into `uvicorn.run` yet (Stage 4). So the two flags are not yet a completed safety story — they encode operator *intent*, not a finished guarantee. Setting both `true` today, ahead of those stages, would bind to a LAN address on a config technicality without actually protecting `/query` or putting TLS on the socket. This is called out in three places so it can't be missed: `_auth_and_tls_enabled()`'s docstring, the `config.yaml` comments on both keys, and the `logger.warning` fired when this path is taken at runtime. Do not flip these flags true until Stage 2/3 and Stage 4 land — see #829's design doc for what "true" is supposed to mean once they do.

New tests (`test_auth_and_tls_both_enabled_allows_a_non_loopback_bind`, `test_one_flag_alone_or_malformed_config_is_not_enough` (8 cases: missing/false/malformed shapes), `test_main_allows_a_lan_bind_when_auth_and_tls_are_both_configured`, `test_shipped_config_leaves_auth_and_tls_both_off`) plus a fresh mutation test (`and` → `or` in `_auth_and_tls_enabled`, caught by 7/8 parametrized cases). Full suite, `ruff`, `flake8`+WPS, `invariant-guard` (33/33), `config-guard` (0 failures), and `doc-sync` (0 drift) all re-run clean after this change.

---

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -346,6 +346,13 @@ policy:
       - "xai-[a-zA-Z0-9]{20,}"       # xAI (Grok) API keys
 
 api:
+  # Enforced at runtime, not just documented: gate.py's main() refuses to serve
+  # on a non-loopback address, because /query, /health, / and /static/* carry no
+  # authentication. Set CYCLAW_ALLOW_NON_LOOPBACK_BIND=1 to override, and put
+  # your own auth in front of CyClaw before you do. config-guard's C4 asserts the
+  # same rule statically, but only against committed config -- this covers the
+  # working copy. The container is unaffected (its CMD binds 0.0.0.0 inside the
+  # network namespace and docker-compose publishes 127.0.0.1:8787:8787).
   host: "127.0.0.1"  # DevSkim: ignore DS162092 - loopback-only binding by design
   port: 8787
   # Overall server-side deadline for one /query (retrieval + graph + LLM). The
--- a/config.yaml
+++ b/config.yaml
@@ -390,23 +390,33 @@ api:
     # fall back to CYCLAW_DB_URL (the personality DB URL) — set this
     # explicitly to opt rate-limiting into Postgres.
     # database_url: null
-  # Stage 1 config surface for docs/AUTHENTICATION_DESIGN.md §5/§7. NOT wired
-  # to uvicorn yet -- that is Stage 4 (cyclaw-gen-cert + ssl_certfile/
-  # ssl_keyfile). Today this key participates ONLY in gate.py's loopback bind
-  # guard (_auth_and_tls_enabled), alongside auth.enabled below. Setting it
-  # true does not put TLS on the socket. Leave false.
+  # NOT wired to uvicorn: nothing here reaches ssl_certfile/ssl_keyfile, so
+  # setting this true does not put TLS on the socket. Today this key
+  # participates ONLY in gate.py's loopback bind guard (_auth_and_tls_enabled),
+  # alongside auth.enabled below. Must be the literal boolean true -- a quoted
+  # "true" is a string, and gate.py deliberately reads it as OFF rather than
+  # letting stray quotes decide a security gate. Leave false.
+  # Background: docs/AUTHENTICATION_DESIGN.md §5/§7 (that doc lands with the
+  # authentication work; this comment stands on its own without it).
   tls:
     enabled: false
 
-# Stage 1 of docs/AUTHENTICATION_DESIGN.md. This flag has NO effect on any
-# request path yet -- nothing checks it except gate.py's loopback bind guard
-# (_auth_and_tls_enabled), which treats "auth.enabled AND api.tls.enabled" as
-# permission for a non-loopback bind (design §7), so a LAN operator is not
-# stuck carrying CYCLAW_ALLOW_NON_LOOPBACK_BIND forever once real enforcement
-# ships. Setting this true today does NOT put a credential in front of
-# /query -- that needs Stage 2 (sessions/login) and Stage 3 (route
-# enforcement), neither of which exists yet. Leave false until those stages
-# ship; see the design doc before turning this (and api.tls.enabled) on.
+# Per-user authentication. This flag has NO effect on any request path yet --
+# nothing checks it except gate.py's loopback bind guard, which allows a
+# non-loopback bind when auth.enabled AND api.tls.enabled are both true AND
+# /query is actually carrying an authentication dependency. That third
+# condition is probed at runtime, not assumed, precisely so these two flags
+# cannot open a LAN bind while route enforcement is still unbuilt: setting
+# them true today does NOT put a credential in front of /query, and the guard
+# refuses accordingly rather than binding with a warning. Once enforcement
+# ships the guard opens on its own, so a LAN operator is not stuck carrying
+# CYCLAW_ALLOW_NON_LOOPBACK_BIND forever.
+#
+# Must be the literal boolean true -- a quoted "true" is a string, and gate.py
+# deliberately reads it as OFF rather than letting stray quotes decide a
+# security gate. Leave false.
+# Background: docs/AUTHENTICATION_DESIGN.md (that doc lands with the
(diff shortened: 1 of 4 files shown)


cgfixit:
@codex review branch and pr description linked to this pr and within context of prs 829 and 830

PR #829, opened 2026-08-08
[security] - per-user authentication + TLS design, and the Stage 1 primitives

cgfixit:
## Proposed changes

This answers the requirement you wrote in `docs/zIdeas/note.txt` at the bar you set for it — **true authentication, no shortcuts**, because other LAN machines may need access later.

It is deliberately split. `docs/AUTHENTICATION_DESIGN.md` is the design, for review **before** anything touches a request path. `utils/authn.py` is Stage 1: the two primitives everything above them will rest on, imported by nothing, so it can be reviewed and tested without changing how a single route behaves.

### What I found while grounding the design

Two things worth knowing before you read the design:

**The transport already exists.** `static/terminal.js`'s `authHeaders()` attaches `Authorization: Bearer` to *every* call including `/query`, and `telegram/client.py:745` does the same — its config field is already documented *"Optional CyClaw API key for POST /query."* Requiring credentials on `/query` isn't a new protocol; it's turning on a check both real clients already satisfy.

**Failed auth is already rate-limited.** The limiter dependency runs *before* `require_api_key` on every protected route, pinned by `TestFailedAuthDoesNotBypassRateLimit`. That anti-guessing property is already correct.

So what's missing isn't authentication — it's **identity**. A single shared secret can't say *which* machine asked, and can't be revoked for one device without rotating it for every device. On a LAN with several machines, that's the gap that matters, and it's why the design is per-user rather than a wider allow-list on the existing key.

### Two design decisions worth your attention

**TLS is part of the feature, not an enhancement.** Without it, the session cookie or bearer token crosses the LAN in plaintext and anyone who can sniff the segment has the credential. A login form over plain HTTP would be theatre. The design treats it as a required stage.

**scrypt, not argon2id — and I corrected myself here.** I initially suggested argon2 when we discussed this. That was wrong for this repo: `argon2-cffi` would be a new **runtime** dependency, which CLAUDE.md §7 makes High-tier and which needs exact pins in both `pyproject.toml` and `constraints.txt` plus a `dep-guard` pass. `hashlib.scrypt` is stdlib, memory-hard (RFC 7914), and measured at ~0.11 s/verification here at n=2¹⁴. argon2id is the stronger primitive in the abstract; at these parameters scrypt is not the weak link, and the dependency cost is real. Stdlib-first is this repo's own convention.

### Stage 1 contents

- **Self-describing records** (`scrypt$n$r$p$salt$hash`) so the work factor can be raised later without invalidating accounts — `verify_password` returns `needs_rehash` and the caller upgrades on the next successful login.
- **Fails closed on any malformed record** rather than raising, because this runs on an unauthenticated route where a 500 would confirm the row exists.
- **Per-account lockout** with exponential backoff and a ceiling. Per-account because the LAN case means many source addresses; a ceiling rather than a permanent lock because a permanent lock hands anyone who knows a username a DoS against its owner.
- **Username canonicalisation** so `Operator` and `operator` can't become two rows that look identical in an audit log, anchored to a slug for the same reason `agentic/registry.py` anchors skill names.

**None.** No graph edge, gate, router, entry point, sanitizer pattern, or config value touched. No route changed. No new runtime dependency. `utils/authn.py` is imported by nothing — `invariant-guard` 33/33, `doc-sync` 0 drift.

## Types of changes

- [x] Documentation Update
- [x] Invariant / Governance refinement

**Scope note:** Design doc + one new inert `utils/` module. No core request path.

## Benefits / why

- It replaces "authentication, someday" with a written scheme that names the threat model delta, the migration path for every existing client, and a staged rollout where each stage leaves the tree working.
- Stage 1 is the part most worth reviewing slowly and least worth rushing — password verification is where auth systems actually fail, and it's reviewable here in isolation rather than buried in a PR that also moves routes.
- It costs nothing to merge: no dependency, no config, no behaviour.

## Risks to monitor

- **The design is a proposal, not a decision.** §10 lists four open items I did not decide for you: username set, session lifetime, whether `/health` stays open, and the bootstrap flow. §7 also proposes re-keying #825's bind guard to "non-loopback allowed when auth **and** TLS are on," which is a better gate than the current env override — but that only lands at Stage 5.
- **scrypt parameters are a judgement call.** n=2¹⁴ is ~0.11 s here. On a much slower machine that becomes noticeable at login; on a much faster one it's cheaper for an attacker. Parameters are in the record precisely so they can be raised without a reset, but the starting value is worth your eye.
- **Enabling auth will break unauthenticated callers** — `ci_rag_smoke.py`, `CyClaw-Sandbox`, any local `curl`. That's why it ships `auth.enabled: false` and why §6 lists the migration per client. The design deliberately has **no loopback bypass**: a `127.0.0.1` exemption would be exactly the shortcut you ruled out.

## Checklist

- [x] This change preserves all 6 security invariants and I6 module isolation (`invariant-guard` 33/33; no request path touched)
- [x] No new external network dependencies — and deliberately no new runtime dependency at all
- [x] Threat model interaction documented (design §3 and §7; `THREAT_MODEL.md` itself changes at Stage 5, not here)

## Further comments

**ELI5:** Right now CyClaw has one shared password for its admin bits, and none at all on the ask-a-question route. That's fine when only your own machine can reach it. The moment you let other machines on the network in, two things break: you can't tell *who* asked, and if one laptop goes missing you have to change the password on every device at once.

So this adds real accounts. Each person gets their own login, the audit log can say who did what, and one device can be cut off without disturbing the others. It also insists on HTTPS, because otherwise the password travels across your network in readable form and the login screen is decoration.

This PR is the plan plus the piece that checks passwords — nothing is switched on yet.

**Validation performed:**

- Mutation-tested, and **two rounds changed the code rather than confirming it**:
  - Removing the explicit scrypt-parameter pre-check failed *no* test. Rather than add a test to justify it, I checked both directions and found it genuinely redundant — scrypt itself rejects a non-power-of-two `n`, `r`/`p` below one, and `dklen=0`. Deleted rather than kept as dead weight, with the `try/except` documented as the single check.
  - The timing-safe compare can't be caught by behaviour, so it's pinned at the source instead; swapping `compare_digest` for `==` now fails that test.
- I also tightened one security comment after testing it: I'd written that an empty stored hash "would reach `compare_digest(b"", b"")`". Probing it showed there is no path to an empty `derived` at all, because scrypt rejects `dklen=0` — so the comparison is *unreachable*, not merely unlikely. The comment now says what was actually proven.
- 51 tests covering roundtrip, salt uniqueness, nine malformed-record shapes, absurd-parameter memory safety, non-string inputs, rehash-on-weak-parameters, username/password policy, lockout arithmetic including overflow, and session-id entropy.
- `ruff` clean · **`flake8` + wemake-python-styleguide clean** (the CI lint gate `ruff` doesn't cover) · `invariant-guard` 33/33 · `doc-sync` 0 drift · full suite green.

Suite was run behind the usual uncommitted Python-3.11 `rmtree` shim (this container is 3.11; the repo requires ≥3.12). `git status` verified clean of it before commit.

---

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -346,6 +346,13 @@ policy:
       - "xai-[a-zA-Z0-9]{20,}"       # xAI (Grok) API keys
 
 api:
+  # Enforced at runtime, not just documented: gate.py's main() refuses to serve
+  # on a non-loopback address, because /query, /health, / and /static/* carry no
+  # authentication. Set CYCLAW_ALLOW_NON_LOOPBACK_BIND=1 to override, and put
+  # your own auth in front of CyClaw before you do. config-guard's C4 asserts the
+  # same rule statically, but only against committed config -- this covers the
+  # working copy. The container is unaffected (its CMD binds 0.0.0.0 inside the
+  # network namespace and docker-compose publishes 127.0.0.1:8787:8787).
   host: "127.0.0.1"  # DevSkim: ignore DS162092 - loopback-only binding by design
   port: 8787
   # Overall server-side deadline for one /query (retrieval + graph + LLM). The
--- a/config.yaml
+++ b/config.yaml
@@ -390,23 +390,33 @@ api:
     # fall back to CYCLAW_DB_URL (the personality DB URL) — set this
     # explicitly to opt rate-limiting into Postgres.
     # database_url: null
-  # Stage 1 config surface for docs/AUTHENTICATION_DESIGN.md §5/§7. NOT wired
-  # to uvicorn yet -- that is Stage 4 (cyclaw-gen-cert + ssl_certfile/
-  # ssl_keyfile). Today this key participates ONLY in gate.py's loopback bind
-  # guard (_auth_and_tls_enabled), alongside auth.enabled below. Setting it
-  # true does not put TLS on the socket. Leave false.
+  # NOT wired to uvicorn: nothing here reaches ssl_certfile/ssl_keyfile, so
+  # setting this true does not put TLS on the socket. Today this key
+  # participates ONLY in gate.py's loopback bind guard (_auth_and_tls_enabled),
+  # alongside auth.enabled below. Must be the literal boolean true -- a quoted
+  # "true" is a string, and gate.py deliberately reads it as OFF rather than
+  # letting stray quotes decide a security gate. Leave false.
+  # Background: docs/AUTHENTICATION_DESIGN.md §5/§7 (that doc lands with the
+  # authentication work; this comment stands on its own without it).
   tls:
     enabled: false
 
-# Stage 1 of docs/AUTHENTICATION_DESIGN.md. This flag has NO effect on any
-# request path yet -- nothing checks it except gate.py's loopback bind guard
-# (_auth_and_tls_enabled), which treats "auth.enabled AND api.tls.enabled" as
-# permission for a non-loopback bind (design §7), so a LAN operator is not
-# stuck carrying CYCLAW_ALLOW_NON_LOOPBACK_BIND forever once real enforcement
-# ships. Setting this true today does NOT put a credential in front of
-# /query -- that needs Stage 2 (sessions/login) and Stage 3 (route
-# enforcement), neither of which exists yet. Leave false until those stages
-# ship; see the design doc before turning this (and api.tls.enabled) on.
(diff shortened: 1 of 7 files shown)


cgfixit:
@codex review branch and pr description linked to this pr and within context of prs 825 and 830

PR #851, opened 2026-08-09
[security] - fail-closed guardrails bool and block MSSQL four-part names

cgfixit:
## What
- Reject non-boolean `guardrails.enabled` at config load (YAML `"false"` no longer arms the layer).
- `guardrail_bridge` requires literal `True` before importing guardrails.
- Reject MSSQL four-part / linked-server object names in `assert_read_only_sql`.
- `test_driver_absent` no longer depends on the host missing psycopg.

## Why / benefit
Closes real fail-open config and SQL boundary gaps found in a ponytail-filtered optimize scan of `origin/main`.

## Risk to monitor
- Operators who relied on stringy YAML booleans will get a config error (intended).
- Three-part `db.schema.table` names remain allowed on purpose.

## Verify
`GROK_API_KEY=dummy python -m pytest tests/test_guardrails_config.py tests/test_guardrail_bridge.py tests/test_sqlconnect_client.py -q`

## Merge order
- **This PR:** P1 of 2 (merge first; independent of P2)
- **Full stack:** P1 → P2 (do not reorder)
- **Topology:** parallel (disjoint paths)

## Base
- GitHub base branch: `main`
- Forked from: `origin/main@678d2f7`

### DIFF
--- a/agentic/sqlconnect/client.py
+++ b/agentic/sqlconnect/client.py
@@ -115,6 +115,17 @@ _FORBIDDEN_FN_RE = re.compile(
     re.IGNORECASE,
 )
 
+# MSSQL four-part / linked-server object refs (server.db.schema.obj and the
+# empty-catalog form server..schema.obj). OPENQUERY/OPENROWSET are already in
+# _FORBIDDEN_RE; four-part dotted names are the other way to leave the DSN's
+# intended database without using those keywords. Three-part names
+# (db.schema.obj) on the same instance are still allowed.
+_MSSQL_FOUR_PART_RE = re.compile(
+    r"(?:\[[^\]]+\]|[A-Za-z_][\w$]*)"
+    r"(?:\s*\.\s*(?:\[[^\]]+\]|[A-Za-z_][\w$]*|)){3,}",
+    re.IGNORECASE,
+)
+
 # Quoted regions can legitimately contain SQL keywords or comment/``;``
 # punctuation as *data* or as a quoted column name -- e.g.
 # ``SELECT 'please do not delete'`` or ``SELECT "delete" FROM t``. Those are never
@@ -319,6 +330,14 @@ def assert_read_only_sql(sql: str) -> str:
             f"forbidden keyword in read-only query: {fn_hit.group(0)!r}",
             code="SQLCONNECT_BAD_QUERY",
             details={"keyword": fn_hit.group(0)},
+        )
+    # Keep identifiers so bracket-quoted four-part names are still visible.
+    multipart = _MSSQL_FOUR_PART_RE.search(_strip_quoted(cleaned, keep_identifiers=True))
+    if multipart:
+        raise SqlConnectError(
+            "linked-server / four-part object names are not allowed in read-only queries",
+            code="SQLCONNECT_BAD_QUERY",
+            details={"name": multipart.group(0)},
         )
     return cleaned
 
--- a/guardrails/config.py
+++ b/guardrails/config.py
@@ -108,12 +108,22 @@ class GuardrailsConfig:
     # --- Validation -------------------------------------------------------
 
     def __post_init__(self) -> None:
+        self._validate_enabled()
         self._validate_engine()
         self._validate_base_url()
         self._validate_threshold()
         self._validate_nemo_config_dir()
         self._validate_metrics_path()
         self._validate_rail_lists()
+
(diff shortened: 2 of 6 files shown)


cgfixit:
@codex review (to see if you concur with this guardrails Boolean failclose to avoid false sense of security with the prior method; I have yet to enable guardrails yet due to the langgraph essentially being a guardrail but def want more layers especially with the container hardening I think the guardrails will make more sense - anyway the review here is basically is this worth it / does it achieve its claim?)

Update:

## What PR #851 Actually Does (beyond what it literally said)

Two independent changes bundled together (labeled P1 of 2):

1. **`guardrails.enabled` type strictness** — `guardrail_bridge.py` now requires the literal Python `True` (`is True`) instead of a truthy check, so a YAML typo like `enabled: "false"` (a non-empty string, therefore truthy in Python) can no longer accidentally arm the layer. Separately, `GuardrailsConfig.__post_init__` now raises `GuardrailsConfigError` if `enabled` isn't a real bool at all.
2. **MSSQL four-part/linked-server name blocking** — `assert_read_only_sql` now rejects `server.db.schema.obj` and `server..schema.obj` patterns, closing a linked-server escape hatch that let a "read-only" query reach outside the intended database without tripping the existing `OPENQUERY`/`OPENROWSET` keyword filter. Three-part same-instance names remain allowed, by design.

Both changes have direct negative-path test coverage, including parametrized bad-value tests (`"false"`, `"true"`, `"0"`, `"1"`, `0`, `1`, `None`) and four SQL injection-shaped linked-server strings.

## Does It Achieve Its Claim?

Partially — and there's a real gap you should catch before merging.

**Where it's solid:** The `_guardrails_enabled()` change genuinely closes a fail-open bug. Under the old code (`if not (cfg.get("guardrails") or {}).get("enabled", False)`), a string `"false"` is truthy in Python, so a config typo would have silently *enabled* guardrails when the operator intended them off — the opposite direction of the "false sense of security" you're worried about, and arguably worse (surprise enablement, not surprise bypass). The fix is correct and the tests prove it.

**Where the claim overreaches:** The two enforcement paths don't actually compose the way the PR description implies.

- `build_input_guard`/`build_output_guard` call `_guardrails_enabled(cfg)` — a plain dict check — and return `None` immediately on anything that isn't literal `True`. This happens *before* `load_guardrails_config()` is ever called.
- The strict dataclass validation (`_validate_enabled`, the one that raises `GuardrailsConfigError` on non-bool values) lives inside `GuardrailsConfig.__post_init__`, which only runs when `load_guardrails_config(path)` is actually invoked.
- Since the bridge already short-circuits to `None` for any non-`True` value *before* that function is called, the "hard fail loudly" validation path is effectively dead code on the real `build_input_guard`/`build_output_guard` call path. The tests for it (`test_non_bool_enabled_raises`) call `load_guardrails_config` directly, bypassing the bridge gate — so they pass, but they don't prove the production code path ever reaches that check.

Net effect: your actual runtime behavior on a malformed `enabled` value is **silent disablement**, not the loud config-error you'd want for a security control. That's a materially different guarantee than what "fail-closed" should mean here — silent-off is still a false sense of security if anyone assumes guardrails are running because the app started up without complaint.

**What it doesn't address at all:** Your original concern was about the guardrail *evaluator* failing at runtime (exception, timeout, malformed model response) mid-request — this PR doesn't touch that path. It's entirely about a config-parsing typo, not about behavior when guardrails are enabled and then error out during actual evaluation. That's a separate and, for your threat model, more important fail-closed question that remains unanswered by this PR.

## Verdict

| Aspect | Assessment |
|---|---|
| Config-typo fail-open bug | Genuinely fixed, well tested |
| "Fail loudly on bad config" claim | Not actually reachable via the real call path — dead-code validation |
| Runtime evaluator fail-closed (your stated concern) | Not addressed by this PR at all |
| MSSQL four-part blocking | Solid, unrelated but good hardening |
| Worth merging | Yes, as incremental hardening — but don't credit it with solving the problem you actually asked about |

Your instinct that LangGraph isn't a substitute for guardrails still holds, and this PR doesn't change that calculus — it's a narrow config-safety patch plus an unrelated SQL boundary fix, not the runtime fail-closed guarantee you're waiting to build before flipping guardrails on. Before enabling guardrails, I'd push for a test that kills/throws inside the actual guardrail evaluation call (not just the config gate) and confirms the request is blocked, not passed through.

Sources

PR #852, opened 2026-08-09
[fix] - stop smoke false-pass and mark guardrail fail-open in audit

cgfixit:
## What
- `smoke.sh` full pytest step requires exit 0 and `passed > 0` (no more false-pass on crashes).
- Fsconnect selftest label uses ASCII `+` instead of `∪` (Windows CP1252-safe).
- Guardrail input/output exception paths set `guardrail_degraded` and audit logs it.

## Why / benefit
Operator reliability: smoke no longer lies; fail-open rails are visible in audit.

## Risk to monitor
- Smoke may start failing where it previously greenwashed a bad suite (correct).
- Audit schema gains one boolean field (backward compatible consumers ignore unknown keys).

## Verify
`GROK_API_KEY=dummy python -m pytest tests/test_graph.py::TestGuardrailInputNode tests/test_graph.py::TestGuardrailOutputNode -q`

## Merge order
- **This PR:** P2 of 2 (merge after P1; independent)
- **Full stack:** P1 → P2
- **Topology:** parallel (disjoint paths)

## Base
- GitHub base branch: `main`
- Forked from: `origin/main@678d2f7`


cgfixit:
If grok build is hit or miss on pr template it seems like it's a githook needing configured or a pre-launch or just a grok build setting but the hooks have worked well with kimi CHINA code haha sry for the outburst

PR #859, opened 2026-08-10
[fix] - recover LM Studio fallback after a both-backends-down boot

cgfixit:
## Proposed changes

Direct answer to the question this session was asked to verify: **with `models.local_llm.fallback.enabled: true`, does CyClaw still work if Ollama is absent and LM Studio is present, the way it did before the Ollama-primary merge?** Empirically: yes for a normal boot order, but **no** for one real sequence — booting the server before the model backend is up, which pre-fallback (pointing straight at one backend) simply tolerated.

Reproduced directly against `llm.client.resolve_local_backend` / `LocalLLMClient` on `main` @ `386bf15`:

| Scenario | Before this PR | After |
|---|---|---|
| Ollama up, LM Studio up | `ollama` (primary) | unchanged |
| Ollama down, LM Studio up (fallback enabled) | `lmstudio` (fallback) — correct | unchanged |
| Both down at boot → LM Studio starts a moment later | **stuck on `ollama`, every request fails until gate.py restarts** | adopts `lmstudio` on the next request, no restart |

Root cause: `resolve_local_backend` caches its result in a process-level dict keyed by config. When both probes fail at boot it still returns primary (so the server doesn't die), but that guess was cached identically to a real selection. `gate.py` builds one `LocalLLMClient` at import, which copies `base_url`/`model` into instance attributes at construction — so the dead address was permanent for the process, and `/health` (which shares the same cache) kept reporting it too.

Fix, two parts:
1. `resolve_local_backend`'s both-failed branch no longer writes to the cache. It returns a `ResolvedLocalBackend` flagged `degraded=True` instead, so the next `resolve_local_backend()` call re-probes rather than replaying the guess. A **positive** selection (primary reachable, or fallback reachable) is cached exactly as before — this only changes the "nothing answered" branch.
2. `LocalLLMClient` remembers whether its own construction was degraded and, only then, calls `resolve_local_backend` once at the top of `generate()` before spending a full request timeout on a known-dead address. Verified a healthy boot pays **zero** extra probes per request (`test_healthy_backend_is_not_re_probed_on_every_generate`).

**Invariant / Governance Impact:** none of the six invariants touched. This is inside `llm/client.py`'s existing local-backend resolution — no new import, no graph edge change, no change to the triple gate (Grok/Claude are untouched). `invariant-guard`: 35 passed, 0 failed.

## Types of changes

**Scope note:** Core graph/gate path (LocalLLMClient is constructed by `gate.py`, but no route/edge/auth logic changes).

## Benefits / why

This is the compatibility guarantee the recent Ollama-primary work needs to actually hold: an operator whose model server takes longer to start than `gate.py` (a normal occurrence, not a misconfiguration) gets a working fallback instead of a silently pinned dead backend. Before this fix, enabling `fallback.enabled: true` was necessary but not sufficient for pre-merge parity — the specific "started late" case still regressed.

## Risks to monitor

- **Re-probe cost is bounded** at `probe_timeout_sec` (default 1.5s) and only paid on requests while a client is in the degraded state — i.e., exactly the condition where every request is failing anyway. Once a backend is adopted, the client returns to zero-probe steady state.
- **Not fixed here, flagged for awareness:** the probe is liveness-only (`GET /v1/models` succeeds), not model-aware. An Ollama instance that answers but is missing the configured tag (`qwen3.6:27b`) still wins the probe and blocks a working LM Studio fallback from ever being tried. Making the probe model-aware is a separate, larger change (it would need to parse the models list and compare against the configured tag, with its own false-positive risk if `/v1/models` doesn't enumerate loaded models identically across backends) — noted here rather than folded into this fix.
- **Detecting drift:** four new tests in `tests/test_client.py` pin the degraded flag, the no-cache-on-double-failure behavior, client-level re-adoption after a backend appears, and the zero-extra-probes contract on the healthy path.

## Checklist

- [x] This change preserves all 6 security invariants and I6 module isolation (`invariant-guard` 35/35; no graph/gate/auth edit)
- [x] No new external network dependencies or mandatory online LLM assumptions — this strengthens the existing offline/local-only fallback path

## Further comments

**Verification:** `tests/test_client.py` + `tests/test_health.py` full pass. `ruff check --select E,F,I,B,C4,UP,S` clean on both touched files. `invariant-guard` 35/35.

**ELI5:** CyClaw tries Ollama first and, if you've turned on the backup, LM Studio second. If it checked at startup and *neither* was running yet, it used to remember "neither works" forever — even after you actually started LM Studio a minute later, every question would keep failing until you restarted the whole server. Now that "neither works" moment isn't remembered; the next question tries again and picks up LM Studio (or Ollama, whichever answers) as soon as it's actually running. If either was already running at startup, nothing about that path changed.

**Merge topology:** independent — cut from `origin/main`, touches `llm/client.py` + `tests/test_client.py` only, no overlap with the other PRs in this batch.

---

### DIFF
--- a/llm/client.py
+++ b/llm/client.py
@@ -24,7 +24,7 @@ import math
 import os
 import time
 from collections.abc import Callable
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from urllib.parse import urlparse
 
 import httpx
@@ -293,6 +293,10 @@ class ResolvedLocalBackend:
     model: str
     source: str  # "primary" | "fallback"
     api_key: str = ""
+    # True when NEITHER backend answered its probe and this is the
+    # boot-must-not-die fallback to primary, rather than a positive selection.
+    # Callers use it to know the choice was a guess worth re-checking later.
+    degraded: bool = False
 
 
 # Process-level resolution cache so LocalLLMClient and /health share one probe.
@@ -452,8 +456,15 @@ def resolve_local_backend(llm_cfg: dict, *, force: bool = False) -> ResolvedLoca
         "local LLM backend: primary and fallback probes failed; using primary (%s)",
         primary_provider,
     )
-    _resolved_local_backends[key] = primary
-    return primary
+    # Deliberately NOT cached. Caching this pinned the process to a backend that
+    # was merely the default-on-failure: booting gate.py before starting the
+    # model server (an ordinary sequence, and one that simply worked when the
+    # base_url pointed straight at a single backend) meant the later-arriving
+    # server was never adopted, /health kept reporting the dead primary, and only
+    # a restart recovered. Leaving the miss uncached costs one short probe pair
+    # per resolve while nothing is reachable -- a path whose requests fail anyway
+    # -- and lets the next caller see a backend that has since come up.
+    return replace(primary, degraded=True)
 
 
 class LocalLLMClient:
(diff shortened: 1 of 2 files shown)


cgfixit:
## Prelim research: "ollama.cpp" vs Ollama vs llama.cpp (5-min timebox, requested out-of-band)

Quick fact-check, not a code change — logging findings here since this PR is the LM Studio-fallback context they relate to.

**"ollama.cpp" is not a real, separately-named project.** Searched for it directly; no project by that exact name exists in the local-LLM ecosystem. What *is* real, and matches what was confirmed a couple weeks back: **Ollama itself is a wrapper built on top of `llama.cpp`'s `ggml` tensor engine** — Ollama doesn't reimplement inference, it packages `llama.cpp`/ggml with a friendlier CLI, model registry, and OpenAI-compatible API surface. `llama.cpp` (Georgi Gerganov, since March 2023) is the actual C/C++ engine underneath.

**Relevant to this PR's fallback design:** LM Studio is built on the same `llama.cpp`/ggml engine as Ollama. Both backends consume the same GGUF format — the single-file format `llama.cpp` defined that bundles weights, tokenizer, and architecture metadata. So the Ollama-primary → LM Studio-fallback split in `config.yaml` (`qwen3.6:27b` tag vs `qwen3.6-27b-instruct` id) is two different naming/packaging conventions over the *same* underlying inference engine and model format, not two different engines — which is the right shape for the graceful-fallback goal `#859` verifies.

**One caveat worth flagging, unverified against this repo's specific quant:** 2026 search results surfaced recurring "Unknown Model Architecture" GGUF-loading failures across both Ollama and `llama.cpp` when a model uses an architecture the installed build doesn't yet recognize (new arch support lags behind model releases). I did not check whether CyClaw's shipped `qwen3.6:27b` build is affected — flagging as a thing to watch if either backend ever fails to *load* the model (distinct from the liveness-only probe this PR's fix targets, which only checks that a server answers `/v1/models`, not that the target model loaded cleanly).

**Sources:**
- [llama.cpp vs. Ollama: Running LLMs Locally for Enterprises](https://picovoice.ai/blog/local-llms-llamacpp-ollama/)
- [What Is llama.cpp? Run GGUF Models Locally](https://explainx.ai/blog/what-is-llama-cpp-run-models-locally-2026)
- [Fix "Unknown Model Architecture" in Ollama and llama.cpp (2026)](https://runaihome.com/blog/unknown-model-architecture-gguf-ollama-llama-cpp-fix-2026/)

No action taken on this PR's diff from this comment — informational only.

---
_Generated by [Claude Code] (Extended research should come from phase 2 of this later) (https://claude.ai/code)_

---

^ Too funny I go to correct you since I didn't ask for another check that lm studio failback good, I call it ollama.cpp (lot of llamas apparently), you show no mercy which is funny honestly and you said the same in both places, and apparently ollama is just a llama.cpp wrapper. But ollama was literally a terminal command to configure . Anyways worth it to know, but not really a priority to change that again now

PR #867, opened 2026-08-10
[docs] - fix stale banned_patterns counts and close doc-sync's .claude/ blind spot

cgfixit:
## Proposed changes

`doc_sync.py`'s D4 check was reporting *"banned_patterns count 40 consistent everywhere it's cited"* while four agent-facing files carried a stale count. Two independent gaps let that happen, and both are fixed here — the checker first, then the values it should have caught.

**Gap 1 — `.claude/**` was never scanned.** D4's own comment already records this exact lesson from a previous round:

> *"A blind spot in a drift checker is worse than no checker, because it is trusted."*

…and lists the files added in response to that earlier miss. Skills, commands, and rules were still outside the net.

**Gap 2 — the claim regex could not match the spelling in question.** It was `r"(\d+)[\s-]+pattern"`, which requires whitespace or a hyphen immediately before `pattern`. The string `"32 banned_patterns"` has `banned_` in between, so even a *scanned* file using that phrasing would have passed clean.

With `.claude/{skills,commands,rules}/**.md` globbed in and the regex widened to `(?:banned_)?pattern`, the checker immediately found **four** stale claims — two more than manual review had spotted:

| File | Claimed | Actual |
|---|---|---|
| `.claude/skills/fable-protocol/SKILL.md` | 32 | 40 |
| `.claude/commands/fable-protocol.md` | 32 | 40 |
| `.claude/skills/injection-redteam/SKILL.md` | 33 | 40 |
| `.claude/skills/invariant-guard/SKILL.md` | 33 | 40 |

The last one is the sharpest: it documents the **G3 check that counts the patterns**, and G3 itself reports 40. A drift checker's own documentation had drifted about the thing it checks.

The `.claude` paths are **globbed rather than listed**, so a newly added skill citing the count is covered without anyone remembering to edit the checker — otherwise this same PR would need writing again in three months.

Two further stale claims in the same class, found by cross-checking CLAUDE.md's load-bearing numbers against their real sources:

- `CLAUDE.md` — *"ubuntu + windows matrix"* → the matrix is three OSes (`ci.yml:179`), and **windows is `continue-on-error`** (`ci.yml:183`), so the old sentence both omitted macOS and implied windows gates when it does not.
- `CLAUDE.md` — *"~100 files"* → **128** `test_*.py` files.
- `.claude/rules/PROJECT_RULES.md` — listed 14 coverage modules, omitting `gate_memory` and `memory`; `pyproject.toml` declares 16. This file loads as project instructions alongside CLAUDE.md, so it is read every session.

**Invariant / Governance Impact:** none of the six. Documentation and one skill-local checker script; no runtime code, config, graph edge, or test touched. `invariant-guard` 35/35. Note the corrected `invariant-guard/SKILL.md` line is prose about G3, not G3's logic.

---

## Types of changes

**Optional free-text scope note:** Docs + audits, plus the `.claude/skills/doc-sync` checker itself. No core graph/gate/soul path, no retrieval path, no CI workflow change.

---

## Benefits / why

Stale numbers in `.claude/**` are worse than stale numbers in ordinary docs. These files are prompt context: they are read at the start of essentially every agent session and their claims get repeated back with confidence. An agent told the scanner has 32 patterns will reason about coverage, gaps, and risk against a number that is 20% low — and will sound certain doing it.

The durable part of this PR is the checker change, not the four corrected numbers. Those numbers drifted because nothing was watching that directory; fixing them without fixing the watcher would guarantee a third round of the same PR. The glob means future skills are covered by default rather than by memory.

---

## Risks to monitor

- **The widened scan can produce false positives.** D4 now flags any number > 5 followed by `pattern`/`banned_pattern` in `.claude/**` that isn't 40. A future skill legitimately discussing, say, "12 patterns" of something unrelated would be reported as drift. It currently reports **0 drift**, so there is no false positive today — but this is the trade-off being made, and the right response to a future false hit is to reword the doc or narrow the regex, not to re-exclude the directory.
- **`doc_sync` reports, it does not gate.** It exits 0 either way, and `verify-skills` runs `continue-on-error: true`. So this improves visibility, not enforcement — nobody's build will break if a count drifts again, it will just be *visible* when someone runs the checker. Worth knowing before assuming this is now impossible to regress.
- **Scope of the count fix is textual.** Nothing about the actual 40 patterns, the sanitizer, or `TestShippedConfigContract` changed — this PR only makes prose agree with `config.yaml`. If the real count changes later, all five citation sites plus config must move together; D4 is now what tells you that.
- **`fable-protocol` has copies outside this repo.** `CLAUDE.md` §9 documents it as registered both in-repo and at the user-level `~/.claude/skills/`. This PR fixes the two in-repo copies; a user-level copy would need the same edit and is not reachable from here. Flagging so the fix isn't assumed to be global.

---

## Verification

- `python3 .claude/skills/doc-sync/doc_sync.py` — **0 drift items found**, exit 0 (was reporting 4 stale claims mid-change, which is how the two unspotted ones surfaced).
- `bash .claude/skills/doc-sync/verify.sh` — OK, including its planted-drift self-test (`detection self-test: PASS (planted drift detected, exit 2)`), so the checker is confirmed still capable of failing rather than trivially passing.
- `python3 .claude/skills/invariant-guard/check_invariants.py` — 35 passed, 0 failed.
- `ruff check --select E,F,I,B,C4,UP,S .` — All checks passed. (The repo's own config excludes `.claude` and ignores `E501`; an explicit per-file invocation surfaces one pre-existing long line at `doc_sync.py:196` that this PR does not touch.)
- Every corrected number verified against its source: `banned_patterns` counted via `yaml.safe_load` (40 entries, 40 unique), test files via `ls tests/test_*.py | wc -l` (128), matrix via `ci.yml:179,183`, coverage sources via `pyproject.toml:184`.

Note on local testing scope: this sandbox runs Python 3.11.15 against a project requiring `>=3.12,<3.13`, so a full local pytest run is not a usable gate this session. No test or runtime code changed in this PR; CI's 3.12 matrix remains the authoritative gate.

---

## Further comments

ELI5: there's a script whose whole job is to notice when the docs and the code disagree about a number. It was looking in most places, but not in the folder holding the instructions that get fed to AI agents at the start of every session — and its search pattern had a small flaw that meant it would have missed the phrasing used there anyway.

So four different files confidently stated the injection scanner has 32 or 33 patterns. It has 40. One of those files was the documentation for the very check that counts them.

The fix teaches the script to look in that folder and to recognise the phrasing, then corrects what it found. It searches the folder with a wildcard rather than a hand-written list, so the next skill someone adds is covered automatically instead of needing this same fix again.

---

### DIFF
--- a/.claude/commands/fable-protocol.md
+++ b/.claude/commands/fable-protocol.md
@@ -163,7 +163,7 @@ Lineage SafeClaw → PsyClaw → CyClaw. Know cold:
   - Integrity: SHA-256 soul.md drift detection + SQLite shadow DB. In-flight:
     3-layer semantic drift detection — (1) structural diffing, (2) NLI entailment
     via DeBERTa-v3-base-MNLI, (3) embedding distance via existing MiniLM stack.
-  - OWASP-aligned injection scanner (32 banned_patterns)
+  - OWASP-aligned injection scanner (40 banned_patterns)
   - MCP server: retrieval-only, sampling:null; telemetry kill-block
   - Soul governance: soul.md mutation requires an explicit human `reason` string
     and goes through PersonalityManager (atomic write) — governed, not silently
--- a/.claude/rules/PROJECT_RULES.md
+++ b/.claude/rules/PROJECT_RULES.md
@@ -49,7 +49,7 @@ Scoped behavioral rules and non-negotiable constraints for Claude Code sessions 
 
 ### Testing
 
-- **Coverage Target:** 80% minimum (measured across `gate`, `gate_ops`, `gate_auth`, `graph`, `mcp_hybrid_server`, `metrics`, `llm`, `retrieval`, `utils`, `sync`, `agentic`, `guardrails`, `harness`, `telegram`)
+- **Coverage Target:** 80% minimum (measured across the 16 sources in `pyproject.toml`'s `[tool.coverage.run]`: `gate`, `gate_ops`, `gate_auth`, `gate_memory`, `graph`, `mcp_hybrid_server`, `metrics`, `llm`, `retrieval`, `utils`, `sync`, `agentic`, `guardrails`, `harness`, `telegram`, `memory`)
 - **Test Command:** `GROK_API_KEY=dummy pytest tests/ -q --tb=short`
 - **No Live Services:** All external deps mocked via `tests/conftest.py`
 - **Exit Codes:** Respect exit code conventions (0=success, 2=operation failed, 3=env/config error, 4=write refused)
--- a/.claude/skills/doc-sync/doc_sync.py
+++ b/.claude/skills/doc-sync/doc_sync.py
@@ -132,17 +132,35 @@ def main(argv: list[str] | None = None) -> int:
     # "consistent everywhere it's cited" -- the count is cited in a mermaid node
     # and a threat table, which no other check reads either. A blind spot in a
     # drift checker is worse than no checker, because it is trusted.
+    #
+    # Second round of the same lesson: .claude/** was still unscanned after the
+    # files above were added, so the fable-protocol skill and its command copy
+    # sat on a stale "32 banned_patterns" while this check again reported
+    # "consistent everywhere it's cited". Agent-facing prompt files are read by
+    # every session, so a wrong number there is repeated back with confidence.
     for opt in (
         "guardrails/rails.py", "agentic/fsconnect/client.py",
         "README.md", "docs/THREAT_MODEL.md", "INVARIANTS.md",
+        "AGENTS.md",
     ):
         fp = root / opt
         if fp.exists():
             cite_files[opt] = fp.read_text(encoding="utf-8")
+    # Agent-facing prompt/rule files. Globbed rather than listed because skills
+    # and commands are added routinely, and a new one citing the count must not
+    # need an edit here to be covered.
+    for sub in (".claude/skills", ".claude/commands", ".claude/rules"):
+        base = root / sub
+        if base.is_dir():
+            for fp in sorted(base.rglob("*.md")):
+                cite_files[str(fp.relative_to(root))] = fp.read_text(encoding="utf-8")
     drift_files = []
     for name, text in cite_files.items():
-        # Find "<n> patterns" / "<n>-pattern" claims and check they equal real_n.
-        for m in re.finditer(r"(\d+)[\s-]+pattern", text):
+        # Find "<n> patterns" / "<n>-pattern" / "<n> banned_patterns" claims and
+        # check they equal real_n. The banned_patterns spelling is matched
+        # explicitly: "32 banned_patterns" has no whitespace directly before
+        # "pattern", so the generic alternative below never saw it.
+        for m in re.finditer(r"(\d+)[\s-]+(?:banned_)?pattern", text):
             claimed = int(m.group(1))
             if claimed != real_n and claimed > 5:  # ignore small unrelated numbers
                 drift_files.append(f"{name} claims {claimed}")
--- a/.claude/skills/fable-protocol/SKILL.md
+++ b/.claude/skills/fable-protocol/SKILL.md
@@ -188,7 +188,7 @@ Lineage SafeClaw → PsyClaw → CyClaw. Know cold:
   - Integrity: SHA-256 soul.md drift detection + SQLite shadow DB. In-flight:
     3-layer semantic drift detection — (1) structural diffing, (2) NLI entailment
     via DeBERTa-v3-base-MNLI, (3) embedding distance via existing MiniLM stack.
-  - OWASP-aligned injection scanner (32 banned_patterns)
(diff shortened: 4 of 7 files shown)


cgfixit:
Note: research latest changes for 5 and see if it's true that I may be giving it too many skills on the new models. I know either way some should be removed but would be worth checking everything once

cgfixit:
Answering the skills question separately from this PR's actual diff (banned_patterns/doc-sync) — this doesn't touch or block that.

**Short answer: your instinct is right, and it's backed by Anthropic's own primary sources, not just a hunch.**

## What's actually verified

- **Lower hallucination/sycophancy, unprompted self-verification** — confirmed via Anthropic's own Sonnet 5 release notes (measured MASK-eval improvement over 4.6; partner feedback describes the model writing a reproducing test and self-verifying a fix without being asked). `fable-protocol`'s "some of this is partly native" framing is directionally correct.
- **Sonnet 5 regressed on hostile-system-prompt / prefill resistance vs 4.6** — confirmed via Anthropic's own System Card (June 30, 2026). This one's worth sitting with: a big injected skill file isn't just token overhead on this model generation, it's measurably *more* attack surface than it was on 4.6. Your own `fable-protocol` §7 already has this exactly right (the `[S5]` rows) — that part doesn't need a fix, it needs to stay.
- **"As models become more capable, they need less prescriptive engineering, not more"** — this is a direct, stated position in Anthropic's own "Effective Context Engineering for AI Agents" piece, which I fetched and read myself rather than trusting a paraphrase. I want to flag one thing precisely because your whole `fable-protocol` is built around not doing this: my first research pass came back citing a specific number (Claude Code's system prompt cut ~800→164 tokens, "80% with no perf loss") attributed to that article. I went and fetched the article directly to confirm before repeating it to you, and **that specific figure is not in it** — the piece makes the general directional claim but no such number. I'm not passing it along. The general claim stands on a primary source; the specific one doesn't, so treat any place you've seen that stat as unverified.

## What I'd actually do with it

Not "delete skills" — most of your 32 are project-specific knowledge transfer (invariants, config contracts, dependency pins, CyClaw's threat model) that no model capability improvement makes redundant, ever. A model getting smarter doesn't make it know your RRF fusion constant is 0.028 on the RRF scale, not cosine.

The real candidates are the **generic** portions of the couple of skills doing general LLM-discipline coaching rather than domain knowledge — `fable-protocol` §1–4 and §6 (mark speculation, self-check, anti-sycophancy-as-capability-compensator) most of all, `karpathy-guidelines` to a lesser degree. Keep `fable-protocol` §5.5 (Sonnet-5-specific routing facts, not discipline), §7 (already correctly calibrated), §8 (100% you/CyClaw-specific), §9. The "no sugarcoating" line in §6 is arguably not a capability compensator at all — that reads like a style preference you want regardless of what the model can do natively, worth keeping on that basis rather than the capability one.

One connected finding from this PR's own diff, worth noting since it's the same root problem: `fable-protocol` currently exists in **three** copies — the canonical `.claude/skills/`, a `.claude/commands/` copy, and (per CLAUDE.md §9) a user-level one outside this repo — and this PR just caught the first two having independently drifted on an unrelated fact (the banned_patterns count). That's the sprawl cost showing up concretely: it's not just "is this skill still needed," it's "how many places does trimming it have to happen."

I haven't changed anything — this is a finding for you to act on, not something I acted on unilaterally. Say the word and I'll draft the actual trim as its own PR (keeping it separate from #867).


---

cgfixit:
@claude yes, please do

cgfixit:
Done — #873.

Worth knowing before you look at it: on a full section-by-section read of `fable-protocol` (rather than the skim I did for my last comment here), the actual redundant-with-native-capability content turned out to be much smaller than I first suggested. Most of the document is reasoning structure and project-specific policy, not raw hallucination-compensation — that doesn't go away just because the model got better. Only one item (§2.3, self-check) had direct, verified overlap with what Anthropic's own release notes describe, so that's compressed, not the sweeping cut I loosely floated.

The bigger, more concrete fix is elsewhere: `.claude/commands/fable-protocol.md` was a hand-maintained 245-line duplicate of the skill body, when every other skill+command pair in this repo (`doc-sync`, `karpathy-guidelines`) uses a ~15-line pointer instead. That's the actual mechanism behind the drift you asked about — I made it a pointer too, which doesn't just fix the current staleness, it removes the second copy that could go stale again.

One heads-up on merge order: #873 and this PR both touch the same two files and **will conflict** in `.claude/commands/fable-protocol.md` whichever merges second — I trial-merged both directions locally, documented the exact resolution in #873's body (take #873's version; it fully supersedes this PR's line-level fix to that file). `SKILL.md` merges clean regardless of order.

Also flagging, since I can't fix it myself: CLAUDE.md says this protocol is also registered at a user-level `~/.claude/skills/fable-protocol/SKILL.md`. It doesn't exist in this sandbox container at all — it's on your actual machine, outside anything I can reach from here. If you want it in sync with the repo copy, that's a manual step on your end.


---

PR #870, opened 2026-08-11
[infra] - close settings.json checker-allowlist gap, document skill-caching policy

cgfixit:
## Branch naming
Head branch `claude/cyclaw-setup-config-ie330a` — Claude Code driver, matches `claude/`.

## Proposed changes
Five change sets, scoped to `.claude/`, `.gitignore`, and `CLAUDE.md` (docs/tooling only — no runtime code):

1. **`.claude/settings.json`** — `permissions.allow` only pre-approved 2 of the 10
   `check_*.py`/`verify.sh` checker scripts that now exist under `.claude/skills/`
   (`config-guard`, `dep-guard`, `OTel-Hardening`, `index-doctor`,
   `injection-redteam`, and `verify-deps` gained checkers since the allowlist was
   last touched, but were never added). `CLAUDE.md` §7 already states these
   checkers, plus documented test/lint/one-shot-index-build commands, are
   "never ask" — the allowlist just hadn't caught up. Added entries for all
   remaining checkers/`verify.sh` scripts, `ruff format`, the documented
   `pytest`/`ci_rag_smoke.py` invocations, and `retrieval.indexer`. Deliberately
   **excluded**: the persistent server launch (`gate.py`/`cyclaw-server`) and
   `clear_cache`'s destructive `--apply` flag — those are qualitatively
   different risk (long-running/port-bound, or file-deleting) even though §7
   loosely groups them under "run commands."
2. **`.claude/README.md`** — added a "Skill Caching Policy" section documenting
   *why* `.claude/skills/` vendors only project-scoped skills. All 32 skills
   named in `CLAUDE.md` §9 were already present (nothing was missing there);
   this documents the decision to *not* vendor personal skills or
   Claude-Code-built-in skills into the repo: no current CyClaw caller (YAGNI),
   upstream drift risk for built-ins, and identity leakage into a shared repo
   for user-scope skills. (A follow-up commit genericized this section to
   remove named user-scope skills from the repo text itself.)
3. **Doc-drift + verify.sh repairs (commits `2feb7d3`, `2ce41ad`)** — the
   follow-up drift pass also touched files the earlier revision of this
   description didn't list:
   - **`CLAUDE.md`**: two new §4 traps and matching §8 command-block updates —
     the macOS install path (Apple Silicon needs **plain** `torch==2.13.0`;
     the `+cpu` local-version wheel doesn't exist on the PyTorch index for
     macOS, and both manifests hardcode `+cpu` by design for Linux/Windows),
[...]

PR #873, opened 2026-08-12
[docs] - de-dup fable-protocol's command copy, audit for model-generation redundancy

cgfixit:
## Proposed changes

Follow-up to a question you dropped on #867 (asking me to research whether the Sonnet 5 / Opus 5 generation makes `fable-protocol` over-scaffolded) and the "yes, please do" that followed. Two changes:

### 1. De-dup `.claude/commands/fable-protocol.md` — the actual root-cause fix

Every other skill+command pair in this repo follows the same convention: `doc-sync.md` and `karpathy-guidelines.md` are ~15-line thin pointers ("Invoke the X skill... see `.claude/skills/X/SKILL.md` for full detail"). `fable-protocol.md` was the one outlier — a 245-line **independent copy** of the skill's ~260-line body, hand-maintained in parallel.

That's not hypothetical risk, it's the exact bug #867 fixes: the command copy was still citing an outdated pattern count and an outdated `Opus 4.8` routing reference after the canonical skill had already moved past both. I replaced it with a thin pointer matching the established convention, which doesn't just fix this instance — it removes the mechanism that let it drift in the first place. There's no longer a second copy to fall behind.

### 2. Audited the canonical `SKILL.md` for redundancy with native Sonnet 5 behavior — smaller cut than I first suggested

My PR-comment reply loosely floated trimming "§1–4 and §6." Going back and reading the whole document section-by-section rather than skimming it, that doesn't hold up: **most of it turned out to be reasoning structure, output format, or CyClaw/user-specific policy — not raw hallucination/dishonesty compensation.** A better-calibrated model still needs told *what* to decompose, *how* to format uncertainty, and *which* project facts matter; none of that gets solved by the model getting smarter. Saying so plainly rather than forcing the bigger cut to match my earlier framing, per the document's own §6.4.

The one item with **direct, verified** overlap: §2.3 ("self-check before finalizing") maps almost exactly onto Anthropic's own release-note description of unprompted self-verification behavior. Compressed, not deleted — kept as "cheap insurance" per the document's own existing v1.1 philosophy, and because Opus/other models still need it at full strength.

Two items I deliberately did **not** touch, and want to name explicitly: §1.5/§3.1/§4.1 (verify stale-prone/current-state facts) stay full strength — lower hallucination is a different property from knowing things past a training cutoff, and no model generation fixes the second one. And the whole thing is a bit more urgent to keep lean than pure token-cost, not less: Sonnet 5's own system card documents a **regression** on hostile-system-prompt/prefill resistance vs. 4.6, so a large injected file is measurably more attack surface on this generation, not just more tokens — worth stating as a second reason to trim, independent of the redundancy argument.

Added a dated `v1.2` note documenting this audit's scope and verdict, matching the file's own existing versioning convention.

**Invariant / Governance Impact:** none. Two `.claude/` files, no code/config/test touched.

---

## ⚠️ Conflicts with #867 — read before merging

This PR and #867 both touch `.claude/commands/fable-protocol.md` and `.claude/skills/fable-protocol/SKILL.md`. I trial-merged both directions locally before opening this:

- **`SKILL.md` merges clean, either order.** My edits (the v1.2 note, §2.3) and #867's edit (the pattern count at line ~210) sit on non-adjacent lines.
- **`.claude/commands/fable-protocol.md` WILL show a merge conflict**, whichever PR merges second — because this PR replaces that file's entire body with a thin pointer, and #867 edits one line *inside* that same body. **Resolution: take this PR's version.** It's a full supersession, not a partial edit — #867's targeted fix to that file becomes moot once there's no more duplicated content to fix. I verified this resolution locally (`git checkout --ours`/stage-2 during the trial merge, re-confirmed 0 doc-sync drift afterward) rather than asserting it would work.

---

## Types of changes

**Optional free-text scope note:** `.claude/` skill/command housekeeping only. No core graph/gate/soul path, no code.

---

## Benefits / why

The command-file de-dup is the durable fix — it's not "fewer words," it's removing a maintenance liability that's already caused one observed bug (#867) and would keep causing more every time the skill changes and the command copy doesn't. The SKILL.md audit is smaller but still real: one item confirmed redundant with verified native model behavior, compressed rather than padded out with unchanged content, with the document's own logic (§6.4: say when you were wrong) applied to my own earlier over-broad framing.

---

## Risks to monitor

- **The user-level copy is out of reach from here.** CLAUDE.md §9 documents `fable-protocol` as also registered at `~/.claude/skills/fable-protocol/SKILL.md` outside this repo. I checked — it doesn't exist in this container at all, meaning it lives on your actual machine/environment, not in any repo-cloned sandbox. I can't touch it from this session. If you want it in sync, that's a manual step on your end (or point it at the repo copy the same way the command file now does).
- **This is one judgment call about what's "verified redundant," not an automated determination.** I marked exactly one item (§2.3) as confirmed-overlapping with primary-source Anthropic claims (release notes + system card, independently checked — I caught and dropped one unverifiable stat a research pass surfaced before it reached you, see my #867 comment). Everything else was a deliberate "keep" call based on my own read of the document's actual content vs. the verified research, not a blanket policy. If you read it differently on any specific line, that's a fair disagreement to have — nothing here is irreversible.
- **Detection of drift:** the de-dup itself is the drift-detection improvement — there's no longer a second copy of this content to check against the first.

---

## Verification

- `python3 .claude/skills/doc-sync/doc_sync.py` — 0 drift items (caught and fixed one false-positive of my own along the way: my first draft of the command file's explanatory note quoted the literal stale number "32" for context, which the D4 checker's regex reasonably mis-read as a fresh claim — reworded to describe the drift without repeating the exact old digit).
- `python3 .claude/skills/invariant-guard/check_invariants.py` — 35 passed, 0 failed.
- `ruff check --select E,F,I,B,C4,UP,S .` — All checks passed.
- Trial-merged against #867 in both directions locally; zero unresolved conflicts in the final tree; `tests/test_agentic_isolation.py` + `tests/test_security.py` (the two suites most likely to notice anything structurally odd) still pass on the merged tree.
- No test/code files touched by this PR itself — it's `.claude/` content only, so the project's own test suite is not the relevant gate here; doc-sync and invariant-guard are.

---

## Further comments

ELI5: you asked whether you're carrying instructions your AI doesn't need anymore now that the models got better at some of this on their own. The honest answer, after actually reading the whole document line by line instead of skimming it for the PR comment: mostly no — most of what's in there is telling the model *how to work on your specific project* and *how to format its answers*, which no amount of the model getting smarter makes unnecessary. One paragraph really was double-covering something the model now does on its own, so that one's trimmed. The bigger, more concrete win here isn't inside that document at all — it's that a *second, hand-copied version* of the whole thing existed elsewhere in the repo and had already quietly drifted out of date, which is exactly the kind of thing you were worried about. That copy is gone now; there's only one source of truth left to keep current.

---


cgfixit:
Too late already merged I'll just manually fix these 2 later

No action required this is a note to myself

cgfixit:
**Resync against current `main` (2026-08-12).** Note: this PR was reported to me as already merged — it wasn't (`state: open`, `merged: false` at the time of this comment). Flagging in case that's a surprise; likely a mix-up with #867 (which is merged, and which this PR follows up on).

Updated this branch to current `main`@`4c634a4` (2 commits ahead of this PR's base, #868 + #869 — neither touches the two files this PR changes). Used `git merge origin/main` rather than a plain rebase: this branch already carries its own merge commit (`5658edc`) with your verified conflict resolution against #867 on `.claude/commands/fable-protocol.md`, and a plain `git rebase origin/main` drops merge commits by default — it replayed only the pre-merge commit and produced a false conflict on that same file, re-fighting a fight you'd already won. Aborted that and merged instead, which is a fast-forward-compatible update (no history rewrite of your existing commits). Zero conflicts, exactly as expected from the file-level check. Re-verified: `invariant-guard` 35/35, `doc-sync` 0 drift, no conflict markers in the tree. Pushed (`5658edc` → `9ba844e`).

**Cross-check against #870:** zero file overlap — #870 never touches `fable-protocol.md`/`SKILL.md`, so merge order between the two doesn't matter for conflicts.

---

PR #882, opened 2026-08-13
[fsconnect] - enable confined macOS list and read setup

cgfixit:
## Proposed changes  - Add an idempotent macOS setup flow that creates `~/CyClaw-FS` mode 700, writes a non-secret README, and best-effort excludes the jail from Time Machine. - Update the existing `config.yaml` path with a typed Python/PyYAML helper: enable only list/stat/read, confine read/write roots to the expanded jail, force strict roots, and keep writes/indexing/hard-delete/symlinks/UNC/removable-volume access off. - Make default install prepare and enable the read/list profile; `--no-fsconnect` prepares only. - Preserve patched tracked configuration across managed-checkout updates with `git pull --ff-only --autostash`; conflicts fail visibly for operator resolution. - Keep uninstall retention-safe: the jail survives unless `--remove-fsconnect` is explicitly passed and confirmed. - Add macOS CI smoke coverage and behavior/static regression tests.  **Invariant / Governance Impact** - I6 remains preserved: setup stays out of band and no core request-path file imports `agentic.fsconnect`. - I1-I5, RAG-first, graph topology, audit convergence, and soul governance are untouched. - Evidence: invariant guard passed `35 passed, 0 failed`; isolation tests pass. - Setup forces `writes_enabled: false` and `index_enabled: false` every run. Shipped `config.yaml` remains disabled.  ## Types of changes  - [ ] Bugfix (non-breaking change which fixes an issue) - [x] New feature (non-breaking change which adds functionality) - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) - [ ] Documentation Update (if none of the other choices apply) - [x] Invariant / Governance refinement (use this for changes that strengthen or evolve the 6 invariants, I6 isolation, or harness phases)  **Scope note:** macOS installer/setup/uninstaller and out-of-band CI tests only.  ## Benefits / why  - A default macOS install gets useful confined list/stat/read access without exposing the home directory. - Re-running setup narrows unsafe prior configuration instead of widening it or enabling mutation. - Uninstall cannot silently remove user files from the jail. - Bash 3.2/BSD userland compatibility is retained; no Homebrew or new dependency is required.  ## Risks to monitor  - The active single config remains a tracked checkout file; `--autostash` preserves cleanly reapplicable local changes, while a true upstream conflict stops for human resolution. - Installer drift could widen roots or flip writes/indexing; exact-contract and byte-idempotence tests pin the safe output. - POSIX installer integration tests skip on Windows; macOS CI is the execution authority.  ## Merge topology  - Independent; base is `main`. - Branch point: `d53ae02dc840d55e6707a7cac88bbd5cacee7653`. - Human merge order: Phase 1 -> Phase 2 -> retarget/merge Phase 3 -> retarget/merge Phase 4.  ## Checklist  - [x] I have read the latest architecture, phase, invariant, and threat-model guidance - [x] This change preserves all 6 security invariants and I6 module isolation - [x] Full sandbox-equivalent validation has been run and passes with no regressions - [x] No new external network dependencies or mandatory online LLM assumptions were introduced - [x] Existing write/audit/quota/trash guards pass; setup forcibly leaves writes/indexing off - [x] macOS harness/setup guidance and CI smoke behavior were updated - [x] Commit messages follow the title prefix convention - [x] Before/after invariant matrix and verification evidence are included below  ## Further comments  | Boundary | Before | After | |---|---|---| | Jail | Not prepared | `~/CyClaw-FS`, mode 700, README | | Allowed operations | Shipped broad template, disabled | Installed profile is exactly list/stat/read | | Roots | No Mac setup profile | Expanded jail only; strict roots | | Writes/index | Off | Forced off on every setup run | | Installer opt-out | None | `--no-fsconnect` prepares without enablement | | Uninstall | No jail policy | Retain by default; explicit confirmed removal | | Core/I6 | Isolated | Unchanged; 35/35 guard checks pass |  Verification on the conflict-free four-phase trial merge:  - `GROK_API_KEY=dummy py -3.12 -m pytest tests/ -q --tb=short -p no:cacheprovider` -> exit 0, 100% (297.5s). - Setup/Mac-script targeted tests -> pass; three POSIX-only integration cases skip on Windows. - Manual Git Bash runs with isolated HOME/config covered prepare-only, default twice, installer default, and exact safe output. - `bash -n` passed for setup/install/uninstall; YAML and workflow parsing passed. - Touched-file Ruff, `compileall`, invariant guard (35/35), doc-sync (0 drift), conflict-marker, and `git diff --check` validation passed. - Exact CI lint reproduction: `flake8 7.3.0` + `wemake-python-styleguide 1.6.2` passed after a reviewed per-file WPS metric grant; Ruff, Pyflakes, and pycodestyle correctness checks remain active.
- Post-fix full suite: `GROK_API_KEY=dummy py -3.12 -m pytest tests/ -q --tb=short -p no:cacheprovider` -> exit 0, 100% (275.8s).
- `mypy 2.1.0` exits with its own `INTERNAL ERROR` before analysis; no mypy diagnostics were produced.  ELI5: a Mac install creates one private CyClaw folder and lets CyClaw list and read only there. Setup repeatedly turns dangerous switches off. It never enables writes or indexing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,7 +222,10 @@ jobs:
           # --repo-path points the installer at THIS PR's own checkout instead of
           # cloning origin/main -- without it the smoke test would silently
           # validate main forever, never the branch under test.
-          bash macos/install-cyclaw.sh --repo-path "$GITHUB_WORKSPACE" --skip-python-deps < /dev/null
+          # The early smoke intentionally skips config enablement because PyYAML is
+          # installed later in this job. tests/test_macos_fsconnect_setup.py runs
+          # the default enablement path after dependencies are present.
+          bash macos/install-cyclaw.sh --repo-path "$GITHUB_WORKSPACE" --skip-python-deps --no-fsconnect < /dev/null
 
           [ -d "$HOME/.CyClaw" ] || fail "home directory not created"
           # NOT "repo" -- that dir is only created on the clone path; --repo-path
@@ -232,6 +235,9 @@ jobs:
             [ -d "$HOME/.CyClaw/$d" ] || fail "missing \$HOME/.CyClaw/$d"
           done
           [ -x "$HOME/.CyClaw/bin/cyclaw" ] || fail "cyclaw shim missing or not executable"
+          [ -d "$HOME/CyClaw-FS" ] || fail "fsconnect jail not prepared"
+          [ "$(stat -f '%Lp' "$HOME/CyClaw-FS")" = "700" ] || fail "fsconnect jail mode is not 700"
+          [ -f "$HOME/CyClaw-FS/README.txt" ] || fail "fsconnect jail README missing"
           grep -qF "$GITHUB_WORKSPACE" "$HOME/.CyClaw/bin/cyclaw" \
             || fail "shim does not reference the checked-out repo (--repo-path not honored)"
 
--- a/macos/_enable_fsconnect_readlist.py
+++ b/macos/_enable_fsconnect_readlist.py
@@ -22,6 +22,7 @@ _SAFETY_VALUES: dict[str, object] = {
     "index_enabled": False,
     "allow_hard_delete": False,
     "allow_unc_roots": False,
+    "allow_macos_volume_roots": False,
     "follow_symlinks": False,
     "scan_content": True,
 }
--- a/macos/_enable_fsconnect_readlist.py
+++ b/macos/_enable_fsconnect_readlist.py
@@ -1,0 +1,144 @@
+"""Enable CyClaw's confined macOS read/list profile in the active config file."""
+
+from __future__ import annotations
+
+import argparse
+import logging
+import os
+import stat
+import tempfile
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+LOGGER = logging.getLogger(__name__)
+_READ_OPS = ["fs_list", "fs_stat", "fs_read"]
+_SAFETY_VALUES: dict[str, object] = {
+    "enabled": True,
+    "allowed_fs_ops": _READ_OPS,
+    "writes_enabled": False,
+    "strict_roots": True,
+    "index_enabled": False,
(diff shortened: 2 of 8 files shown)


cgfixit:
Two small follow-ups from reviewing this stack (#881-#884) alongside a companion fix I opened on `main` for a related, pre-existing gap in `osutil.py`'s `reveal` containment check (#885):

**1. `macos/_enable_fsconnect_readlist.py`'s block-boundary detection is a text heuristic, not YAML-aware.** `_replace_fsconnect_block()` finds the end of the `fsconnect:` block by scanning for the next non-blank, column-0-starting line. I checked this against the real `config.yaml` (lines 730-777) and it's correct today — the block is cleanly delimited by a blank line then a `# ===...` separator comment before `sqlconnect:` starts. It's only fragile against a *future* config.yaml edit that inserts a column-0 comment mid-block, and `_assert_contract()`'s post-write re-parse already catches a bad splice before anything is written (fail loud, not fail silent) — so this isn't a live bug. Worth a small regression test asserting the `fsconnect:` block stays free of column-0 lines until its natural end, so this script's assumption can't silently break later.

**2. The new `--autostash` pull isn't explained to the operator.** `install-cyclaw.sh`'s switch from `git pull --ff-only` to `... --ff-only --autostash` (to preserve the config.yaml edits `setup-fsconnect.sh` writes into the tracked checkout across re-installs) is a reasonable fix and well-commented in the script. But if the stash reapply ever hits a real conflict, the script (`set -euo pipefail`) aborts mid-update with the comment saying changes are "kept for operator resolution" — true, but nothing in the docs tells the operator to check `git stash list`/`git status` if a reinstall ever exits non-zero after the pull step. Minor doc addition candidate (could land in #884's `HARNESS_MACOS.md` update instead, since that's the docs PR in this stack).

Neither is blocking — flagging for whichever of you/Codex wants to fold them in.

---

PR #884, opened 2026-08-13
[docs] - document macOS fsconnect and add an inert launchd template

cgfixit:
## Proposed changes

- Document the macOS fsconnect jail separately from the coding-harness home, including installed defaults, TCC guidance, grep/glob opt-in, write-playbook sequencing, and `--no-fsconnect`.
- Add macOS security-review notes for APFS identity, strict roots, Files and Folders access, removable volumes, and the explicit absence of privacy-profile/FDA automation.
- Record FSEvents incremental indexing as future design intent only; event paths would be hints and `ScopedRoots` held-fd descent would remain authority.
- Synchronize existing README/CLAUDE fsconnect summaries without unrelated rewrites.
- Ship a disabled weekly launchd trash-maintenance template with no KeepAlive, no secrets, explicit log-directory setup, and a command that is a no-op while writes remain off.

**Invariant / Governance Impact**
- I6 is unchanged: no core source files or imports are modified.
- I1-I5, RAG-first, graph topology, audit convergence, and soul governance are untouched.
- The template is not installed or loaded; it cannot enable writes or indexing.
- Evidence: invariant guard passed `35 passed, 0 failed`; isolation tests and the full suite pass.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation Update (if none of the other choices apply)
- [x] Invariant / Governance refinement (use this for changes that strengthen or evolve the 6 invariants, I6 isolation, or harness phases)

**Scope note:** macOS/fsconnect documentation plus one inert LaunchAgent template and static test.

## Benefits / why

- Operators can distinguish the fsconnect data jail from `~/.CyClaw` and understand exactly what setup enables.
- TCC and APFS constraints are documented without encouraging broad home access or privacy-profile automation.
- The future FSEvents phase has an explicit authority boundary before implementation begins.
- Later trash retention can use a reviewable native scheduler template without embedding secrets or silently running today.

## Risks to monitor

- Documentation references Phase 2 setup names and Phase 3 volume behavior, so merge order matters.
- The launchd template contains placeholders and must be edited/tested manually before explicit loading.
- Live launchd scheduling/logging was not exercised from Windows; a physical-Mac smoke remains required.

## Merge topology

- Stacked on `codex/macos-fsconnect-setup-readlist` / draft PR #882.
- PR base is the parent branch, not `main`.
[...]

PR #885, opened 2026-08-13
[fsconnect] - make reveal's root containment check Darwin-aware

cgfixit:
## Proposed changes

`agentic/fsconnect/osutil.py`'s `_within_roots` — the sole containment guard
for the `reveal` CLI command (`python -m agentic.fsconnect.cli reveal`) —
compares `realpath`-resolved paths using plain `os.path.normcase`, which is a
no-op on POSIX. On macOS's default case-insensitive APFS this means `reveal`
can spuriously refuse a legitimate target whose spelling differs only by
letter case (or an embedded Unicode format-control character) from how the
root is spelled in `config.yaml` — an easy thing to hit via Finder/shell
tab-completion. This adds a small, Darwin-only comparison-identity helper
(`_path_identity`: casefold + strip Unicode `Cf` category characters, mirrors
the same defensive posture as the connector's other Darwin path-handling)
so `reveal`'s containment check behaves consistently on macOS. Linux/Windows
behavior is byte-for-byte unchanged (the new helper is a pass-through to the
existing `os.path.normcase` off Darwin).

Found while reviewing the four open `codex/macos-fsconnect-*` PRs (#881-#884)
for the fsconnect macOS work: those PRs add an equivalent Darwin-aware
identity helper (`pathsafe._root_match_identity`) for root-overlap detection
and root selection inside `ScopedRoots`, but don't touch this separate,
pre-existing `osutil.py` containment check that `reveal` depends on — this
PR closes that specific gap on `main` independently (this file predates all
four open PRs and isn't touched by any of them). Once #881 merges, the two
near-identical helpers (`pathsafe._root_match_identity` and this PR's
`osutil._path_identity`) could be consolidated into one shared function to
avoid the duplication — left as a follow-up rather than blocking on merge
order.

- Change is entirely within `agentic/fsconnect/` (out-of-band, I6-isolated
  layer) — no core-path (`gate.py`/`graph.py`/`mcp_hybrid_server.py`) file is
  touched.
- `python3 .claude/skills/invariant-guard/check_invariants.py` — 35 passed, 0
  failed.
- I1-I5 (RAG-first, topology=policy, triple-gated fallback, audit
  convergence, soul governance) are unaffected; this change touches neither
  the graph nor `gate.py`/soul handling.

## Types of changes

**Optional free-text scope note:** Out-of-band `agentic/fsconnect/` layer only.

## Benefits / why

- Keeps the `reveal` command's path-containment semantics consistent with
  how the rest of the connector is starting to treat macOS/APFS path
  comparisons (case-insensitive by default, Unicode format-control-aware),
  rather than leaving one containment check on the old POSIX-generic
  behavior.
- Fail-closed either way: this only *widens* which already
  `realpath`-resolved, already-identical filesystem entities compare equal
  on Darwin — it cannot cause `reveal` to admit a target outside a
  configured root; the previous behavior's failure mode was only spurious
  refusals of legitimate macOS targets, never over-admission.

## Risks to monitor

- `_path_identity`'s Unicode `Cf`-stripping is included for consistency with
  the sibling `pathsafe` helper but is the less load-bearing half of the fix
  (the case-folding is what matters in practice on real APFS volumes) — if
  it ever proves surprising in practice, it's isolated to this one function
  and easy to narrow to case-folding only.
- Not physically verified on real macOS hardware/APFS (this sandbox is
  Linux) — logic-level tests (below) exercise the comparison via
  monkeypatched `sys.platform`/`os.path.realpath`, matching the existing
  test style in `tests/test_fsconnect_pathsafe.py` for the equivalent
  `pathsafe` helper.

## Checklist

- [x] I have read the latest architecture, invariant, and threat-model guidance
- [x] This change preserves all 6 security invariants and I6 module isolation (no core-path file touched; see Invariant Impact above)
- [x] Full sandbox-equivalent validation has been run and passes with no regressions
- [ ] For any agentic/fsconnect/harness change: two-phase audit, quota enforcement, governed delete/trash, and write guards have been verified — n/a, this PR touches only the read-only `reveal` op's containment check, not writes/quota/trash
- [ ] Relevant architecture docs updated — n/a, no documented contract changed (internal helper only)
- [ ] Before/after invariant matrix — n/a, small out-of-band fix; invariant-guard evidence above covers it

## Further comments

Verification run in this sandbox (Python 3.12 venv, `torch==2.13.0` from
plain PyPI since the CPU-index host is blocked by this environment's egress
policy — functionally equivalent for these tests, no GPU code paths exercised):

- `GROK_API_KEY=dummy pytest tests/test_fsconnect_osutil.py tests/test_fsconnect_pathsafe.py tests/test_fsconnect_client.py tests/test_fsconnect_config.py tests/test_fsconnect_cli.py tests/test_fsconnect_indexer.py tests/test_fsconnect_writer.py tests/test_agentic_isolation.py -q` → 100% pass
- `ruff check --select E,F,I,B,C4,UP,S agentic/fsconnect/osutil.py tests/test_fsconnect_osutil.py` → clean
- `python3 .claude/skills/invariant-guard/check_invariants.py` → 35 passed, 0 failed

ELI5: on Macs, folder names aren't picky about capital letters — "Notes" and
"notes" are the same folder. The `reveal` command (open a folder in Finder)
didn't know that, so it could wrongly say "that's outside your allowed
folder" for a folder that's actually the right one, just typed with
different capitalization. This teaches that one check about how Mac folder
names work, the same way another part of this codebase already does
elsewhere. It can't let anything through that shouldn't be — it can only
stop incorrectly *blocking* things that are actually fine.

---
_Generated by [Claude Code](https://claude.ai/code/session_0176DgEYJ9m3fczM4qBMKcvk)_

### DIFF
--- a/agentic/fsconnect/osutil.py
+++ b/agentic/fsconnect/osutil.py
@@ -15,6 +15,7 @@ import os
 import shutil
 import subprocess  # noqa: S404 -- argv-list file-manager launch only; never shell=True
 import sys
+import unicodedata
 from pathlib import Path
 
 from agentic.fsconnect.pathsafe import _norm_contains
@@ -29,15 +30,37 @@ def _file_manager_argv(path: str) -> list[str]:
     return ["xdg-open", path]
 
 
+def _path_identity(path: str) -> str:
+    """Comparison-only identity for containment checks; never an access authority.
+
+    Plain ``os.path.normcase`` is a no-op on POSIX, which is wrong on macOS:
+    APFS is case-insensitive by default and Finder/shell tab-completion make a
+    case-varied spelling of the same real path routine. On Darwin, also strip
+    Unicode format-control characters (category ``Cf``, e.g. zero-width
+    non-joiner) before casefolding so an alias built from invisible characters
+    doesn't read as a different path. Both sides of every comparison in this
+    module are already ``realpath``-resolved before this runs, so folding case
+    here only widens which *already-verified-identical* filesystem entities
+    compare equal -- it cannot admit a target that isn't genuinely the same
+    entity as (or a real descendant of) an allowed root.
+    """
+    normalized = os.path.normcase(path)
+    if sys.platform != "darwin":  # pragma: no cover - exercised via monkeypatch
+        return normalized
+    return "".join(ch for ch in normalized if unicodedata.category(ch) != "Cf").casefold()
+
+
 def _within_roots(target: Path, allowed_roots: list[str]) -> str | None:
     """The canonical path of ``target`` if it is equal-to or under one of
     ``allowed_roots``, else ``None``.
 
     Both sides are ``realpath``-canonicalized (resolving symlinks) and
-    ``normcase``-normalized, then compared with the same segment-aware containment
-    check the read/write paths use (``pathsafe._norm_contains`` — closes the
-    sibling-prefix bypass). A symlink under a root that points outside it resolves
-    outside and is therefore refused.
+    identity-normalized (``_path_identity`` -- ``normcase`` everywhere,
+    additionally case/Cf-folded on macOS), then compared with the same
+    segment-aware containment check the read/write paths use
+    (``pathsafe._norm_contains`` — closes the sibling-prefix bypass). A symlink
+    under a root that points outside it resolves outside and is therefore
+    refused.
 
     Returning the canonical path (not a bool) lets ``reveal`` launch EXACTLY the
     path that was containment-checked. Launching the original, unresolved string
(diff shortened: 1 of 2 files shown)


cgfixit:
@codex since not all directories on APFS are not case-sensitive, I think a few changes need to be made to make the distinction between when it is case-sensitive and when it's not. It appears this treats the entire file system as not case-sensitive. Please research this briefly, analyze the pr branch code and apply fixes if needed.

cgfixit:
You were right — this treated the whole filesystem as case-insensitive, which isn't true for APFS (case-sensitivity is a per-volume format choice).

One note for the record: the bot comment above claiming commit `3d3542b` already fixed this doesn't check out — I verified via the PR's commit list before relying on it, and that commit never existed on this branch (the blob links in that comment resolve back to the original, still-flawed commit that's now merged). Flagging in case it caused confusion.

Since this PR is already merged, I researched the correct fix (confirmed macOS doesn't actually expose a way for Python to ask "is this volume case-sensitive" — verified against CPython's own source, not guessed) and opened the real follow-up as a fresh PR per this repo's merged-PR recovery convention: **#886**. It also fixes the identical assumption in `pathsafe._root_match_identity`, which landed via #881 with the same flaw.

---

PR #911, opened 2026-08-14
[feat] - Keychain wrapper + generated launchd plists for Telegram T1/T2

cgfixit:
## Branch naming
`claude/macos-keychain-telegram-launchagent` — Claude Code driver, matches the required `claude/` prefix.

## Title
`[feat] - Keychain wrapper + generated launchd plists for Telegram T1/T2`

---

## Proposed changes

Third of four "Next integrations" from `docs/work/MACOS_LAUNCHD_INTEGRATION_PLAN.md` (introduced in #908; #909 = uninstall unschedule, #910 = fsconnect-trash). This is the token-bearing integration the planning doc explicitly said needed a Keychain wrapper *first* — so this PR ships that wrapper before the plist generators that depend on it.

**`macos/cyclaw-keychain-env.sh`**: fetches one secret from the macOS Keychain (`security find-generic-password`) and exports it as a named environment variable before `exec`'ing the wrapped command. Composable by design — chain multiple invocations to inject multiple secrets (each layer exports one variable, then `exec`s into the next layer or the real command). Fails closed: refuses to exec anything if the Keychain item is missing or empty. `macos/cyclaw-keychain-set.sh` wraps `security add-generic-password` with a no-echo prompt for storing the secret.

**`python -m telegram.cli poll-plist`** (T2 KeepAlive poller) and **`health-plist`** (T1 periodic `/health` probe + notify-on-fail): both Darwin-only, both generate a plist from real resolved paths, both inject `TELEGRAM_BOT_TOKEN` via the Keychain wrapper (`poll-plist` optionally chains a second layer for `CYCLAW_API_KEY` via `--api-key-service`). Same never-auto-load posture as #909/#910's generators — `install` only writes the plist and prints the `launchctl bootstrap` command.

**Incidental fix**: found that both static plist templates (`macos/LaunchAgents/com.cgfixit.cyclaw.telegram-{health,poll}.plist`) had a literal `--` inside their XML comments (a pre-existing `--chat-id` mention), which makes the whole document invalid XML per spec — silently uncaught since nothing previously parsed them via `plistlib`. Fixed both and added a regression test (`test_all_shipped_launchagent_templates_are_well_formed_xml`) that plistlib-parses every `macos/LaunchAgents/*.plist` so this can't recur silently. (I made and caught the same mistake fresh in this PR's own new comment text — `--api-key-service` — while writing it; that instance is fixed too.)

---

## Security Fixes: Codex P2 Findings Addressed

Three Codex security findings in the Keychain/Telegram/LaunchAgent design were identified and fixed:

### 1. Keychain argv Exposure (Codex P2)
**Finding:** Original design passed bot token to `security add-generic-password` via `-w "$SECRET"` argv flag, exposing the secret to `ps`/procfs inspection during the operation window.

**Fix:** Changed to bare `-w` flag (no attached value), which forces `security` to prompt for the password via secure TTY read (readpassphrase(3), no echo). Secret never enters the script's variables or the `security` process's argv. Added TTY check (`[ ! -t 0 ]`) to fail closed when run non-interactively. ✅ Verified by test `test_set_script_stores_via_pty_and_never_echoes_secret_to_argv` using real `pty.openpty()` terminal simulation.

### 2. ACL Attribution Error (Codex P2)
**Finding:** Prior documentation falsely claimed Keychain "Always Allow" grants would be attributed to `/bin/bash` (the script interpreter), leading to overly broad trust model.

**Fix:** 
- Corrected both wrapper scripts' documentation to accurately explain macOS Security framework ACL behavior: ACL grants are attributed to the **direct caller of the Security API** (in this case `/usr/bin/security`, not bash)
- Added `-T /usr/bin/security` flag to `cyclaw-keychain-set.sh` to explicitly pin ACL trust at store-time, narrowing it from "default creator-trust + operator prompt" to "only `/usr/bin/security` binary"
- Updated `cyclaw-keychain-env.sh` comment to document actual trust boundary and clarify that narrowing ACL to "just this script" is impossible (Keychain trusts binary code signatures, not shell scripts)

### 3. Hardcoded Env Var Names (Codex P2)
**Finding:** `telegram/cli.py` hardcoded `TELEGRAM_BOT_TOKEN` and `CYCLAW_API_KEY` instead of reading configured names from config.yaml, breaking deployments using custom env var names.

**Fix:** Updated `telegram/cli.py` to read `cfg.bot_token_env` and `cfg.query.api_key_env` from loaded config, passing custom names to Keychain wrappers. Updated `poll-plist` and `health-plist` generators to read configured names. ✅ Verified by three regression tests: `test_poll_plist_honors_configured_bot_token_env`, `test_poll_plist_honors_configured_api_key_env`, `test_health_plist_honors_configured_bot_token_env`.

---

## Invariant / Governance Impact

- No invariant touched. Pure out-of-band addition to `telegram/cli.py` + two new `macos/*.sh` scripts + the shared `utils/launchd_plist.py` helper (same one #910 introduced — duplicated here since this branch is cut from `main`, not from #910's branch; see that PR's body for why they're intentionally not stacked).
- `invariant-guard`: 35/35 pass, before and after.

---

## Types of changes
- [ ] Bugfix / Breaking change / Documentation Update / Invariant refinement

**Scope note:** Out-of-band `telegram/` layer + `macos/` install scripts + three Codex P2 security findings addressed.

---

## Benefits / why

- Closes the documented gap directly: `docs/channels/TELEGRAM_DESIGN.md`'s own T1 checklist said "use `TELEGRAM_BOT_TOKEN` through a Keychain/runtime wrapper... do not paste the token into the plist" as an aspiration with no shipped wrapper. Now one ships.
- Removes the shipped templates' `REPLACE_OR_USE_KEYCHAIN_WRAPPER` placeholder tension flagged in the planning doc — the generators never produce a plist with a plaintext token slot at all.
- Addresses three Codex P2 security findings: argv exposure, ACL attribution, hardcoded env var names.
- The chaining design (`wrap_with_keychain_secrets`) generalizes past Telegram — any future generator (e.g. the gate.py/harness integration next in line) that needs a secret can reuse the same primitive.

---

## Risks to monitor

- Not tested on real macOS (Linux sandbox, same documented limitation as #909/#910) — no real `security`/Keychain call was exercised. The wrapper script's own logic (argv validation, chaining, fail-closed on missing/empty item) *was* manually verified end-to-end against a fake `security` binary standing in for the real one (usage error, missing item, empty item, invalid var name, single secret, two chained secrets — all six paths pass). Real Keychain prompting/ACL behavior (the "always allow" dialog, per-app access) is unverified — see the consolidated macOS testing checklist on PR #912 for real-hardware validation steps.
- `health-plist` builds a shell command string (`bash -lc "curl ... || telegram.cli send ..."`) — every path/argument embedded in it is `shlex.quote()`'d, verified by a dedicated test with a config path containing a space.
- Two secrets can be chained but the CLI doesn't validate the two service names are distinct — passing the same service for both would just export the same value to both var names, a harmless no-op rather than a hazard.

---

## Checklist

- [x] Preserves all 6 invariants + I6 (invariant-guard 35/35, unaffected surface)
- [x] Full sandbox validation: `GROK_API_KEY=dummy pytest tests/ -q --tb=short` green, zero failures
- [x] No new external dependencies (stdlib only, plus the macOS-native `security` CLI at runtime)
- [x] All Codex P2 findings addressed with documented fixes + regression tests
- [x] `docs/channels/TELEGRAM_DESIGN.md` updated (operator checklist, follow-up disposition table, M5 rollout notes)
- [x] `python3 .claude/skills/doc-sync/doc_sync.py` — 0 drift items
- [x] `python3 .claude/skills/dep-guard/check_deps.py` — D10 still passes after adding `--cov=utils.launchd_plist`
- [x] Both new shell scripts pass `bash -n` and were manually exercised against a fake `security` binary (all 6 code paths)
- [x] Windows test collection crash fixed via `if os.name != "nt"` guard on pty import

---

## Further comments

**ELI5:** Telegram notifications and two-way chat needed a bot token to run unattended, and the honest answer for "how do I do that safely on a Mac" was previously just a comment saying "use a wrapper" — no wrapper existed. Now there's a real one: it reads the token from the macOS Keychain (the same secure vault your Mac already uses for Wi-Fi passwords and Safari logins) right before starting the Telegram process, so the token is never sitting in a plain-text config file. Two new commands generate the launchd config files with that wrapper already wired in, using your real install paths — still stopping short of turning them on, same as the sync and fsconnect-trash generators before this one.

Additionally, three Codex security findings discovered during implementation were fixed: the Keychain store now uses bare `-w` flag (never exposing secrets via argv), ACL trust is correctly attributed to `/usr/bin/security` (not bash), and env var names are now read from config instead of hardcoded.

Verified in this Linux sandbox with the same Python 3.12.3 venv used for #908/#909/#910. Real-hardware validation checklist consolidated on PR #912 for single unified test run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -734,6 +734,7 @@ jobs:
             --cov=utils.ops_runner \
             --cov=utils.guardrail_bridge \
             --cov=utils.selftest \
+            --cov=utils.launchd_plist \
             --cov=llm.client \
             --cov=graph \
             --cov=retrieval.embeddings \
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -145,6 +145,7 @@ jobs:
           --cov=utils.ops_runner \
           --cov=utils.guardrail_bridge \
           --cov=utils.selftest \
+          --cov=utils.launchd_plist \
           --cov=llm.client \
           --cov=graph \
           --cov=retrieval.embeddings \
--- a/agentic/fsconnect/cli.py
+++ b/agentic/fsconnect/cli.py
@@ -18,6 +18,8 @@ Subcommands:
     quota-status  Report per-root quota usage/limits (--recompute to force a walk).
     index    Stage the file-share into the corpus (--apply [--reindex]); dry-run default.
     reveal   Open a writable root in the OS file manager (out-of-band).
+    trash-empty-plist  Generate (never load) the macOS launchd plist for a
+                        weekly trash-empty job, from real resolved paths.
     test     Run the pre-flight self-test.
 
 Write content is supplied via --body / --body-file / stdin -- the connector never
@@ -31,11 +33,14 @@ from __future__ import annotations
 
 import argparse
 import json
+import platform
 import sys
 from collections.abc import Callable
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from agentic.fsconnect.config import FsConnectConfig, load_fsconnect_config
+from utils import launchd_plist
 from utils.errors import (
     FsConnectConfigError,
     FsConnectError,
@@ -50,6 +55,13 @@ EXIT_OK = 0
 EXIT_FAIL = 2
 EXIT_ENV = 3
 EXIT_REFUSED = 4
+
+# Fixed Label the generated plist owns -- matches the shipped static template
+# at macos/LaunchAgents/com.cgfixit.cyclaw.fsconnect-trash.plist (both write
+# to the same well-known path; the generator is the recommended path, the
+# template stays as a hand-editable reference/fallback).
+_TRASH_LAUNCHD_LABEL = "com.cgfixit.cyclaw.fsconnect-trash"
+_DEFAULT_TRASH_REASON = "weekly launchd retention purge"
 
 
 def _heading(text: str) -> None:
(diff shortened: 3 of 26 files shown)


cgfixit:
Need to do manual local sandbox test and just think about/read the code to make sure I'm not missing anything with the keychain integration.. little bit unsettling haha

cgfixit:
Claude:
Did the read-the-code pass myself before you got to it — here's what I found, no sugarcoating.

**What checks out (verified against the actual code, not asserted):**
- No secret is ever a literal in a plist. Confirmed `cmd_poll_plist`/`cmd_health_plist` never add an `EnvironmentVariables` key at all — the only place a token exists on disk is the Keychain itself.
- `export "$VAR_NAME=$SECRET"` is safe against a malicious secret value: bash expands `$SECRET` once when building that double-quoted string and does *not* re-scan the result for `$(...)`/backticks, so even a secret containing shell metacharacters can't get executed. `VAR_NAME` is separately regex-locked (`^[A-Za-z_][A-Za-z0-9_]*$`) *before* the script ever calls `security`, so a bad `--api-key-service`/token-service arg can't inject into the export either.
- The wrapper never echoes the secret anywhere — not on the missing-item or empty-item error paths, not via `set -x` (not used). `find-generic-password`'s stderr is discarded and its stdout is captured straight into a variable, never printed.
- End-to-end wiring is correct: the exact env var names the wrapper exports (`TELEGRAM_BOT_TOKEN`, `CYCLAW_API_KEY`) are the exact ones `telegram/config.py`'s `resolve_bot_token()`/`resolve_api_key()` read back via `os.environ.get(...)`. Traced it both directions.
- Chaining works as documented: `exec "$@"` correctly re-execs the same absolute script path for the next layer, propagating the inherited environment forward.

**What I found that needed fixing, and fixed:** the PR body's claim that the six shell-level paths (usage error, missing item, empty item, invalid var name, single secret, chained secrets) were "manually verified end-to-end against a fake `security` binary" was true but never became a real test — only the Python-side argv construction had coverage. Pushed `tests/test_cyclaw_keychain_scripts.py`, which runs the actual scripts against a fake `security` stand-in (no real Keychain touched). 12 tests, including one case that wasn't manually checked before: a chain where the *first* secret exists but the *second* doesn't now has an explicit test proving it still fails closed instead of silently starting the process with one variable unset. Full suite still green, ruff/flake8+WPS clean.

**What's still genuinely open (not fixed, your call):**
- `poll-plist`'s `KeepAlive: true` + `ThrottleInterval: 10` (launchd's own default, left unbumped — unlike #912's deliberate 30s for gate/harness) means loading the poll agent *before* running `cyclaw-keychain-set.sh` puts it in a crash-loop-lite: `security find-generic-password` fails, wrapper exits 1, launchd relaunches every 10s indefinitely, spamming `telegram-poll.log`. Not a security issue, just an annoying trap for exactly the operator sequencing mistake someone is likely to make once. Want this bumped to match #912's 30s, or left as-is since it's a cheap `security` call, not a heavy startup like gate.py's?
- Real Keychain ACL/"Always Allow" prompting behavior is still unverified — already flagged in this PR's own Risks section, not new, but worth restating since it's the one thing no amount of Linux-sandbox testing can close. Specifically: since `cyclaw-keychain-env.sh` is a bash *script* rather than a signed binary, macOS's Keychain ACL grant is typically attributed to the interpreter (`/bin/bash`) rather than to this specific script — meaning "Always Allow" on first prompt could plausibly grant blanket access to *any* bash script the same user runs, not just this one. I'd flag this as `speculating:` — I can't test real Keychain ACL semantics from here — but it's worth you specifically checking on real hardware, since it changes the actual trust boundary from "this script" to "this interpreter."

---
_Generated by [Claude Code](https://claude.ai/code/session_01DfhYsUwMyF3dkeHjGBKa12)_

---

cgfixit:
Claude fact checking my paranoia:
Verified all three notes against the actual code and addressed what's addressable:

**1 & 3. Keychain ACL attribution** — confirmed true, and confirmed it can't be closed from this sandbox (it's a real macOS trust-model fact, not something in-repo code controls). What I could do: it was only documented in the PR body/a comment, not in the script itself. Now it's in `cyclaw-keychain-env.sh`'s own header, where an operator reading the script directly will actually see it, with an explicit note that this needs checking on real hardware. Same fix applied to #912's independent copy of the same script.

**2. `poll-plist`'s `ThrottleInterval=10` footgun** — confirmed true (loading the T2 poller before running `cyclaw-keychain-set.sh` does produce a tight, log-spamming crash loop). Deliberately did **not** bump it to match #912's 30s: that value was justified there by `gate.py`'s genuinely slow startup (embedding model + ChromaDB), which doesn't apply to this lightweight poller — and a long-poll network client recovering from an actual transient crash benefits from staying at launchd's faster default, not a slower one. Instead strengthened the generator's own printed output to spell out the actual consequence (crash loop, log spam), not just the setup-order requirement, and pinned it with a test.

Both are documentation/messaging-only changes — no behavior change to what the generated plist does. Full suite + ruff + flake8/WPS + invariant-guard all still green on both branches.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DfhYsUwMyF3dkeHjGBKa12)_

---

cgfixit:
@codex review for security rule violations according to my project rules/invariants and security best practices overall - do a brief code scan with no internet access to limit token usage

PR #940, opened 2026-08-15
[security] - enforce /query auth, wire TLS, add RBAC web admin

cgfixit:
## Title

`[security] - enforce /query auth, wire TLS, add RBAC web admin`

## Proposed changes

Land CyClaw authentication Stages 3, 4, and 6 on top of the already-merged Stage 1/2/5 stack (`docs/AUTHENTICATION_DESIGN.md`). No parallel auth stack.

1. **Stage 3** — When `auth.enabled` is the literal boolean `true`, `POST /query` requires a `cyclaw_session` cookie or a named device token (`require_session_or_token`). CSRF stays off the query path. `/health` stays open. `CYCLAW_API_KEY` still owns `/soul/*` and `/ops/*`. Terminal login form; `/query` no longer sends the ops key as Bearer. Audit events gain `username` when identity resolved. Bind-guard probe flips true only when the named dependency is actually attached (shipped default stays off → probe false).
2. **Stage 4** — `gate._serve` passes `ssl_certfile`/`ssl_keyfile` when `api.tls.enabled` is literal `true`; missing/unreadable files fail closed. `cyclaw-gen-cert` is an openssl wrapper (no new runtime dep). `https://` twins added to `security.allowed_origins`. Docker `CMD` still bypasses `_serve` (documented).
3. **Stage 6** — Roles `admin`/`operator`/`audit` with SQLite+Postgres migration. Last enabled admin cannot be deleted, disabled, or demoted. Gate HTTP admin under `/auth/users*` and `GET /auth/audit/summary`. Audit role cannot call `/query`. Users + Audit panels on the terminal (`/users`); harness shares the same users table with cookie `cyclaw_harness_session` and stays loopback.

Shipped `auth.enabled` and `api.tls.enabled` remain `false`. First password is still `cyclaw-user passwd admin` (discard-hash bootstrap; no printed secret).

- I1 RAG-first: unchanged. `retrieve` still the only entry. Optional `GraphState.username` is identity, not a new first node.
- I2 topology=policy: unchanged. No new conditional-edge source.
- I3 triple-gate: unchanged. No client construction.
- I4 audit convergence: preserved. `username` is an extra field on existing `rag_query` / gate events; `audit_logger` still END.
- I5 soul governance: unchanged. Soul routes still `require_api_key`.
- I6 module isolation: preserved. `gate`/`gate_auth`/`graph`/`mcp` still do not import `harness`/`agentic`/`sync`/`guardrails`/`telegram`. Harness imports `utils.authn_manager` (utils is core, not OOB).
- Evidence: `python ~/.grok/skills/invariant-guard/check_invariants.py --repo-root` → 35/35 PASS on this HEAD.

## Types of changes

- [x] Documentation Update
- [ ] Bugfix
- [ ] Breaking change (shipped defaults unchanged; behavior changes only when the operator flips `auth.enabled`)

Core graph/gate/soul path | Docs + audits | Infrastructure / CI only (`--cov=utils.gen_cert`)

## Benefits / why

- LAN `curl`/PowerShell can no longer hit `/query` once the operator turns auth on (the item the threat-model ninth amendment left open).
- Enabling `api.tls.enabled` actually puts TLS on `python gate.py` / `cyclaw-server`, so Secure cookies work instead of being dropped on plaintext.
- Operator intent #2 (first-launch account + web UX + RBAC) is usable: login UI, Users panel on both consoles, three roles, last-admin protection. Bootstrap still has no default password.

## Risks to monitor

- Enabling `auth.enabled` without `cyclaw-user passwd admin` locks `/query` until a password is set (intended).
- Enabling `api.tls.enabled` without cert files refuses boot (intended). Docker `CMD` still has no TLS.
- `setup-guide.md` is also edited on open PR #939 (macos keys). Hunks do not conflict (trial-merge `--write-tree` clean). Merge either order.
- Audit role denied on `/query` — observers must not be given a device token they expect to query with.
- This does **not** make CyClaw internet-safe. Do not raise rate limits or auth `/health`.

## Checklist

- [x] I have read the latest architecture / SECURITY.md as needed
- [x] This change preserves all 6 security invariants and I6 module isolation (matrix above; checker 35/35)
- [x] cyclaw-sandbox + CI emulation stamp written (`verify_ci_emulation.py`)
- [x] Draft PR only; no push to `main`
- [x] Relevant design / threat-model / CLAUDE.md / setup-guide updated

## Verify

- `GROK_API_KEY=dummy python -m pytest tests/test_gate_query_auth.py tests/test_gate_auth.py tests/test_gate_auth_admin.py tests/test_authn_rbac.py tests/test_gen_cert.py tests/test_terminal_contract.py tests/test_gate.py::TestLoopbackBindGuard tests/test_gate.py::TestProxyHeaderTrust` — exit 0
- Full `pytest tests/` — two `test_ci_coverage_flag_contract` fails for missing `--cov=utils.gen_cert`; flags added in Stage 6 commit; that contract file then exit 0. Remainder of the suite had already passed.
- `ruff check --select E,F,I,B,C4,UP,S` on touched Python — exit 0
- invariant-guard 35/35 PASS
- `verify_ci_emulation.py` — stamped this HEAD (see stamp)

## Merge order

- This PR: P1 of 1
- Full stack: P1
- Topology: parallel from `main`. Independent of #939 except non-conflicting `setup-guide.md` hunks. Merge either first.

## Base

- GitHub base branch: `main`
- Forked from: `origin/main@d6bc631`

## Further comments

Before/after invariant matrix:

| ID | Before | After |
|---|---|---|
| I1 retrieve entry | retrieve first | unchanged |
| I2 topology | 3 routers only | unchanged |
| I3 triple-gate | hybrid+enabled+confirm | unchanged |
| I4 audit | all paths → audit_logger | same + optional `username` field |
| I5 soul | reason + atomic write | unchanged |
| I6 isolation | core ⊄ OOB | unchanged (harness→utils.authn_* only) |
| G1 telemetry kill | before heavy import | unchanged |
| G2 API-key fail-closed | hmac + unset→401 | unchanged; per-user auth is additive |
| G3 sanitizer | 40 patterns | unchanged |
| G4 BM25 JSON | json | unchanged |
| G5 MCP no-sampling | None | unchanged |

ELI5: turning auth on means `/query` needs a real user (cookie or device token). Turning TLS on means the server actually speaks HTTPS. Admins can manage users in the web UI; operators cannot delete people; auditors can look at the audit tab but cannot query.

There is no GitHub PR #969. Only other open PR is #939 (macos keys). This PR does not fix #939's Conda/ubuntu/sandbox red.


cgfixit (commented):
Static review of the full diff (head 95f1ade; CI not yet reported). Overall well-constructed: RBAC is enforced server-side on every admin route, last-admin protection exists at the manager layer, TLS fails closed, and the bind-guard probe logic is preserved. No critical holes found. Findings below, ordered by importance.

## Security

**1. Last-admin protection has a check-then-act race (low severity, real).** `_is_last_enabled_admin()` runs `get_user()` + `count_enabled_admins()` under one lock acquisition, then `set_role` / `delete_user` / `disable_user` mutate under a *separate* `with self._lock:` block. Two concurrent requests demoting the only two enabled admins both pass the check (each still sees the other as an enabled admin) and both commit → zero enabled admins, admin lockout. Low likelihood on a single-operator LAN app, but the fix is cheap: hold `self._lock` across check + mutation in `set_role` / `delete_user` / `_set_disabled`.

**2. Auth dependency runs before the rate limiter on `/query` (minor).** `attach_identity_to_query` does `route.dependant.dependencies.insert(0, ...)`, so `require_session_or_token` executes before `enforce_rate_limit`. Requests with a missing/garbage credential get an un-throttled 401 plus a DB lookup (session validate / token verify) per attempt. Device tokens are high-entropy so brute force isn't practical, but this is an unauthenticated, un-rate-limited DB hit on the exact path the rate limiter was protecting. Consider appending after the rate-limit dependency, or rate-limiting 401s.

**3. Harness session cookie is `secure=False` unconditionally (accepted-but-note).** `harness_login` sets `cyclaw_harness_session` with `secure=False`. Defensible for loopback-only, but if the harness is ever reached over the LAN via SSH port-forward while gate TLS is on, the cookie rides plaintext. Deriving the flag from `api.tls.enabled` (or a one-line comment justifying it) would close this.

**4. Password reset via `window.prompt()` in `auth_admin.js` (cosmetic-security).** Plaintext prompt, visible on screen, no confirm field. Server-side policy in `authn.hash_password` still rejects weak passwords, so impact is shoulder-surfing only — but an odd choice for a security feature.

**5. `/query` 401s are not audited (observation).** Failed auth on `/query` raises inside the dependency, before `query_endpoint` — so unlike `prompt_injection_blocked` / `graph_error`, query-path auth rejections leave no audit event. A revoked token being retried, or a LAN caller probing, is invisible to the audit role's new summary tab. Consider an `auth_query_denied` event in the dependency (the audit callable is already injected into `register_auth_routes`).

## Functionality

**6. Dead condition in `harness_set_password`.** `operator_blocked = account.role == _ROLE_OPERATOR and target.role == _ROLE_ADMIN` already implies the actor is an operator, making `account.role != _ROLE_ADMIN and operator_blocked` redundant; the later `account.role not in {admin, operator}` is what actually rejects audit-role actors. Harmless, but the structure invites a future edit to the wrong guard. Collapse to one check.

**7. `_serve` TLS failure exits 0.** On TLS misconfiguration `_serve` prints to stderr and `return`s — a clean exit. A supervisor (launchd/systemd) watching the port or exit code may restart-loop or report healthy. `sys.exit(1)`/raising would be more operationally fail-closed. (`validate_tls_config` at import already hard-fails the same condition, so this path is mostly unreachable via `main()` — but make the exit code honest.)

**8. `attach_identity_to_query` relies on FastAPI private API.** `get_parameterless_sub_dependant` is internal surface and has shifted across FastAPI versions. Pinned by tests today; a future FastAPI bump could change runtime behavior. Failure mode is at least fail-closed (probe reads False → LAN bind refused), but a comment pinning the version assumption — or attaching the dependency at decoration time behind the `auth_manager is not None` flag — would be sturdier.

**9. `attach_identity_to_query` returns after the first matching route.** Only one `/query` POST exists today, but a second registration (versioned API) would silently go unauthenticated while the probe still reads True from the first. A `found` flag + warning on multiple matches would harden it.

## Verified sound

- JS panel hiding is cosmetic-only; every admin route re-checks role server-side.
- Bearer-path admin writes require admin role; session path requires CSRF via `hmac.compare_digest`.
- `auth_admin.js` uses `textContent` throughout — no XSS sink.
- Audit-role denial on `/query` covers cookie and device-token identity alike — a Telegram-style token can't launder an audit role into queries.
- Role migration (`ensure_users_role_column` + bootstrap promotion) is idempotent and tested against the pre-Stage-6 schema.
- `test_shipped_app_does_not_yet_enforce_a_credential_on_query` correctly pins that the shipped default keeps the probe False — the named-no-op false-open trap is explicitly avoided.

Recommendation: address (or explicitly defer with a note) #1, #2, and #5 before un-drafting; #6–#9 are nits. Nothing blocks the PR's security goals.

cgfixit:
Security revisit needed: 4. Password reset via window.prompt() in auth_admin.js (cosmetic-security). Plaintext prompt, visible on screen, no confirm field. Server-side policy in authn.hash_password still rejects weak passwords, so impact is shoulder-surfing only — but an odd choice for a security feature.

PR #942, opened 2026-08-15
[security] - pin the whole LangSmith tracing namespace, not two of four names

cgfixit:
## Proposed changes

`/OTel-Hardening` Step 1 flagged three drifted langchain-family pins. Step 2 (reading the **installed** source, not docs or memory) found a real, demonstrable telemetry leak that the pin drift had opened.

**An ambient `LANGSMITH_TRACING_V2=true` defeated the entire telemetry kill.**

`langchain-core` 1.5.0 no longer evaluates tracing itself — `_tracing_v2_is_enabled` (`tracers/context.py:132-135`) delegates entirely to `langsmith` 0.10.15's `tracing_is_enabled()`, which calls `get_env_var("TRACING_V2", default=get_env_var("TRACING"))` (`utils.py:141`). `get_env_var` (`utils.py:418-442`) tries the `LANGSMITH_` prefix before `LANGCHAIN_` and **skips only empty values**, so the live precedence is:

```
LANGSMITH_TRACING_V2  >  LANGCHAIN_TRACING_V2  >  LANGSMITH_TRACING  >  LANGCHAIN_TRACING
        ^ not pinned            ^ pinned "false"        ^ pinned "false"       ^ not pinned
```

`TELEMETRY_KILL` pinned the middle two. A value at the **highest-precedence** name therefore won outright over both, latched permanently in `get_env_var`'s `@functools.lru_cache`, and attached `LangChainTracer` to every graph invocation — i.e. every `/query`. **No API key was required:** `_validate_api_key_if_hosted` (`client.py:712-738`) only warns, so the existing `_TRACING_CREDENTIALS` scrub never prevented the upload.

`LANGSMITH_TRACING_V2` is also the name LangSmith's own docs now recommend, which makes it the *likeliest* stray var to find in an operator's shell profile, a container base image, or a site-wide observability agent — exactly the ambient-environment case `utils/telemetry_kill.py` exists to defend against.

### Reproduced end to end

Against the installed `langsmith` 0.10.15, with an ambient `LANGSMITH_TRACING_V2=true` and the kill dict applied afterwards (as `apply_telemetry_kill()` does):

| kill dict | `tracing_is_enabled()` | upload target |
|---|---|---|
| **pre-fix** (this PR's parent) | **`True`** | `https://api.smith.langchain.com` |
| **post-fix** (this PR) | `False` | — |

### What changed

1. **All four tracing names pinned** to `"false"` in `TELEMETRY_KILL`. `"false"` is inert on every reader: langsmith requires exactly `"true"`, and langchain-core's `env_var_is_set` (`utils/env.py:18-23`) treats `"false"` as unset — which additionally stops the legacy `LANGCHAIN_TRACING` v1 check (`callbacks/manager.py:2492-2506`) raising its `RuntimeError` on an ambient truthy value. Safe unconditionally; carries none of `HF_HUB_OFFLINE`'s first-run bootstrap risk.
2. **`LANGSMITH_ENDPOINT` + `LANGSMITH_RUNS_ENDPOINTS` added to the scrub tuple** as defense in depth. The tuple already popped `LANGCHAIN_ENDPOINT` but not its `LANGSMITH_` twin, so a future tracing regression could still have been aimed at an arbitrary host (or, via `RUNS_ENDPOINTS`, fanned out to a list of them). Popping only ever removes a destination override — it can never enable an upload.
3. **Contract updated in the same commit** per the skill's Step 3: `BASELINE_KEYS`/`BASELINE_CREDENTIALS` and the re-verified langchain-family pins in `check_otel.py` (`_TODAY` moved to the real scan date), the reference `.env`, and `tests/test_telemetry_kill.py`.
4. **Two stale mutation `sed`s in the skill's own `verify.sh` fixed** — see "Risks to monitor".

**Invariant / Governance Impact:** none of the six invariants is touched. G1 specifically re-checked: `gate.py`'s `_TELEMETRY_KILL = apply_telemetry_kill()` binding is untouched in name and position (`invariant-guard` → `PASS _TELEMETRY_KILL (line 51) precedes first heavy import (line 70)`, 35 passed / 0 failed). I6 holds — `utils/telemetry_kill.py` remains stdlib-only.

---

## Types of changes

**Optional free-text scope note:** Telemetry kill switch + its contract (checker baseline, reference `.env`, tests). No graph/gate/soul/auth path.

---

## Benefits / why

- **Closes a live egress path.** `docs/THREAT_MODEL.md` forbids telemetry outright; this was a reachable, reproducible violation of it that required nothing more exotic than one env var in the ambient environment — and query-derived run inputs/outputs are what would have been uploaded.
- **Closes the namespace, not the instance.** Pinning all four names means no future precedence surprise in the same chain can reopen it; the earlier fix only ever covered the names that happened to be known at the time.
- **The tests can no longer drift out from under the code.** Both the kill-var and credential assertions now derive from the module's own constants, so a name added to either set is enforced automatically — this is precisely the drift that let `LANGSMITH_TRACING_V2` and the `LANGSMITH_` endpoint twins go unpinned and untested.

---

## Risks to monitor

- **`verify.sh`'s T5 mutation was passing for the wrong reason and this PR exposed it.** It hardcoded `langgraph==1.2.6`; once the real pin moved to `1.2.9` the `sed` became a no-op, and the test still "passed" only because the clean tree was by then WARNing about that very drift — the thing the mutation was supposed to induce itself. Updating the baseline removed that incidental warning and surfaced the break. Both it and T3 (which hardcoded the credential tuple's one-line spelling, broken by this PR's reformat) are now pattern-matched rather than literal-matched. **`bash .claude/skills/OTel-Hardening/verify.sh` → ALL PASS (7/7 mutations)**, where it reported 2 failures mid-change.
- **Two standing limits this PR does not close**, recorded so they are not mistaken for covered:
  - `langsmith` never reads `OTEL_SDK_DISABLED` — it builds its OTLP exporter directly. Do **not** credit the kill dict's OTel keys for the LangSmith OTLP route; `LANGSMITH_OTEL_ENABLED=false` plus tracing-off is what covers it.
  - `~/.langsmith/config.json` (via `LANGSMITH_CONFIG_FILE`/`LANGSMITH_PROFILE`, `client.py:1265-1286`) can re-supply credentials the scrub removed, and its default path is consulted even with those vars unset. Env scrubbing alone can never close that file-based channel — keeping tracing off is the only control that does, which is what this PR hardens.
- **Broad-looking but additive:** the kill dict grows 16 → 18 keys and the scrub tuple 3 → 5. Nothing is removed or weakened; every added value is either inert (`"false"`) or a removal.

---

## Checklist

- [x] This change preserves all 6 security invariants and I6 module isolation (`invariant-guard` → 35 passed, 0 failed; G1 binding untouched)
- [x] Full pytest run passes with no regressions
- [ ] For any agentic/fsconnect/harness change: n/a — no OOB subsystem touched

---

## Further comments

**Verification run locally** (Python 3.12 venv, `GROK_API_KEY=dummy`):

- `pytest tests/` — full suite green (exit 0).
- `pytest tests/test_telemetry_kill.py tests/test_embeddings.py` — 48 passed.
- `python3 .claude/skills/OTel-Hardening/check_otel.py` — **0 failures, 0 warnings** (was 0 failures / 3 warnings before: the three drifted pins are now re-verified rather than merely silenced).
- `bash .claude/skills/OTel-Hardening/verify.sh` — ALL PASS (clean tree + 7 mutation self-tests).
- `ruff check --select E,F,I,B,C4,UP,S` on all three touched Python files — clean.
- `invariant-guard` → 35 passed, 0 failed.

**Mutation-checked the new guard** (it is not vacuous): deleting only the `LANGSMITH_TRACING_V2` line — i.e. restoring the exact pre-fix state — fails `test_langchain_tracing_disabled` and `test_ambient_langsmith_tracing_v2_cannot_re_enable_upload` with `AssertionError: LANGSMITH_TRACING_V2='true'` (the ambient hostile value surviving). Restoring it makes both pass.

The new test asserts the **behaviour**, not just the env var: it calls langsmith's own `tracing_is_enabled()` under a hostile environment that sets all four names to `true`, and skips cleanly (rather than failing) where langsmith is not installed.

**Scan provenance:** three parallel readers over the installed `langchain_core` 1.5.0 / `langsmith` 0.10.15 / `langgraph` 1.2.9 (+ `sdk`/`checkpoint`/`prebuilt`) source, then an adversarial coverage pass against `TELEMETRY_KILL` that was asked to assume a hostile ambient environment and find anything *outside* the kill set that could re-enable an upload on its own. It returned exactly one such var — this one. Everything else it surfaced (`LANGSMITH_TRACING_MODE`, `_OTEL_ONLY`, `_ENDPOINT`, `_RUNS_ENDPOINTS`, `_CONFIG_FILE`, `_PROFILE`, `_SAMPLING_RATE`, `_USE_PYO3_CLIENT`, the `OTEL_EXPORTER_OTLP_*` pair, `LANGCHAIN_BASE_URL`) is reachable only downstream of a `Client` that exists only once tracing is already on, so all of it is closed by this fix rather than needing its own. `LANGSMITH_TEST_TRACKING` and `LANGGRAPH_API_KEY` ride code paths (langsmith pytest markers, `langgraph_sdk`) that grep confirms this repo never invokes. **LangGraph itself carries no telemetry mechanism of its own** — confirmed across all four installed distributions.

**ELI5:** CyClaw force-sets a list of "do not phone home" switches before any library loads. The tracing switch turned out to have four different names, tried in a fixed order, and CyClaw was only holding down the middle two. If an operator's machine had the *first* name set to on — the one LangSmith's current docs actually recommend — it beat both switches CyClaw was holding, and every question anyone asked CyClaw would have been uploaded to LangSmith's servers. It didn't even need an API key. This holds down all four.

**Merge topology:** independent — cut from `origin/main` (`f561d06`). Touches no file shared with open PR #940 (auth stages) or #941 (CodeQL fixes).

---
_Generated by [Claude Code](https://claude.ai/code/session_01UvSAa5b6T9CD8bKrcnot6m)_

### DIFF
--- a/.claude/skills/OTel-Hardening/check_otel.py
+++ b/.claude/skills/OTel-Hardening/check_otel.py
@@ -61,7 +61,14 @@ from pathlib import Path
 # dict but absent here is new and expected -- T2 reports it as INFO so this
 # list gets updated, not because it is dangerous by itself.
 BASELINE_KEYS = frozenset({
-    "LANGCHAIN_TRACING_V2", "LANGSMITH_TRACING", "LANGSMITH_OTEL_ENABLED",
+    # All four tracing names, not just two: langsmith's get_env_var resolves
+    # LANGSMITH_TRACING_V2 > LANGCHAIN_TRACING_V2 > LANGSMITH_TRACING >
+    # LANGCHAIN_TRACING, so leaving any one unpinned lets an ambient value at
+    # that name win over every pinned name below it (verified 2026-08-15
+    # against installed langsmith 0.10.15 -- see utils/telemetry_kill.py).
+    "LANGSMITH_TRACING_V2", "LANGCHAIN_TRACING_V2",
+    "LANGSMITH_TRACING", "LANGCHAIN_TRACING",
+    "LANGSMITH_OTEL_ENABLED",
     "LANGGRAPH_CLI_NO_ANALYTICS",
     "NEMO_GUARDRAILS_NO_USAGE_STATS", "ANONYMIZED_TELEMETRY",
     "HF_HUB_DISABLE_TELEMETRY", "DO_NOT_TRACK", "ORT_TELEMETRY_OPT_OUT",
@@ -70,7 +77,12 @@ BASELINE_KEYS = frozenset({
     "OTEL_METRICS_EXPORTER", "OTEL_LOGS_EXPORTER",
 })
 
-BASELINE_CREDENTIALS = frozenset({"LANGCHAIN_API_KEY", "LANGSMITH_API_KEY", "LANGCHAIN_ENDPOINT"})
+BASELINE_CREDENTIALS = frozenset({
+    "LANGCHAIN_API_KEY", "LANGSMITH_API_KEY", "LANGCHAIN_ENDPOINT",
+    # Destination overrides, popped as defense-in-depth since 2026-08-15 so a
+    # future tracing regression cannot also be aimed at an arbitrary host.
+    "LANGSMITH_ENDPOINT", "LANGSMITH_RUNS_ENDPOINTS",
+})
 
 # Vendor pins whose telemetry surface this dict targets, and the version last
 # verified against that vendor's actual source (not just its docs). Pulled
@@ -79,9 +91,14 @@ BASELINE_CREDENTIALS = frozenset({"LANGCHAIN_API_KEY", "LANGSMITH_API_KEY", "LAN
 # live-search half of this skill against that vendor's current release notes.
 LAST_VERIFIED_VENDOR_PINS = {
     "chromadb": "1.5.9",
-    "langchain": "1.3.11",
-    "langchain-core": "1.4.8",
-    "langgraph": "1.2.6",
+    # The three langchain-family pins were re-verified 2026-08-15 by reading
+    # the installed source directly (langchain-core 1.5.0 delegates tracing
+    # entirely to langsmith 0.10.15; langgraph 1.2.9 + sdk/checkpoint/prebuilt
+    # carry no telemetry mechanism of their own). That pass found and closed
+    # the LANGSMITH_TRACING_V2 precedence gap -- see utils/telemetry_kill.py.
+    "langchain": "1.3.14",
+    "langchain-core": "1.5.0",
+    "langgraph": "1.2.9",
     "nemoguardrails": "0.23.0",
     "sentence-transformers": "5.6.0",
 }
@@ -96,7 +113,7 @@ TRANSITIVE_VENDORS_TO_REPORT = ("huggingface_hub", "onnxruntime", "opentelemetry
 # rule; vendors do not ship telemetry changes on a schedule.
 STALE_AFTER_DAYS = 120
 
-_TODAY = date(2026, 7, 29)  # scan-day anchor for THIS script's own last edit; see SKILL.md Step 4
+_TODAY = date(2026, 8, 15)  # scan-day anchor for THIS script's own last edit; see SKILL.md Step 4
 
 _VERIFIED_DATE_RE = re.compile(r"[Vv]erified\s+(\d{4}-\d{2}-\d{2})")
 
--- a/.claude/skills/OTel-Hardening/verify.sh
+++ b/.claude/skills/OTel-Hardening/verify.sh
@@ -65,8 +65,23 @@ fi
 rm -rf "$a"
 
 # 4. T3 FAIL: shrink _TRACING_CREDENTIALS to two names.
+# Rewritten by regex over the whole tuple rather than an exact-text sed: the
+# previous version hardcoded the one-line three-name spelling, so reformatting
+# the tuple (or adding a name to it) silently turned this mutation into a no-op
+# and the test then "passed" without ever mutating anything.
 a="$(_mktree)"
-sed -i.bak 's/_TRACING_CREDENTIALS = ("LANGCHAIN_API_KEY", "LANGSMITH_API_KEY", "LANGCHAIN_ENDPOINT")/_TRACING_CREDENTIALS = ("LANGCHAIN_API_KEY", "LANGSMITH_API_KEY")/' "$a/utils/telemetry_kill.py"
+python3 - "$a/utils/telemetry_kill.py" <<'PY'
+import re, sys
+path = sys.argv[1]
+src = open(path, encoding="utf-8").read()
+src, n = re.subn(
+    r"_TRACING_CREDENTIALS\s*=\s*\([^)]*\)",
+    '_TRACING_CREDENTIALS = ("LANGCHAIN_API_KEY", "LANGSMITH_API_KEY")',
+    src, count=1, flags=re.S,
+)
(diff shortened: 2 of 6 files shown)


cgfixit:
@claude isnt there a otel key for home brew in terminal for telemetry? I did it in terminal but don't see that in the env kill switch for brew no telemetry? Should it be for Mac's that aren't mind since I already ran that cmd

cgfixit:
Good catch, and the instinct about *other people's* Macs is the right one — but `TELEMETRY_KILL` is the wrong lever for it, so I'd keep it out of this PR.

**The var is real.** `HOMEBREW_NO_ANALYTICS=1` is Homebrew's documented env opt-out. What you ran in terminal (`brew analytics off`) writes a **persistent config file** rather than an env var, so your machine is already covered permanently — you don't need the env var at all on your own Mac.

**Why it wouldn't work in `TELEMETRY_KILL`.** That dict is applied with `os.environ[...] = ...` inside the CyClaw **Python** process, so it only ever reaches a program CyClaw itself spawns as a subprocess. CyClaw never spawns `brew`. It declares no Homebrew dependency at all:

> `macos/install-cyclaw.sh:37-38` — *"no Homebrew dependency declared or required (Homebrew Python works fine if present, but is never assumed)"*

and the only `brew` string in the entire repo is inside a **warning message** telling you how to install Python 3.12 (`install-cyclaw.sh:143`). So the key would sit in the dict unread — exactly like `ORT_TELEMETRY_OPT_OUT`, which the module already documents as a deliberately inert entry kept only for parity with the reference `.env`. I'd rather not add a second one: an unread key in the kill dict implies a protection CyClaw isn't actually providing.

**Where it *would* genuinely work** is `docs/security-philosophy/cyclaw_telemetry_kill.env`. That file is designed to be `source`d into a shell (*"Source this before running"*), so an export there really does suppress brew analytics for every subsequent `brew` command in that shell. That's the honest home for it — and it's the file a fresh operator would actually use.

**One catch if we do it.** `check_otel.py`'s T8 asserts the `.env` documents *exactly* the kill dict plus the conditional HF pair, so a shell-only var trips it. Verified just now against this branch:

```
WARN [T8] cyclaw_telemetry_kill.env has drifted from the code --
          missing from doc: []; undocumented extra: ['HOMEBREW_NO_ANALYTICS']
```

So it needs T8 taught about a "shell-only, deliberately not in the Python dict" section — same shape as the exemption `HF_HUB_OFFLINE`/`TRANSFORMERS_OFFLINE` already get. Small, but a real contract change rather than a one-line append.

**Scope.** Homebrew's analytics report brew's own usage, not CyClaw queries — so this is operator machine hygiene rather than CyClaw egress, which is why it sits outside the current threat-model boundary rather than being an oversight in it. Still worth doing for exactly the reason you gave (a fresh Mac following the setup guide has brew analytics on by default).

Happy to ship it as its own small follow-up: `.env` entry + T8 shell-only section + a line in the macOS setup guide so a new operator is told to run `brew analytics off`. Keeping #942 scoped to the LangSmith namespace fix.

Sources: [Homebrew — Anonymous Analytics](https://docs.brew.sh/Analytics)

---

PR #951, opened 2026-08-15
[fix] - measure the rclone retry budget on the monotonic clock, not time.time()

cgfixit:
## Proposed changes

`sync.runner._run_sync_locked` derives `sync_timeout_sec`'s budget — the ceiling on how long one sync may hold the single-instance `sync.lock` across the whole exit-5 retry sequence — from `time.time()`:

```python
deadline = (time.time() + cfg.sync_timeout_sec) if has_budget else None
...
remaining = deadline - time.time()
```

`time.time()` is the wall clock and can *step*, not just tick: an NTP correction or an operator setting the system clock mid-sync moves it. Consequences, in order of severity:

- **Backward step** → `remaining` grows, so the lock is held past the documented `sync_timeout_sec` ceiling. That ceiling exists specifically to stop a retry path from holding the lock for `attempts * sync_timeout_sec + sum(backoff)`.
- **Forward step** → `SyncRuntimeError("... budget exhausted by retries")` fires on a sync that still had budget left, and `run_post_sync_check` is handed a `remaining_budget_sec <= 0` so it skips the integrity check entirely and emits `sync_check_skipped_budget_exhausted`.
- Either direction distorts the per-attempt `subprocess.run(timeout=...)` value.

This switches the four deadline/remaining computations to `time.monotonic()` — the primitive Python documents for elapsed-time arithmetic, and the one **this repo already uses** for exactly this job in `utils/health.py`, `telegram/client.py`, and `telegram/ratelimit.py`. `sync/runner.py` was the outlier.

Deliberately **not** changed: `started_at` (L868), `finished_at` (L1017), the lock-meta stamp (L734), and `duration_sec` (L1109). Those are audit **timestamps**, where the wall clock is the correct source, and `duration_sec` stays consistent with `SyncResult.duration_sec` (`finished_at - started_at`), which derives from the same wall-clock pair.

**Invariant / Governance Impact:** none. This is the out-of-band `sync/` layer (I6-isolated — `gate.py`/`graph.py`/`mcp_hybrid_server.py` never import it). No graph edge, entry point, audit-convergence path, sanitizer pattern, auth surface, or soul path is touched. The `sync_started` → terminal-record audit pairing in this function is unchanged; the audit rows themselves keep their existing wall-clock timestamps.

---

## Types of changes

**Scope note:** Out-of-band agentic/fsconnect/sync layer.

---

## Benefits / why

- **Correctness of a lock-hold ceiling.** The budget is a safety control, not a stat. It bounds how long a single sync can block every other sync attempt. A control whose arithmetic depends on a clock a daemon can move is a control with an unbounded failure mode; `monotonic()` removes the dependency entirely.
- **Offline / air-gapped reliability specifically.** This is where it bites hardest. A machine that has been off the network reconnects and immediately takes a large NTP step — exactly the environment CyClaw targets, and exactly the moment a corpus sync is likely to run.
- **Consistency.** Brings the last elapsed-time holdout in line with the convention `utils/health.py` and `telegram/ratelimit.py` already follow, so the "timestamps use `time.time()`, durations use `time.monotonic()`" split is now uniform and reviewable at a glance.
- Behavior under a stable clock is bit-identical, so there is nothing to re-tune.

---

## Risks to monitor

- **Low.** No API, config key, audit field, or exit code changes. `sync_timeout_sec: 0` (the documented unbounded escape hatch) still short-circuits to `deadline = None` and per-attempt `timeout=None`.
- The one thing worth watching: `monotonic()`'s epoch is arbitrary and process-local, so any *future* code that tries to persist or compare this `deadline` across processes would be wrong. It is a local, function-scoped value today and should stay that way.
- Windows `time.monotonic()` resolution on Python 3.12 is coarser than Linux's. It does not affect this code (the budget is measured in whole seconds, default 3600), and the new test drives a fake clock rather than the real one, so there is no resolution-dependent flake.
- Detection of drift: `tests/test_sync_runner.py::test_run_sync_budget_deadline_is_measured_on_the_monotonic_clock` pins the arithmetic to the monotonic clock. It was confirmed to **fail** against the pre-fix code (all three attempts got ~10.0s, unaffected by the fake clock) and pass after.

---

## Checklist

- [x] This change preserves all 6 security invariants and I6 module isolation — `invariant-guard` re-run, 35 passed / 0 failed
- [x] Full validation run: `GROK_API_KEY=dummy pytest tests/ -q` green (exit 0, full suite), `ruff check --select E,F,I,B,C4,UP,S` clean, `flake8` (wemake) clean on both touched files
- [x] No new dependency (stdlib `time` only — already imported)
- [x] Commit message follows the convention

---

## Further comments

**ELI5:** There are two kinds of clock. One tells you *what time it is* ("3:41pm") and can be corrected — moved forwards or backwards — when the machine syncs with a time server. The other just counts *how long since the computer booted* and only ever goes up. The sync job's "you get 60 minutes, total, and then you must release the lock" stopwatch was being read off the first kind. If the clock got corrected backwards by ten minutes halfway through, the stopwatch quietly gave the job ten extra minutes of holding the lock; corrected forwards, it cut the job off early and skipped the post-sync integrity check. This moves the stopwatch to the second kind of clock. The audit log's "this sync started at 3:41pm" entries stay on the first kind, because there a human reading the log genuinely wants to know what time it was.

**Provenance:** found by `/CyClaw-Optimize` on `main` @ `1559871`, then independently confirmed by reading the surrounding retry/budget logic rather than taken on the scan's word.

**Merge topology:** independent — cut from `origin/main`, no shared-file dependency on any other branch. It is the only PR from this optimize run; the other candidate (adding a traversal cap to `agentic/fsconnect/quota.py::_walk_usage`) was **rejected on verification**, because capping that walk would undercount usage and silently defeat the fail-closed quota accounting it exists to feed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PfSahHUBDVvyfjJFVvvknT)_


cgfixit:
`windows-latest` is red here, and it is not this PR.

```
FAILED tests\test_harness_api_keys.py::test_file_is_written_owner_only
E   AssertionError: secrets file is 0o666, not 0600
```

This PR's diff touches `sync/runner.py` and `tests/test_sync_runner.py` only — nothing under `harness/`. The failure is already present on `main` itself: `ci.yml` run [31914323495](https://github.com/cgfixit/CyClaw/actions/runs/31914323495) @ `1559871` (the #950 merge commit) is `failure` for the same assertion. Every branch cut from that tip inherits it. The leg is `continue-on-error`, so it surfaces without blocking the merge button.

Root cause: Windows `os.chmod()` honours only the read-only bit and cannot express an owner-only POSIX mode, so the assertion is unsatisfiable on that platform rather than newly broken.

Fixed separately in **#952**, which also corrects the four places that claimed "mode 600" without a platform qualifier. Merge #952 first and this PR's Windows leg will go green on the next run; the remaining checks here are unaffected either way.

---

cgfixit:
Merged `claude/fix-windows-env-keys-mode-test` (#952) into this branch — `windows-latest` was the only red check, all others were already green. That leg should now pass on the new run against `fc7d91d`.

`#952` still stands on its own for `main` (its own CI is fully green there too) — this merge just unblocks this PR without waiting on merge order.

---

PR #954, opened 2026-08-16
[deps] Add websockets to pyproject.toml and fix pip-audit torch comment

cgfixit:
## What
- Add `websockets==15.0.1` to `pyproject.toml` `[project.dependencies]` so the direct import from `langgraph-sdk` is explicit and legacy/no-deps installs cannot strand the app on websockets 12.x.
- Fix the stale comment in `.github/workflows/pip-audit.yml` that claimed torch is "intentionally NOT pinned in requirements.txt"; `requirements.txt` now pins `torch==2.13.0+cpu`.

## Why / benefit
Manifest consistency: `requirements.txt` and `constraints.txt` already pin websockets, but `pyproject.toml` omitted it, making editable/`uv` installs differ from the legacy path. The pip-audit comment contradicted the file it referenced.

## Risk to monitor
- Verify editable installs still resolve websockets 15.0.1 via constraints.txt.
- No runtime behavior change; websockets was already being installed transitively.

## Verification
- `python .claude/skills/dep-guard/check_deps.py` → 0 failures, 0 warnings
- `python -c "import tomllib"` / `yaml.safe_load` parse checks → OK

### DIFF
--- a/.github/workflows/pip-audit.yml
+++ b/.github/workflows/pip-audit.yml
@@ -85,11 +85,9 @@ jobs:
       - name: Download torch wheel (cache miss only)
         if: runner.os != 'macOS' && steps.torch-wheel-cache.outputs.cache-hit != 'true'
         shell: bash
-        # torch is intentionally NOT pinned in requirements.txt — see its header:
-        # a bare resolve pulls the ~2.5 GB CUDA build from PyPI on Linux. Fetch
-        # the lean CPU wheel from the PyTorch index (matching ci.yml and the
-        # documented CVE-2025-32434 >=2.6.0 floor); --no-deps avoids pulling the
-        # default CUDA build.
+        # torch is pinned in requirements.txt as the CPU build; fetch the lean
+        # wheel from the PyTorch index (matching ci.yml and the documented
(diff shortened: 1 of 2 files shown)


cgfixit (commented):

**Verdict: safe**

### Invariant matrix
I1–I6 + G1–G5: unchanged. Manifest/comment only. No runtime graph/gate/soul change. Making `websockets==15.0.1` a direct `pyproject.toml` pin matches the existing `requirements.txt` / `constraints.txt` pin so uv/editable cannot silently land on websockets 12.x via langgraph-sdk.

### Template
What/Why/Risk/Verify present. Missing Types/Checklist/Merge order/Base (acceptable for deps/infra; still fill Merge order so the 953–957 parallel batch has a number).

### Findings
- pip-audit.yml change is a comment correction only (torch *is* pinned as `2.13.0+cpu` in requirements.txt). Good honesty.
- No new license, secret, or workflow key.

### Did not run
`dep-guard` locally on this SHA (PR body claims 0 failures / 0 warnings). CI on this PR is green (58 checks, 0 fail).

PR #955, opened 2026-08-16
[harness] Audit-log /api/keys writes with names only

cgfixit:
## What
- Import `audit_log` in `harness/server.py` and emit a `harness_api_keys_updated` audit event from `POST /api/keys`.
- The event carries only the key names (`written["written"]`), matching the existing route docstring and `env_keys.write_keys` contract.
- Add `test_post_emits_audit_event_with_names_only` to verify the event is emitted and that no secret value leaks into it.

## Why / benefit
Security/auditability: every other security-relevant surface (`gate.py`, `sync/`, `agentic/`) writes to `logs/audit.jsonl`. The harness key panel was the outlier, leaving credential-management actions invisible to the audit trail.

## Risk to monitor
- Verify the audit line does not contain key values (the new test asserts this).
- No change to response shape or file-write behavior.

## Verification
- `GROK_API_KEY=dummy pytest tests/test_harness_api_keys.py -q` → 20 passed, 1 skipped

### DIFF
--- a/harness/server.py
+++ b/harness/server.py
@@ -84,7 +84,7 @@ from llm.client import ResolvedLocalBackend, resolve_local_backend
 from schemas.api import AuthCreateUserRequest, AuthLoginRequest, AuthSetPasswordRequest, AuthSetRoleRequest
 from utils.auth import require_api_key
 from utils.errors import AgenticError
-from utils.logger import _get_config, redact_sensitive
+from utils.logger import _get_config, audit_log, redact_sensitive
 from utils.ops_runner import OpsError, OpsResult, run_agentic_op
 from utils.ratelimit import RateLimiter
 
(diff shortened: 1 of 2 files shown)


cgfixit (commented):

**Verdict: safe** (OOB harness; I6 direction is allowed: harness → `utils.logger`)

### Invariant matrix
| Inv | This PR |
|-----|---------|
| I1–I3, I5 | untouched |
| I4 | stronger: `/api/keys` now emits `harness_api_keys_updated` |
| I6 | harness may import `utils`; core still does not import harness |

### Security
`audit_log({"event": "harness_api_keys_updated", "keys": written["written"]})` plus `test_post_emits_audit_event_with_names_only` asserting `_SECRET not in str(event)` is the right fail-closed shape. Do not later add values, fingerprints of values, or raw query text to this event.

### Template
What/Why/Risk/Verify present. Missing Types/Checklist/Merge order/Base. Not core-path so the bot did not require an invariant paragraph; a one-liner that I6 is preserved would still help.

### Did not run
Focused `tests/test_harness_api_keys.py` locally. CI on this PR is green (48 checks, 0 fail).

PR #956, opened 2026-08-16
[test] Harden harness api_key_optional bypass parity with gate.py

cgfixit:
## What
Add two behavioral tests to `tests/test_harness_auth.py` that verify `security.api_key_optional` bypass is still denied for:
- a non-loopback socket peer (mirrors `gate.py` `_is_loopback_peer`);
- a request carrying a reverse-proxy forwarding header (mirrors `gate.py` `_looks_proxied`).

The existing `test_matches_gate_auth_semantics` only greps source text for shared strings; these tests close the actual bypass-condition gap.

## Why / benefit
The source-text parity test would not catch drift in the bypass logic. These tests assert the three necessary conditions (loopback peer, no proxy headers, plus the existing cross-site guard) are enforced the same way in `harness/server.py` as in `gate.py`.

## Risk to monitor
- None: test-only change.

## Verification
- `GROK_API_KEY=dummy pytest tests/test_harness_auth.py -q` → 142 passed

### DIFF
--- a/tests/test_harness_auth.py
+++ b/tests/test_harness_auth.py
@@ -236,6 +236,44 @@ def test_api_key_optional_false_matches_the_default(cfg, monkeypatch):
     assert resp.status_code == 401
 
 
+def test_api_key_optional_true_still_refuses_non_loopback_peer(cfg, monkeypatch):
+    """The bypass is peer-based, not Host-based: a non-loopback socket peer
+    must still see 401 even when the flag is on. Mirrors gate.py's
+    _is_loopback_peer check on /soul/* and /ops/*."""
+    import copy
+
+    monkeypatch.delenv("CYCLAW_API_KEY", raising=False)
+    real_get_config = harness_server._get_config
+
+    def _patched(path: str) -> dict:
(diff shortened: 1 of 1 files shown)


cgfixit (commented):

**Verdict: safe** (test-only; fail-closed auth posture)

### Invariant matrix
I1–I6 unchanged. No production code. Tests pin harness `api_key_optional` bypass to the same two conditions as `gate.py`: loopback peer (`_is_loopback_peer`) and no forwarding header (`_looks_proxied`). That is the right gap to close — the existing source-grep test would not catch a logic drift.

### Findings
- Non-loopback `TestClient(..., client=("192.168.1.100", 51234))` → 401: good.
- `X-Forwarded-For` → 401: good.
- Still CSRF-headed; does not weaken the key-required default.

### Template
What/Why/Risk/Verify present. Missing Merge order/Base. Test-only so Types/Checklist can stay light.

### Did not run
`tests/test_harness_auth.py` locally (PR body: 142 passed). CI on this PR is green (48 checks, 0 fail).

PR #953, opened 2026-08-16
[config] Align local LLM model tag and remove dead audit placeholders

cgfixit:
## What
- Align `models.local_llm.model` with `guardrails.model` by changing `qwen3.8:27b-mlx` → `qwen3.8:27b`.
- Remove the unprofessional inline comment that accompanied the mlx tag.
- Tighten the timeout comment to reflect the shipped default.
- Remove four dead `logging.audit_fields` placeholder keys (`include_top_score`, `include_retrieval_mode`, `include_online_escalated`, `include_model_used`) that are not read anywhere in the codebase.

## Invariant / Governance Impact
- I1 RAG-first: unchanged (no graph / retrieve change).
- I2 topology=policy: unchanged (no edges).
- I3 triple-gate: unchanged (local model **id** only; hybrid + enabled + confirm still required for external).
- I4 audit convergence: unchanged. The four dropped keys were commented placeholders, not live `audit_fields` consumed by `utils/logger.py`.
- I5 soul: untouched.
- I6 isolation: `config.yaml` only; no new core↔OOB import.
- Evidence: comment/tag cleanup; CI on this PR was already green vs `c5d752e4`.

## Why / benefit
Config correctness: the mlx tag was Apple-Silicon-specific and drifted from the "keep in sync" default in `guardrails.model`. The dead placeholders made `config.yaml` noisier than the single-source-of-truth contract requires.

## Risk to monitor
- Any operator relying on the mlx tag locally will need to `ollama pull qwen3.8:27b` (or override config) instead.
- Verify model-pin drift detection no longer flags the shipped local model as unhealthy.

## Verification
- config-guard parse / relational check → OK (PR-author report).
- Focused pytest reported 169 passed (PR-author report).
- GitHub checks on this draft were green before #957 landed.

## Merge order
- **This PR:** last remaining draft of the 953–957 parallel batch.
- **Already on main:** #954, #955, #956, then #957 (`b9c672bb`, 2026-08-16T02:53:44Z).
- **Topology:** parallel / no shared paths with 954–957. Safe to merge after #957 (now done).
- Do not reopen 954–957 for template nits.

## Base
- GitHub base branch: `main`
- Forked from: `origin/main@c5d752e41831154c94fb397b2a3d0462ed7949d7`
- `origin/main` now: `b9c672bbd68752b21ec40c06bd85e64e20dae00d` (#957 squash)

PR #957, opened 2026-08-16
[workflow/sync] Tag claude-action version and simplify event_counts

cgfixit:
## What
- Add the missing `# v1.0.183` version comment to the pinned `anthropics/claude-code-action` SHA in `.github/workflows/claude.yml`, matching every other pinned action in the repo.
- Simplify `SyncResult.event_counts()` in `sync/runner.py`: the dict is initialized with all three known keys, so `.get(ev.kind, 0)` is redundant.
- Add `test_sync_result_event_counts` to pin the helper's behavior.

## Why / benefit
Drift detection: version comments make SHA bumps reviewable. The `event_counts` simplification removes a misleading fallback for an impossible key, and the test closes a small coverage gap.

## Risk to monitor
- None: the workflow change is a comment only; the sync change is behaviorally equivalent for the three documented event kinds.

## Verification
- `GROK_API_KEY=dummy pytest tests/test_sync_runner.py::test_sync_result_audit_dict_no_secret_fields tests/test_sync_runner.py::test_sync_result_event_counts -q` → 2 passed

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -48,7 +48,7 @@ jobs:
           persist-credentials: false
 
       - name: Run Claude Code
-        uses: anthropics/claude-code-action@be7b93b1907a4abad570368f3c74b6fe3807510b
+        uses: anthropics/claude-code-action@be7b93b1907a4abad570368f3c74b6fe3807510b  # v1.0.183
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/sync/runner.py
+++ b/sync/runner.py
@@ -234,7 +234,7 @@ class SyncResult:
     def event_counts(self) -> dict:
         counts = {"added": 0, "modified": 0, "deleted": 0}
         for ev in self.events:
-            counts[ev.kind] = counts.get(ev.kind, 0) + 1
+            counts[ev.kind] += 1
         return counts
 
     def to_audit_dict(self) -> dict:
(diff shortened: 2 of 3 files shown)


cgfixit:
Automated fix: windows-latest failed test_run_sync_check_timeout_is_clipped_to_remaining_budget with assert 3600.0000000000005 <= 3600. Pre-existing Windows time.time() overshoot; clipped remaining budget to cfg.sync_timeout_sec. Not caused by the event_counts change.

cgfixit (commented):
## /pr-review vs origin/main `c5d752e4` (includes babysit CI fix `0d0b494c`)

**Verdict: safe** after the Windows budget clip. Original nits + CI flake are both OOB `sync/` / workflow comment.

### Invariant matrix
| Inv | This PR |
|-----|---------|
| I1–I3, I5 | untouched |
| I4 | unchanged (`event_counts` is a helper; audit still emits `sync_completed` / `sync_failed`) |
| I6 | `sync/` is OOB; no new core import |

### Original change
- `claude.yml` `# v1.0.183` comment on an already-pinned SHA: comment-only.
- `SyncResult.event_counts()` `counts[ev.kind] += 1` is equivalent for the three parser kinds (`added`/`modified`/`deleted`). Slightly less defensive if a new kind appears — the new test pins the three-key contract.

### CI (class A, pre-existing flake — not caused by event_counts)
`windows-latest` / `test_run_sync_check_timeout_is_clipped_to_remaining_budget`:
`assert 3600.0000000000005 <= 3600`

Deadline is already `time.monotonic() + sync_timeout_sec` (landed earlier). IEEE-754 still allows `(t+3600)-t` to be a hair over 3600 on Windows. #953–#956 passed the same test; this branch hit it.

Babysit commit `0d0b494c` clips `run_timeout` and `remaining_for_check` with `min(..., float(cfg.sync_timeout_sec))`. Does **not** loosen the test. That is the right invariant (check phase never gets more than the configured ceiling).

### Template
What/Why/Risk/Verify present. Missing Types/Checklist/Merge order/Base. Not core-path.

### Did not run
I did not re-run the full suite locally. Child ran ruff + focused pytest + `verify_ci_emulation.py` (exit 0) before push. Re-run of `windows-latest` is **in progress** on `0d0b494c` — not green yet.

PR #995, opened 2026-08-19
[docs] - add keychain integration review vs mcporter

cgfixit:
## Proposed changes

Adds `docs/audits/KEYCHAIN_INTEGRATION_REVIEW_2026-08-19.md` — a comparative review of CyClaw's macOS Keychain secret handling against `openclaw/mcporter`, requested after the v0.8.0 release announcement.

The review first corrects the premise it was asked under: **mcporter has no OS-keychain secret storage**, at `v0.8.0` or at current `0.13.7`. Its runtime credential store is a cleartext JSON vault at `<XDG_DATA_HOME>/mcporter/credentials.json`, protected only by `DEFAULT_ATOMIC_FILE_MODE = 0o600`; non-OAuth secrets are never stored, only resolved from `process.env` at connection time. The only `keychain` hits anywhere in the tree are Developer ID signing (`scripts/codesign-native.sh`, `docs/RELEASE.md`). Verified by searching `src/`, `scripts/`, and `docs/` at both the `v0.8.0` tag and `60dfe26`.

The useful comparison is therefore not backend choice — CyClaw is ahead there — but the handling discipline around whichever backend is chosen. Four mcporter techniques have no CyClaw counterpart: write-then-verify (`assertSecurePath`), an explicitly-constructed and read-back-verified Windows ACL, cross-process file locking on every credential mutation (`withFileLock`), and a documented split between fail-open caches and fail-closed security state.

Eight findings against CyClaw follow, each paired with mcporter's counterpart. The most severe: `_keychain_get()` discards `security`'s exit code and stderr via `2>/dev/null || true`, so a locked keychain is indistinguishable from a missing item — and in the `CYCLAW_API_KEY` branch that can silently **regenerate the key over a perfectly good Keychain item**, reachable unattended through the `--schedule-rotate` LaunchAgent.

**Invariant / Governance Impact:** none. Documentation only — no code, config, or test changes. The six invariants and I6 module isolation are untouched; nothing in this PR is reachable from `gate.py`, `graph.py`, or `mcp_hybrid_server.py`.

---

## Types of changes

**Scope note:** Docs + audits.

---

## Benefits / why

- Records the negative result so nobody re-runs this comparison expecting to find keychain code in mcporter that isn't there.
- Surfaces one failure mode (Finding 1) that can leave an operator holding a `CYCLAW_API_KEY` the running server rejects, with no error explaining why, from an unattended scheduled job.
- Names two whole classes rather than instances: `TMPDIR` interpolated into a Tcl heredoc (Finding 2), and an unpinned keychain search list (Finding 6). Both close with one-line changes.
- Records what CyClaw already does *better* than mcporter — argv-free secret storage via bare `-w`, encrypted-at-rest with an ACL, the fail-closed launch wrapper, and the honest VERIFY ON REAL HARDWARE comments — so none of it gets "simplified" away later.
- Explicitly rules out a latency concern rather than inventing one: the read path is one `security` exec per secret per process start, paid at launchd job start, never in a request path. The doc argues against adding caching.

---

## Risks to monitor

- **This PR changes no behavior.** The risk is the opposite: findings sitting unactioned. Findings 1, 3, and 4 can each cost a working deployment.
- Every code-touching remedy lands in `macos/setup-cyclaw-keys.sh` and `macos/cyclaw-keychain-*.sh` — secret-handling paths, therefore **High tier under CLAUDE.md §7** and gated on explicit approval. None were made here.
- Findings 1 and 5 need on-hardware verification before anything is hardcoded: the `security(1)` item-not-found exit code, and whether `-U` replaces the trusted-application list on an item created before `-T` was introduced. Both are marked VERIFY ON REAL HARDWARE, matching the convention already in those scripts.
- Drift detection: the doc pins its sources (CyClaw `e8bea4d`, mcporter `60dfe26` / `0.13.7`, tag `v0.8.0`). If mcporter later adds a real credential backend, the premise-correction section goes stale and should be re-run rather than trusted.

---

## Checklist

- [x] This change preserves all 6 security invariants and I6 module isolation (docs-only; no code path touched)
- [x] Every claim is cited to a file and line in one of the two trees, or marked as needing hardware verification

Not applicable (docs-only): sandbox validation, agentic/fsconnect write-guard verification, architecture/threat-model updates.

---

## Further comments

**ELI5:** The ask was "read how this well-regarded tool does macOS Keychain and tell me what I could do better." It turned out the tool doesn't use the Keychain at all — it keeps its secrets in a normal file that only you can read. So the review pivoted to the thing that actually transfers: not *where* you put the secret, but how carefully you handle it on the way in and out. CyClaw's storage is genuinely stronger. Its handling has gaps mcporter doesn't, and the worst one is that when CyClaw can't read the Keychain, it can't tell "the secret isn't there" apart from "I wasn't allowed to look" — and in one path it responds by making a new secret, which quietly breaks the server that's still using the old one.

Existing test coverage to extend rather than duplicate: `tests/test_cyclaw_keychain_scripts.py` (12 tests, fake `security` on PATH) and `tests/test_setup_cyclaw_keys.py` (20 tests). The fake-`security` fixture is the natural place to add exit-code cases for Finding 1.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mwevy41AAGRmkWQ1kXbZgb)_

### DIFF
--- a/docs/audits/KEYCHAIN_INTEGRATION_REVIEW_2026-08-19.md
+++ b/docs/audits/KEYCHAIN_INTEGRATION_REVIEW_2026-08-19.md
@@ -1,0 +1,326 @@
+# Keychain Integration Review — CyClaw vs. mcporter, 2026-08-19
+
+Comparative review of CyClaw's macOS Keychain secret handling against
+`openclaw/mcporter`, requested after the v0.8.0 release announcement. Sources:
+CyClaw at `e8bea4d`; mcporter cloned at `60dfe26` (package version `0.13.7`),
+with tag `v0.8.0` fetched separately for the version the request named.
+
+---
+
+## Premise correction: mcporter has no OS-keychain secret storage
+
+The comparison as posed does not have a counterpart on the mcporter side.
+mcporter does not put secrets in the macOS Keychain, at `v0.8.0` or at
+`0.13.7`. A case-insensitive search for `keychain`, `keytar`, `libsecret`,
+`security find-generic-password`, and `wincred` across `src/`, `scripts/`, and
+`docs/` returns, at `v0.8.0`, nothing at all, and at `0.13.7` only four hits,
+all in release engineering:
+
+- `scripts/codesign-native.sh` passes `--keychain-profile` / `--keychain` to
+  `notarytool`.
+- `docs/RELEASE.md` documents `NOTARYTOOL_KEYCHAIN_PROFILE` and the managed
+  `openclaw-developer-id-release.keychain-db` used for Developer ID signing.
+
+That is a build-time signing keychain, not a runtime secret store. The
+v0.8.0 release notes are about OAuth error handling, JSON output on fallback
+paths, JSONC config parsing, and daemon reliability — no credential-storage
+change is listed.
+
+## What mcporter actually does with credentials
+
+mcporter's runtime secret store is a plain JSON file. `getOAuthVaultPath()`
+(`src/oauth-vault.ts`) resolves to `<XDG_DATA_HOME>/mcporter/credentials.json`,
+falling back to `~/.mcporter/credentials.json`, holding OAuth access and
+refresh tokens, dynamic client registrations, PKCE code verifiers, and OAuth
+state in cleartext. Confidentiality rests entirely on filesystem permissions:
+`DEFAULT_ATOMIC_FILE_MODE = 0o600` in `src/fs-json.ts`, applied at
+`fs.writeFile` time with flag `wx` into a same-directory temp file that is then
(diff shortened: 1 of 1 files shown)


cgfixit:
"the worst one is that when CyClaw can't read the Keychain, it can't tell "the secret isn't there" apart from "I wasn't allowed to look" — and in one path it responds by making a new secret, which quietly breaks the server that's still using the old one."


Hahaha wow okay definitely need to revisit to at least log access/existence errors and the last part sounds more dramatic than it is but the failure mode for the "one path" needs to be fixed and code better understood next time I look at this

PR #1027, opened 2026-08-21
[harness] - pin the /tools catalog to live routes and cover /api/keys auth

cgfixit:
## Proposed changes

Ships **S1** of `docs/work/WAKU_HARNESS_TOOLS_PLAN.md`: a two-way contract test between `harness/tools_view.py`'s `_HARNESS_SURFACES` catalog (the data behind `GET /api/tools` and the `/tools` console command) and the routes `harness/server.py` actually registers. The drift was measured before this PR, not hypothetical: **24 catalog rows against 46 registered `/api/*` routes.**

Implementation turned up a worse instance than the plan anticipated. `/api/keys` (`harness/env_keys.py`'s managed-credential surface) was guarded in `harness/server.py` from the start but missing from **three** places at once:

| Place | State before this PR |
|---|---|
| `_HARNESS_SURFACES` catalog row | missing → invisible in `/tools` |
| `tests/test_harness_auth.py`'s `GUARDED` list | missing → its 401-on-missing-key behavior was **untested** |
| `CLAUDE.md`'s guarded-route count | said 26; ground truth is 28 |

Verified by AST-evaluating the `GUARDED` literal against a parse of `harness/server.py`: zero stale entries in that list, and the gap is exactly `GET`+`POST /api/keys`. One omission, three consequences, in the endpoint that fronts stored secrets.

### Changes

- **`harness/tools_view.py`** — two catalog rows (`web-search`, `keys`) for routes that existed but were uncataloged.
- **`tests/test_harness_auth.py`** — the two missing `GUARDED` entries for `GET`/`POST /api/keys`, closing the untested credential surface.
- **`tests/test_harness_tools_contract.py`** (new) — three checks nothing else in the suite covers:
  1. Every catalog row's declared method actually matches the live route's allowed methods. (Production `_as_harness`'s `wired` flag is **path-only** — it never checks method — so a row claiming `GET` on a POST-only route renders `●` today regardless.)
  2. Every registered `/api/*` route is cataloged or named in a commented `_CATALOG_EXEMPT` set (sub-actions of a cataloged family, and the whole `/api/auth/*` session+CSRF domain).
  3. Every route carrying the harness's API-key dependency chain has a `GUARDED` entry in `test_harness_auth.py`, walking the route's `dependant` tree by name (mirroring `gate.py`'s own `_dependant_call_names` technique, reimplemented locally rather than imported — `test_harness_isolation.py` enforces harness/gate never import each other).

  Both drift-catchers were proven, not just asserted: I reverted one catalog row and one `GUARDED` entry in turn, confirmed the new test failed each time with the expected message, then restored.
- **`CLAUDE.md`** — guarded-route count corrected 26 → 28, per the plan's own stated policy: fix the doc to match the test, never the reverse.

**Invariant / Governance Impact:** none. I1–I5 are gateway/graph properties this diff never touches. I6 holds — the new test imports `harness` only, never `gate`.

## Types of changes / Checklist:

- [x] Bugfix (non-breaking change which fixes an issue) — closes a real, unmitigated coverage gap on the credential endpoint
- [x] Invariant / Governance refinement (use this for changes that strengthen or evolve the 6 invariants, I6 isolation, or harness phases) — mechanically pins a catalog↔route contract that had none

**Scope note:** Out-of-band `harness/` console layer + its tests. Core graph/gate/soul path untouched.

## Benefits / why

- Closes a real security-relevant test gap: the one route family fronting stored secrets (`/api/keys`) had zero auth regression coverage until this PR.
- Turns a catalog that could silently drift forever into one a CI-style check enforces: adding or renaming a route now forces a deliberate cataloged-or-exempt choice instead of quietly rendering `○`/invisible in `/tools`.
- Corrects a prose claim in `CLAUDE.md` that `doc-sync`'s existing D5 check cannot see (it only covers `gate*.py` → docs, not the harness catalog or its auth-test list) — this PR's test is the mechanism that will keep it correct going forward.
- No new dependency, no new route, no runtime behavior change — pure test/catalog/doc tightening.

## Risks to monitor

- The new contract test's `_CATALOG_EXEMPT` / `_KEY_EXEMPT` sets are hand-curated judgment calls (sub-actions of a cataloged family, plus the `/api/auth/*` domain). A future route addition that *should* be cataloged but gets exempted instead would pass this test while still being invisible in `/tools` — the test can only catch omission, not a wrong exemption choice. Reviewer attention on new `_CATALOG_EXEMPT`/`_KEY_EXEMPT` entries going forward is the mitigation.
- `tests/test_harness_tools_contract.py`'s guard-contract check re-derives `gate.py`'s dependant-tree-walk technique locally rather than importing it (required by I6-style isolation between `harness` and `gate`, even in test code — see the module's own comment). If `gate.py`'s technique changes, this copy will not follow automatically.
- This PR's own verification could not run the full `pytest tests/ -q` suite: this session's egress policy blocks `download.pytorch.org`, so a full 3.12 venv with torch/chromadb could not be built. I built a project-scoped venv covering exactly the packages `tests/test_harness*.py` needs (fastapi, httpx, pyyaml, rank_bm25, numpy — all from `constraints.txt` pins) and ran the full harness corpus (~525 tests) plus `ruff`/`invariant-guard`/`doc-sync`/`config-guard` clean. The retrieval/agentic/gate test lanes were not exercised locally; this diff does not touch any file those lanes cover.

## Further comments

ELI5: the harness console has a small built-in "directory" (`/tools`) that's supposed to list every button/command it exposes. That directory was hand-maintained and had drifted out of sync with the actual code — two real features existed that the directory didn't know about, and one of them was the page for managing API keys, which also had no automated test checking that a stranger without the password gets rejected. This PR adds the two missing directory entries, adds the missing rejection test, and — the more important part — adds a new test that will automatically catch the *next* time this kind of drift happens, instead of relying on someone noticing by hand again.

**Verification evidence:**

```
ruff check --select E,F,I,B,C4,UP,S harness/tools_view.py tests/test_harness_auth.py \
    tests/test_harness_tools_contract.py          -> All checks passed!
python3 .claude/skills/invariant-guard/check_invariants.py   -> 35 passed, 0 failed  (exit 0)
python3 .claude/skills/doc-sync/doc_sync.py                  -> 0 drift item(s) found (exit 0)
python3 .claude/skills/config-guard/check_config.py          -> 0 failure(s), 1 pre-existing warning (exit 0)
GROK_API_KEY=dummy pytest tests/test_harness*.py -q          -> all green (~525 tests)
```

Plus the drift-proof (reverted, confirmed the new test failed, restored) for both the catalog-row check and the guard-contract check, described above.

---
_Generated by [Claude Code](https://claude.ai/code/session_014aBuPqxnRAsps7sjVBAfiQ)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -121,7 +121,7 @@ The four `/ops/*` endpoints reach out-of-band subsystems ONLY through
 `utils/ops_runner.py` (a `subprocess.run([...])` shim). They never import those
 subsystems.
 
-Every route marked **API key** above — plus the harness console's 26
+Every route marked **API key** above — plus the harness console's 28
 `guarded` routes (`harness/server.py`, including `/api/agent/run`/`push`/
 `publish`) — is gated by `require_api_key` (two independent implementations,
 one per app; see `utils/auth.py`'s module docstring for why). `config.yaml`'s
--- a/harness/tools_view.py
+++ b/harness/tools_view.py
@@ -99,6 +99,10 @@ _HARNESS_SURFACES: tuple[tuple[str, str, str, str, str], ...] = (
              "allowlist-only web fetch (off until /web on)"),
     _surface("web-fetch", "/web fetch", _POST, "/api/web/fetch",
              "GET one allowlisted URL; no crawl, no search engine"),
+    _surface("web-search", "/web search", _POST, "/api/web/search",
+             "grep allowlisted pages for a query (no search engine)"),
+    _surface("keys", "/api", _GET, "/api/keys",
+             "managed credential status (masked tail only, never a value)"),
 )
 
 
--- a/tests/test_harness_auth.py
+++ b/tests/test_harness_auth.py
@@ -72,6 +72,21 @@ GUARDED = [
     ("post", "/api/agent/runs/" + "0" * 32 + "/push", {}),
     ("post", "/api/agent/runs/" + "0" * 32 + "/publish", {"reason": "r", "confirm": True}),
     ("post", "/api/agent/runs/" + "0" * 32 + "/discard", {}),
+    # The managed credential surface (harness/env_keys.py). Both halves were
+    # guarded in harness/server.py from the start but absent from this list, so
+    # the one route family that fronts stored secrets was the only guarded one
+    # with no 401 regression test -- the same omission left it out of
+    # tools_view's catalog and undercounted CLAUDE.md's guarded-route total.
+    # The read is guarded too: which providers an operator has configured is
+    # worth withholding even though no secret is returned.
+    ("get", "/api/keys", None),
+    # ApiKeysRequest takes a name -> secret MAPPING, not flat key/value fields,
+    # and forbids extras -- a {"key":..., "value":...} body would 422 before the
+    # dependency chain proved anything. GROK_API_KEY is in env_keys.MANAGED_KEYS,
+    # so this body is one that would succeed if it got through, per this list's
+    # contract above. CYCLAW_HOME is the tmp home from the cfg fixture, so the
+    # accepts-the-right-key case writes its .env under tmp_path, not ~/.CyClaw.
+    ("post", "/api/keys", {"keys": {"GROK_API_KEY": "x"}}),
 ]
 
 # Left open on purpose: the console must be able to boot and tell the operator it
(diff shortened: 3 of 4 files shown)


cgfixit:
@codex review

PR #1067, opened 2026-08-22
[docs] - reconcile every README + AGENTS.md against current main

cgfixit:
## Proposed changes

A full `/doc-sync` pass across the whole documentation tree, not just the root
`README.md`: every subdirectory `README.md`, `AGENTS.md`, and a new
`docs/spend/README.md` for the previously-undocumented online-LLM token
ledger. The mechanical checker (`.claude/skills/doc-sync/doc_sync.py`, D1-D6)
reported **zero drift both before and after this PR** — everything here is
prose-level drift the checker cannot parse, plus two numbers (a timeout and a
test-file count) that fall outside its six checks.

Four parallel read-only audit agents verified every claim against the actual
code/config before anything was edited; nothing here is inferred or guessed.

**Real contradictions fixed** (code/config confirmed as source of truth):
- `CLAUDE.md`'s request-flow prose said the graph invoke has a 660s timeout;
  `config.yaml`'s `api.graph_timeout_sec` is `780`, and `CLAUDE.md`'s own
  load-bearing table already said `780` — the prose disagreed with the table.
- `CLAUDE.md` and `tests/README.md` both said 181 `test_*.py` files; the tree
  has 183.
- `README.md`'s unauthenticated-route list omitted `GET /auth/setup-status`
  (contradicting the README's own item 12) and didn't explain that
  `POST /auth/bootstrap-password` is gated on a loopback peer + same-origin
  check rather than being genuinely open.
- `docs/README.md` described auth as "Stage 4 TLS pending"; all six stages
  are landed per `docs/AUTHENTICATION_DESIGN.md`, TLS included.
- `static/README.md` claimed `auth_admin.js` loads on either console server.
  `gate.py` carries the repo's only `/static` mount, so `harness.html`'s
  script tag 404s on `:8790` — documented as the actual behavior. (Doc pass
  fixes docs, not behavior — the underlying gap is worth a follow-up.)
- `guardrails/README.md`'s status table named only `check_soul_leak` as
  unenforced; `check_jailbreak` and both topical rails are equally unbuilt
  in the offline floor, which implements exactly two checks.
- `scripts/README.md` omitted `check-pr-template.sh`'s exit code 2.
- `.claude/README.md` asserted every hook-referenced file exists; the
  registered `UserPromptSubmit` hook points at a `context_gate.py` that was
  deleted from `main` (silently no-ops via `2>/dev/null || true`, which is
  why it went unnoticed). Flagged here, not fixed — wiring/unwiring a hook is
  a `settings.json` change per `CLAUDE.md` §7 and needs explicit sign-off.

**Undocumented features now covered:**
- New `docs/spend/README.md` for `utils/spend.py` / `logs/spend.jsonl` —
  previously the only trace in 1,300+ lines of README was a tree-comment
  parenthetical. Linked from the root README (new item 18), `utils/README.md`,
  `llm/README.md`, and the docs index.
- `agentic/README.md` gains the UNSLOP prose-quality rail and `selftest.py`.
- `agentic/deepagent_github/README.md` gains `handoff.py` to its Live list
  (sanitizes every outbound cloud prompt) and a table classifying all 12
  modules live/retired instead of leaving 8 unstated.
- `harness/README.md` gains its two first-run auth routes and `env_keys.py`.
- `data/README.md` gains five configured paths, including the secret-bearing
  `agentic/workspaces` clones and the auth database.
- `llm/README.md` now states that under the shipped config both external
  clients are constructed at boot (mode `hybrid`, both providers `enabled`),
  not just under some hypothetical future config.
- Smaller gaps closed in `sync/`, `opentweet/`, `macos/`, `.githooks/`,
  `schemas/`, `utils/` READMEs and `AGENTS.md` (a dead file reference, a
  `static/` omission, and stating the `codex/` branch prefix its own
  audience needs).

**Copy-paste safety:** bare `<placeholder>` tokens inside shell-labeled fences
parse as redirects (`<foo>` = "read stdin from foo, then redirect stdout"), so
a documented command either syntax-errors on paste or silently eats the next
argument. Quoted 32 such placeholders across `README.md`, `AGENTS.md`,
`CLAUDE.md`, `agentic/README.md`, the CyClaw-Optimize skill (all three
copies), and `copilot-instructions.md`; converted two non-runnable reference
tables (a slash-command list, an Ollama command table) to `text` fences.
Verified with `bash -n` over every shell-labeled fence in the repo: 344
blocks, zero failing.

**Invariant / Governance Impact:** none. This PR touches only Markdown
documentation — no code, no `config.yaml`, no graph edges, no `settings.json`.
The one behavioral gap it surfaces (`static/README.md`'s harness-console
finding) is described accurately, not silently patched.

## Types of changes
- [x] Documentation Update

Scope note: Docs only.

## Benefits / why
Six of the fixes above are contradictions a reader would act on incorrectly —
a stale timeout, a wrong route list, a wrong TLS status, an incomplete
security-rail status table. The rest close real coverage gaps (spend
tracking had no dedicated doc at all) or remove copy-paste traps from the
canonical install/command blocks every skill points readers at.

## Risks to monitor
- The `.claude/README.md` note about the dangling `context_gate.py` hook
  reference is a flag, not a fix — a human should decide whether to restore
  the script or unwire the hook.
- The `static/README.md` finding (harness console can't load the shared
  Users-panel script) documents a real behavior gap; worth a follow-up issue
  if the harness console is meant to have that panel.

## Checklist
- [x] `python3 .claude/skills/doc-sync/doc_sync.py` — 0 drift, before and after
- [x] Every shell-labeled fence in the repo verified with `bash -n` (344/344 pass)
- [x] Every relative Markdown link in changed files resolves
- [ ] Full sandbox validation — not applicable; no Python/config touched

## Further comments
Docs-only PR — lighter checklist per the template's own guidance. Independent
of the three CyClaw-Optimize PRs (#1064, #1065, #1066) from this session; no
shared files.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D9biQCFgkENyomvp2uWHgx)_


cgfixit:
The human did decide to unwire the hook. If anything shouldn't I have picked the good model first and even so I should have monitored the effects of how "confused" it gets half way through when getting all the sudden better but not really at coding (fable vs opus)  way closer after since it's going off of delayed info anyway to know when the context threshold is met

cgfixit:
> The human did decide to unwire the hook. If anything shouldn't I have picked the good model first and even so I should have monitored the effects of how "confused" it gets half way through when getting all the sudden better but not really at coding (fable vs opus)  way closer after since it's going off of delayed info anyway to know when the context threshold is met

Also lol I don't use the wrong effect when I don't even think about it. Unclear if that's a broken clock right twice a day.

cgfixit:
Got it on the hook decision — unwiring is the right call given it's silently no-op'ing anyway.

I tried to make that edit here (drop the `UserPromptSubmit` entry in `.claude/settings.json` pointing at the deleted `context_gate.py`), but this session's local permission classifier blocked it — settings.json/hooks get treated as high-sensitivity even with the go-ahead in this comment, since it's not the same as an interactive approval prompt. I don't want to work around that block.

So: you'll need to make that specific edit yourself (or approve it in an interactive session where the permission prompt can actually fire) — remove the `UserPromptSubmit` block at `.claude/settings.json:40-50`. I can redo the `.claude/README.md` note once it's landed to stop describing it as an open flag.

On the model-routing/context-threshold point — noted, not going to pretend I have a clean answer on quantifying "confused halfway through" against a delayed context-usage signal. If you want to dig into that separately, happy to, but didn't want to let it derail this PR's actual open item.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D9biQCFgkENyomvp2uWHgx)_

---

PR #1103, opened 2026-08-26
[security] - malformed Origin header crashes same-origin checks with 500 instead of 403 (NEEDS EXPLICIT REVIEW)

cgfixit:
## ⚠️ High-tier change — please review carefully before merging

Per `CLAUDE.md` §7, editing a graph edge / auth / sanitizer pattern is explicitly called out as **High tier — "Stop and ask first."** This PR edits the same-origin/cross-site checking logic in `gate.py`, `gate_auth.py`, and `harness/server.py`. I've isolated it into its own PR — separate from the four other, lower-risk fixes from the same optimize scan (#1099–#1102) — specifically so it gets deliberate human attention rather than riding along with routine work. **Please review this one closely before merging**, even though the diff is small and, I believe, purely defensive.

## Branch naming
`claude/cyclaw-optimize-origin-header-fix` — Claude Code driver.

## Title
`[security] - malformed Origin header crashes same-origin checks with 500 instead of 403 (NEEDS EXPLICIT REVIEW)`

---

## Proposed changes

Found in a read-only optimization scan, then independently re-verified against the actual running code — a second adversarial pass tried to refute the finding, confirmed it, and additionally found a **third affected call site** (`harness/server.py`) the original scan missed entirely.

**The bug:** all three same-origin/cross-site checks call `urllib.parse.urlparse()` on the raw, attacker-supplied `Origin` header. `urlparse()` itself raises `ValueError("Invalid IPv6 URL")` for an Origin containing an unbalanced `[` or `]` — I verified this directly: `'http://[evil'`, `'http://[::1'`, and `'http://a]b'` all raise **from the `urlparse()` call itself**, not from a later `.hostname`/`.port` access. `gate_auth.py`'s own comment stated the opposite premise ("`urlparse()` itself never raises, but reading `.port` does") and only guarded `.port` — leaving the `urlparse()` call one line above it completely unguarded. `gate.py` and `harness/server.py` had no guard at all. The exception escapes the dependency/handler, and Starlette turns it into an **unhandled HTTP 500**, instead of the 403 each check exists to return.

**Reachability, verified per site:**
- **`gate.py`'s `_looks_cross_site`**: reachable via `POST /index/build`, after the loopback-peer check — so only from a loopback-adjacent adversary (e.g. a malicious page in the operator's own browser), which is exactly the CSRF threat this check exists to stop.
- **`gate_auth.py`'s `_enforce_same_origin`**: reachable on **15 `/auth/*` routes regardless of `auth.enabled`**, because `register_auth_routes` runs unconditionally and FastAPI resolves dependencies before a handler's own `_require_enabled()` 503 check. So on the **shipped default** (`auth.enabled: false`), an unauthenticated, bodyless `POST /auth/login` with a malformed Origin returns `500` instead of the `503` the disabled feature should report — no credentials or valid body required.
- **`harness/server.py`'s `_enforce_same_origin`**: reachable on every guarded harness route (`gate_auth.py`'s own docstring says it "Mirrors `harness/server.py`'s `_enforce_same_origin` exactly") — this site was missed by the initial scan and caught only by the adversarial re-verification pass.

**The fix:** wrap the `urlparse()` call itself (not just the already-guarded `.port` access) at each of the three sites, catching `ValueError` and **failing closed** — treating an unparseable Origin as cross-site/cross-origin, i.e. exactly the same 403 each check already returns for a mismatched-but-parseable Origin. `gate_auth.py`'s stale comment is corrected in place. Three new regression tests (one per site) assert 403 with the existing `CROSS_SITE_BLOCKED`/`CROSS_ORIGIN_BLOCKED` code, not a 500; the existing `gate_auth.py` malformed-origin test is extended with the new case (parametrized) rather than duplicated.

**What this PR deliberately does NOT change:** no allow/deny condition is relaxed, removed, or added. A well-formed Origin's routing through all three checks is byte-for-byte unchanged — I only added a `try/except ValueError` around a call that previously had none (or, in `gate_auth.py`'s case, extended the existing guard to cover the earlier-raising call too).

**Invariant / Governance Impact:** none of the six invariants change shape. This strengthens (never weakens) the existing CSRF/cross-origin denial behavior — an unparseable Origin now gets the deny path it should have gotten all along, instead of a crash. I is not touched (no core-vs-out-of-band import changes); the `harness/server.py` fix is the harness's own independent copy of the same check pattern, not a new dependency on gate/graph.

---

## Types of changes
- [x] Invariant / Governance refinement (hardens an existing auth-adjacent check's failure mode; does not change what it allows or denies)

**Scope note:** Core graph/gate path (`gate.py`, `gate_auth.py`) + out-of-band harness console (`harness/server.py`).

---

## Benefits / why

Closes an unauthenticated crash-to-500 on a security-relevant code path, most notably the `/auth/*` exposure that fires regardless of whether the auth feature is even enabled. A 500 here is strictly worse than the 403/503 it should be: it's noisier (a logged traceback for what should be routine input validation), and — more importantly — it means the check's *fail-closed* guarantee currently depends on `urlparse()` never raising, which this PR proves is false.

---

## Risks to monitor

This is the change in this batch of five PRs most deserving of careful review, precisely because it touches auth-adjacent logic. The diff is intentionally minimal and I've argued above why I believe it's behavior-preserving on every previously-working path — but I'd specifically ask a reviewer to independently confirm:
1. No other code path relies on this `ValueError` propagating (I grepped for callers catching `ValueError` around these three functions specifically; none exist).
2. The fail-closed direction is correct in all three cases (an unparseable Origin should never be treated as same-origin — I believe this is uncontroversial, but it's exactly the kind of judgment call this repo's own rules say should get a second set of eyes).

---

## Checklist

- [x] I have read the latest architecture guide / `SECURITY.md` conventions for this repo
- [x] This change preserves all 6 security invariants and I6 module isolation — explicit invariant matrix: **before** — a malformed Origin crashes the process (undefined failure mode); **after** — a malformed Origin is treated as cross-site/cross-origin, the same deny path a well-formed-but-mismatched Origin already takes. No graph edge, no `banned_patterns`, no auth *decision* logic changed — only a previously-undefined error path is now defined, and defined toward the safer (deny) outcome.
- [x] `GROK_API_KEY=dummy pytest tests/test_gate_auth.py tests/test_gate_index_build.py tests/test_gate_query_auth.py tests/test_harness_auth.py tests/test_terminal_contract.py -q` green (including three new + one extended regression test); full suite run for regression; `ruff check` clean; best-effort `mypy --strict` on the touched lines clean; `invariant-guard` 35/35
- [x] No new external network dependencies or online-LLM assumptions introduced
- [ ] Two-phase audit / quota / write-guard checklist — not applicable, no fsconnect/harness-write code touched
- [x] Relevant docs updated — the one stale in-code comment (`gate_auth.py`'s claim that `urlparse()` "itself never raises") is corrected as part of this PR; no external doc described this behavior
- [x] Commit message follows the title-prefix convention (`fix:`), includes an explicit before/after invariant statement, and states plainly that this is High-tier per `CLAUDE.md` §7

---

## Further comments

Explicit before/after invariant matrix, as this repo's template asks for on core-path changes:

| Check | Malformed Origin, BEFORE | Malformed Origin, AFTER |
|---|---|---|
| `gate.py` `_looks_cross_site` (→ `/index/build`) | Unhandled `ValueError` → framework 500 | `403 CROSS_SITE_BLOCKED` (same as any other cross-site Origin) |
| `gate_auth.py` `_enforce_same_origin` (→ 15 `/auth/*` routes) | Unhandled `ValueError` → framework 500, even when `auth.enabled` is `false` | `403 CROSS_ORIGIN_BLOCKED` (same as any other cross-origin Origin) |
| `harness/server.py` `_enforce_same_origin` (→ every guarded harness route) | Unhandled `ValueError` → framework 500 | `403 CROSS_ORIGIN_BLOCKED` (same as any other cross-origin Origin) |

ELI5: three copies of the same security check all trusted a header from the browser to be well-formed enough to parse without blowing up. One malformed value (a stray, unbalanced `[` character) proved that assumption wrong and crashed the request instead of just rejecting it the normal way. This teaches all three checks to treat "I can't even parse this" the same as "this doesn't match" — deny either way, just without the crash.

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vkd3GsHP3nKKBgusXS1Cei)_

### DIFF
--- a/gate.py
+++ b/gate.py
@@ -1261,7 +1261,15 @@ def _looks_cross_site(request: Request) -> bool:
         return True
     origin = request.headers.get("origin")
     if origin is not None:
-        hostname = urlparse(origin).hostname
+        try:
+            hostname = urlparse(origin).hostname
+        except ValueError:
+            # A structurally malformed Origin (e.g. an unbalanced IPv6
+            # bracket: "http://[evil") makes urlparse() itself raise, not
+            # merely a lazy .hostname/.port access -- attacker-controlled on
+            # an unauthenticated route, so this must fail closed as
+            # cross-site rather than let the exception escape as a 500.
+            return True
         return hostname is None or not _is_loopback_host(hostname)
     return False
 
--- a/gate_auth.py
+++ b/gate_auth.py
@@ -166,19 +166,38 @@ def register_auth_routes(
         origin = request.headers.get("origin")
         if origin is None:
             return
-        parsed = urlparse(origin)
         try:
-            # .port is a lazy property: urlparse() itself never raises, but
-            # reading .port does, for a non-numeric or out-of-range (>65535)
+            parsed = urlparse(origin)
+        except ValueError:
+            # A structurally malformed Origin (e.g. an unbalanced IPv6
+            # bracket: "http://[evil") makes urlparse() itself raise, not
+            # merely a lazy .hostname/.port access. It can never legitimately
+            # be this request's own origin, so treat it as an ordinary
+            # cross-origin mismatch -- attacker-controlled on an
+            # unauthenticated route, so this must fail closed rather than let
+            # the exception escape as a 500.
+            raise HTTPException(
+                status_code=_HTTP_FORBIDDEN,
+                detail={
+                    _CODE_KEY: "CROSS_ORIGIN_BLOCKED",
+                    _MESSAGE_KEY: "Cross-origin request rejected",
+                    _DETAILS_KEY: {},
+                },
+            ) from None
+        try:
+            # .port is a lazy property: a well-formed urlparse() result can
+            # still raise here, for a non-numeric or out-of-range (>65535)
             # port string -- e.g. Origin: http://localhost:notaport or
-            # http://localhost:99999. Both are attacker-controlled on an
-            # unauthenticated route, so an uncaught ValueError here would
-            # turn a malformed cross-origin request into a 500 instead of the
-            # 403 this check exists to return. A malformed port can never
-            # equal request.url.port (an int or None, never unparseable), so
-            # treating it as None (rather than re-raising) still fails the
-            # comparison below and rejects the request -- it just does so as
-            # CROSS_ORIGIN_BLOCKED, not a crash.
+            # http://localhost:99999. (A structurally malformed Origin raises
+            # earlier, at the urlparse() call itself, guarded above.) Both
+            # are attacker-controlled on an unauthenticated route, so an
+            # uncaught ValueError here would turn a malformed cross-origin
+            # request into a 500 instead of the 403 this check exists to
+            # return. A malformed port can never equal request.url.port (an
+            # int or None, never unparseable), so treating it as None
+            # (rather than re-raising) still fails the comparison below and
+            # rejects the request -- it just does so as CROSS_ORIGIN_BLOCKED,
+            # not a crash.
             origin_port = parsed.port
         except ValueError:
             origin_port = None
--- a/harness/server.py
+++ b/harness/server.py
@@ -586,15 +586,25 @@ def create_app(
                 },
             )
         origin = request.headers.get("origin")
-        if origin is not None and urlparse(origin).hostname not in _LOOPBACK_HOSTS:
-            raise HTTPException(
-                status_code=_HTTP_FORBIDDEN,
-                detail={
-                    _CODE_KEY: "CROSS_ORIGIN_BLOCKED",
-                    _MESSAGE_KEY: "Cross-origin request rejected",
-                    _DETAILS_KEY: {"origin_host": urlparse(origin).hostname},
-                },
-            )
+        if origin is not None:
+            try:
+                # A structurally malformed Origin (e.g. an unbalanced IPv6
+                # bracket: "http://[evil") makes urlparse() itself raise --
+                # attacker-controlled on an unauthenticated route, so this
+                # must fail closed (treated as a non-loopback host, below)
+                # rather than let the exception escape as a 500.
+                origin_hostname = urlparse(origin).hostname
+            except ValueError:
+                origin_hostname = None
+            if origin_hostname not in _LOOPBACK_HOSTS:
+                raise HTTPException(
+                    status_code=_HTTP_FORBIDDEN,
+                    detail={
+                        _CODE_KEY: "CROSS_ORIGIN_BLOCKED",
+                        _MESSAGE_KEY: "Cross-origin request rejected",
+                        _DETAILS_KEY: {"origin_host": origin_hostname},
+                    },
+                )
 
     # Per-instance, like rate_limiter above: a module-level singleton would leak
     # across separately-configured create_app() calls in tests, and a fixed
--- a/tests/test_gate_auth.py
+++ b/tests/test_gate_auth.py
@@ -318,18 +318,19 @@ class TestSameOrigin:
     @pytest.mark.parametrize(
         "malformed_origin",
         [
-            "http://localhost:notaport",  # non-numeric port
-            "http://localhost:99999",  # out of the 0-65535 range
+            "http://localhost:notaport",  # non-numeric port -- urlparse() succeeds, .port raises
+            "http://localhost:99999",  # out of the 0-65535 range -- same, lazy .port raise
+            "http://[evil",  # unbalanced IPv6 bracket -- urlparse() itself raises
         ],
     )
     def test_malformed_origin_port_is_rejected_not_a_500(self, manager, user, malformed_origin):
-        """urlparse().port is a lazy property that raises ValueError for a
-        non-numeric or out-of-range port string -- urlparse() itself never
-        raises, so nothing catches this until something reads .port. Both
-        values are attacker-controlled on an unauthenticated route; an
-        uncaught ValueError here would turn a malformed cross-origin request
-        into an unhandled 500 instead of the 403 this check exists to
-        return."""
+        """Two distinct ValueError sources, both attacker-controlled on an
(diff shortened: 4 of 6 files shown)


cgfixit (commented):
## Review (cyclaw-advisor) — #1103 Origin parse fail-closed

**Verdict: wise and safe to merge after you mark it ready.** Draft + High-tier label is correct. I am not merging it from this review.

### What I checked live
- Diff is three `urlparse(Origin)` sites + three tests. No graph edge, no sanitizer pattern, no auth *decision* change.
- Fail-closed direction is correct: unparseable Origin cannot be this request's origin.
- `gate_auth.py` now guards the `urlparse()` call itself, then keeps the existing lazy `.port` guard. That matches CPython: unbalanced `[` raises in `urlparse()`, bad/out-of-range port raises on `.port`.
- `harness/server.py` treats parse failure as `origin_hostname = None`, then the existing `not in _LOOPBACK_HOSTS` path returns 403. Good.
- `gate.py` `_looks_cross_site` returns True on ValueError → 403 CROSS_SITE_BLOCKED on `/index/build`. Test uses a distinct 127.0.0.2 client so it does not starve the class rate limiter. Good.
- I6: harness keeps its own copy of the check. No new core↔OOB import.
- Live CI on `763cbfd87b3f6cdf6bc5bcd2583b01ac52926887`: ubuntu/windows/macos, invariant-guard, CodeQL, Semgrep, Trivy, DevSkim all completed success (CodeQL check-run neutral is the usual GHAS annotation, not a fail).

### Residual, not blockers
- This does not invent a new allow path. Confirm no caller depended on the 500 (none found in the tests that were added).
- `urlparse` can also raise on some other garbage; all of that correctly becomes deny.
- Do not squash-merge this with the stacked kimi console PRs. Land this alone.

### Invariants
I1 RAG-first, I2 topology, I3 triple-gate, I4 audit convergence, I5 soul, I6 isolation: **no shape change**. CSRF deny path is stricter on undefined input.

Mark ready when you want it on main. I will not auto-merge High-tier auth-adjacent code.

cgfixit (commented):
## Pre-explicit-review findings (cyclaw-advisor + github-expert)

This is **not** an approval. High-tier auth-adjacent change. Operator review still required before merge.

### Verdict
**No required code change before your review.** Fail-closed direction is correct. Do not fold extra hardening into this PR.

Live head `763cbfd87b3f6cdf6bc5bcd2583b01ac52926887`. CI on that SHA: ubuntu/windows/macos, invariant-guard, CodeQL Analyze (python+actions), Semgrep, Trivy, DevSkim, bandit, osv-scanner, real-repo-run-smoke, docker-build — completed success. CodeQL check-run `neutral` is the usual GHAS annotation, not a fail. `mergeable_state: clean`.

Repo-wide `urlparse(origin)` search on live main + this branch: **exactly these three sites**. Nothing else to wrap.

### What the diff actually does
Wraps attacker-controlled `Origin` at `urlparse()` itself (CPython 3.12 `ValueError: Invalid IPv6 URL` on `http://[evil`, `http://[::1`, `http://a]b`). Treats parse failure as deny:
- `gate.py` `_looks_cross_site` → `True` → 403 `CROSS_SITE_BLOCKED`
- `gate_auth.py` `_enforce_same_origin` → immediate 403 `CROSS_ORIGIN_BLOCKED` (`from None` is correct)
- `harness/server.py` → `origin_hostname = None` → existing `not in _LOOPBACK_HOSTS` → 403 `CROSS_ORIGIN_BLOCKED`

`gate_auth.py` keeps the separate lazy `.port` guard (`localhost:notaport` / `:99999`). That is a different raise site; both need to stay. Well-formed `http://[::1]` still parses (`hostname='::1'`). Allow paths for well-formed Origins are unchanged.

### Reachability the PR body understates
`_looks_cross_site` is not only `/index/build`. Same helper also backs:
1. `POST /query` CSRF when `auth.enabled` (`_reject_cross_site_query`, #1055) — cookie-bearing hybrid queries that can bill Grok/Claude.
2. `security.api_key_optional` bypass (`_api_key_bypass_allowed`) — the CSRF replacement for a missing Bearer key on `/soul/*` `/ops/*` `/memory/*`.

So a 500 here is not just noisy logging. On the shipped default it is an unauthenticated crash on 15 `/auth/*` routes (deps run before `_require_enabled` 503). When auth or `api_key_optional` is on, it is a crash on the CSRF check that is supposed to stop a loopback-adjacent page from riding a session or skipping the API key.

### Invariants
I1–I5 untouched. I6: harness keeps its own copy — **do not extract a shared helper into `harness/`**. A `gate.py`/`gate_auth.py` private helper would be style only; not worth expanding this High-tier diff.

### Do not change in this PR
- `_looks_cross_site` is hostname-only. `Origin: http://localhost:notaport` and `:99999` still parse, hostname is `localhost`, check returns same-site. `gate_auth.py` rejects those via `.port`. Pre-existing gap vs the auth checker, not introduced here. Separate follow-up if you want port-fail-closed on `/query`.
- No `/query` or `api_key_optional` malformed-Origin test. `/index/build` covers the same function. Optional later.
- Do not merge with #1111–#1115 (console / invoke / installer). Land this alone.
- Do not treat `Referer` or invent a fourth checker.

### What you should look at by hand
1. Fail-closed on unparseable Origin (yes).
2. No new allow path (none found).
3. `/auth/*` still 403-then-503 ordering when auth is off — 403 on bad Origin is strictly better than the old 500; a well-formed Origin with auth off is still 503.
4. Harness copy still does not import gate.

Ready for your explicit review. I will not merge it.

PR #1125, opened 2026-08-27
[docs] - add operator-facing per-user auth section 

cgfixit:
## Proposed changes

`README.md` documented the `auth.enabled` per-user layer in exactly two places: one bullet in "What It Does" and one row in the Security Model table. Neither tells an operator how to actually turn it on. Anyone enabling it had to read `docs/AUTHENTICATION_DESIGN.md` plus `gate_auth.py`/`utils/authn*.py` to discover the bootstrap account name, the role matrix, the session limits, or why `/auth/*` answers 503 instead of 404.

This adds a `## Per-User Authentication (v1.9)` section (plus its TOC entry) covering:

- **The two credential systems side by side** — `CYCLAW_API_KEY` (Bearer, always on, fail-closed) vs. per-user auth (`auth.enabled`, ships `false`). Confusing the two is the common setup mistake.
- **Enablement walkthrough** — `auth.enabled: true`, the SQLite store at `auth.db_path`, `CYCLAW_AUTH_DB_URL` (and why it is deliberately not `CYCLAW_DB_URL`), the `admin` bootstrap account, `POST /auth/bootstrap-password`'s loopback+same-origin/403/409/503 contract, and the `cyclaw-user` subcommand set.
- **Role matrix** — `admin`/`operator`/`audit` capability table, including `audit` being denied on `/query` (403 `AUTH_ROLE_DENIED`), operator-cannot-touch-admin, and the last-enabled-admin protection.
- **Sessions, tokens, lockout** — `cyclaw_session` cookie + `X-CyClaw-CSRF`, why `/query` is CSRF-exempt, device tokens shown once and stored hashed, the two independent session limits, and the 5-failure/2 s-doubling/900 s-ceiling backoff.
- **The loopback caveat** — enabling auth does not by itself permit a non-loopback bind, and the auth+TLS route stays refused while `security.api_key_optional` is `true`.

Every number is read from `config.yaml` (`idle_timeout_sec: 43200`, `absolute_timeout_sec: 604800`, `db_path`) and `utils/authn.py` (`_LOCKOUT_THRESHOLD = 5`, `_LOCKOUT_BASE_SEC = 2.0`, `_LOCKOUT_CEILING_SEC = 900.0`, `ROLES`) / `utils/authn_manager.py` (`BOOTSTRAP_USERNAME = "admin"`) / `utils/gen_cert.py` (`--days` default `825`) rather than restated from memory.

Two smaller items ride along:

- **`chmod 600 .env`** is now documented in the "All platforms — `.env` file" section. That file is hand-created per the README and lands `0644` under a default umask, which `macos/invoke-cyclaw.sh` has refused to source since 2dc0264 (`600` or `400` only). Without this line the documented setup path silently produces a gateway with no `CYCLAW_API_KEY`.
- **`remaining_work.md`'s top note** removed. It was an informal scratch aside about a Telegram config session, already superseded by the `telegram.enabled: false` row in the shipped-posture table directly below it.

**Invariant / Governance Impact:** none. Documentation only — no code, config, or topology touched. The six invariants and I6 module isolation are untouched by construction.

---

## Types of changes

**Optional free-text scope note:** Docs + audits

---

## Benefits / why

- An operator can now enable per-user auth from the README alone, instead of reading three source modules to find the bootstrap username and the role matrix.
- The `chmod 600 .env` line closes a real documented-path failure: follow the README literally today and `invoke-cyclaw.sh` refuses the dotenv, then warns "CYCLAW_API_KEY not set" without ever naming the file mode as the cause.
- Sourcing every number from `config.yaml`/`utils/authn.py` keeps the section inside the repo's "config.yaml owns the numbers — cite, don't copy-and-drift" convention, so `doc-sync` can keep catching drift.

---

## Risks to monitor

- **Doc drift** is the only real risk: the role matrix and the session/lockout numbers are now stated in a second place. If `auth.session` values, `_LOCKOUT_*`, or the role capability checks in `gate_auth.py` change, this section must move with them. `python3 .claude/skills/doc-sync/doc_sync.py` does not currently assert on these particular values, so drift here would be caught by review, not by CI.
- The `chmod 600` guidance describes `macos/invoke-cyclaw.sh`'s current `600|400` rule. If that rule is widened or dropped, the README line becomes over-strict (harmless, but stale).
- No runtime behavior changes, so there is no rollback beyond reverting the commit.

---

## Checklist

- [x] This change preserves all 6 security invariants and I6 module isolation (docs-only; no code path touched)
- [x] `GROK_API_KEY=dummy pytest tests/test_macos_scripts.py tests/test_terminal_contract.py tests/test_harness_console_contract.py -q` — 180 passed
- [x] `python3 .claude/skills/doc-sync/doc_sync.py` — 0 drift items

---

## Further comments

**ELI5:** CyClaw has two different "passwords". One is a single shared API key that guards the admin routes and is always on. The other is a real multi-user login system with accounts and roles — and it ships turned off. The README explained the first one in detail and barely mentioned the second, so anyone who flipped it on had to go read the source to find out that the first account is called `admin`, that the `audit` role can't ask questions, or that you can't lock yourself out because the last admin is protected. This writes all of that down in one place.

The one behavior-adjacent line is the `chmod 600 .env`. A recent security fix made the macOS launcher refuse to load a secrets file that other accounts on the machine can read — which is correct, but the README still tells people to create that file by hand, and a hand-made file is world-readable by default. So the documented instructions produced a gateway that silently had no API key. The launcher-side half of that (a clearer error, and a fallback bug in the same block) is a separate PR.

## Merge topology

- Independent — cut from `origin/main`, no shared files with the sibling code PRs from this session.
- All `README.md` edits from this session are consolidated here, so the sibling PRs touch no file this one touches.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YURdSrJzFRjk4FN6eLJpp3)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@
 - [What It Does](#what-it-does)
 - [Architecture](#architecture)
 - [API Key Setup (Soul Mutations)](#api-key-setup-soul-mutations)
+- [Per-User Authentication](#per-user-authentication-v19)
 - [Quick Start](#quick-start)
 - [Docker / GHCR](docs/DOCKER.md)
 - [Full Setup Guide](setup-guide.md)
@@ -320,6 +321,18 @@ GROK_API_KEY=your-xai-key-or-dummy-when-offline
 ANTHROPIC_API_KEY=your-anthropic-key
 ```
 
+Then tighten it — a hand-created file inherits the shell's umask (usually
+`0644`, i.e. world-readable), and `macos/invoke-cyclaw.sh` **refuses to source a
+dotenv that is not `600` or `400`** rather than load secrets from a file other
+local accounts can read:
+
+```bash
+chmod 600 .env
+```
+
+`macos/setup-cyclaw-keys.sh` and the harness console's `/api` panel both already
+write `~/.CyClaw/.env` at `600`; only a hand-made file needs this step.
+
 The Claude variable is **`ANTHROPIC_API_KEY`**, not `CLAUDE_API_KEY` —
 `llm/client.py` and `agentic/config.py` both read the former, and nothing in
 the codebase reads the latter. Setting the wrong name is silent: Claude simply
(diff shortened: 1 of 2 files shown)


cgfixit:
@codex assess this PR and fix the verified documentation and security issues below.

Keep this PR documentation-only. Update the prose to match current runtime behavior; do not weaken authentication, CSRF, origin, API-key, or bind controls to make the prose true.

1. Remove the inline curl -d '{"password":"<a long passphrase>"}' bootstrap example. Copying it unchanged installs a publicly documented password that passes the current 12-character policy. Replacing it exposes the real admin password in shell history and curl process arguments. Prefer the first-boot UI or, after the pending admin account has been created on loopback, the no-echo command python -m utils.authn_cli passwd admin without --password. If an HTTP example remains, accept the request body securely through protected stdin without placing the password in argv, history, logs, or environment output.
2. Correct the credential and capability descriptions. Cookie-session writes require X-CyClaw-CSRF and then follow the role matrix. Login and bootstrap cannot require a CSRF token yet and instead use browser-origin checks. Bearer-authenticated /auth/* writes deliberately skip CSRF, but _require_write_actor permits them only for an admin-owned device token. Operator and audit device tokens cannot perform mutating /auth/* calls, including self-password changes. Make the role matrix credential-mode aware and state that an operator session can create only operator or audit accounts, never an admin.
3. Replace the statement that /query “mutates no auth state.” Session validation updates last_seen_ts, device-token validation updates last_used_ts, and the query is audited. State only that /query does not change credentials, roles, or account configuration. Also describe /auth/audit/summary as per-user-auth-gated rather than session-only because qualifying device tokens can access it.
4. Describe bootstrap accurately as loopback-only and browser-origin-checked. Headerless loopback clients such as curl are intentionally accepted, so it is not universally “same-origin only.” Note that the pending bootstrap is trust-on-first-use among local processes and should be completed immediately.
5. Qualify the “Serving it beyond loopback” section. With auth enabled, _reject_cross_site_query currently delegates to _looks_cross_site, whose Origin fallback rejects every non-loopback hostname. A browser request served from https://10.0.0.111:8787 with that matching Origin is therefore rejected. Do not present the LAN browser path as fully operational in this documentation PR. State the present limitation and route the exact scheme/host/port correction and its adversarial tests to a separate isolated code PR. Do not loosen _looks_cross_site globally because it also protects the api_key_optional bypass.
6. Make commands executable after the README’s documented base installation. Use python -m utils.authn_cli ... and python -m utils.gen_cert ..., or explicitly require the project self-install that provides cyclaw-user and cyclaw-gen-cert. Mention that PostgreSQL additionally requires the postgres extra and its TLS/least-privilege setup.
7. Scope chmod 600 .env and the 0644 umask explanation to macOS/Linux; they currently appear under “All platforms.” Preserve the Windows environment-variable and NTFS ACL guidance.
8. State explicitly that named device tokens do not inherit the 12-hour idle or 7-day session limits. They remain valid until revoked.

Add focused documentation-contract coverage preventing inline bootstrap passwords and checking the security-sensitive auth section against the implemented credential paths and configuration constants. Re-run doc-sync, auth/query tests, invariant guard, and the existing CI suite. Update the PR body if its claims or scope change.

PR #1173, opened 2026-08-28
[fix] - gate request-path: CEL monitor source key, off-loop monitor call, 429 audit throttle, sk- redaction parity

cgfixit:
## Branch naming (required for agent-opened PRs)

Head branch: `kimi/scan-gate-request-path` — driver-matched `kimi/` prefix per the table (created by Kimi Code).

## Title

`[fix] - gate request-path: CEL monitor source key, off-loop monitor call, 429 audit throttle, sk- redaction parity`

---

## Proposed changes

Four small, independent correctness/robustness fixes to the `gate.py` request path (plus a one-line `config.yaml` alignment), from an approved optimization scan of `main` @ `ea8f8035`:

1. **C1 — CEL monitor read a nonexistent graph state key.** `query_endpoint` built `source_hashes` from `result.get("sources", [])`, but `GraphState` exposes `answer_sources` (`graph.py:153`, written by every answering node and read by `audit_logger_node`). `"sources"` is never populated, so `source_hashes` was permanently empty and CEL rules over the documented field (`config.yaml` `numbat.cel` comment lists `source_hashes`) could never match. Now reads `answer_sources`. New integration test (`TestCelMonitorRequestPath::test_monitor_emits_on_real_source_hash`) drives a real `/query` through the real `monitor_request` with a real celpy rule keyed on the exact expected `hash_query("test.md:0")` and asserts the Numbat event is emitted.
2. **C2 — monitor call blocked the event loop and hashed unconditionally.** `monitor_request(...)` (CEL compile/eval + potential Numbat file write) and `hash_query(req.query)` ran synchronously inside the async handler. The monitor call now runs via `asyncio.to_thread` (same pattern as `_audit` / `_check_rate_limit_async`), and the whole block — including all hashing — is skipped unless `numbat.cel.enabled` is literally `true` (mirroring `utils.numbat_cel._is_literal_true`). Behavior when disabled is unchanged (the module already no-ops); this just stops paying the hashing cost per query. Covered by `test_monitor_skipped_when_cel_disabled`.
[...]

PR #1177, opened 2026-08-28
[[[ [tests] - assert needs_user_confirm both directions, share gateway client fixture, pin raises match=

cgfixit:
## Branch naming (required for agent-opened PRs)

Head branch `kimi/scan-test-hygiene` uses the driver-matched `kimi/` prefix (Kimi Code session), per the allowlist in `utils/agent_identity.py`.

## Title

`[tests] - assert needs_user_confirm both directions, share gateway client fixture, pin raises match=`

---

## Proposed changes

Test-suite hygiene only — **no production code changes** (`tests/` + one JSONL fixture). Four findings from the approved optimization scan:

1. **`tests/test_graph_outcomes.py` + `tests/fixtures/graph_outcomes.jsonl` — assert `needs_user_confirm` in BOTH directions.** The runner previously asserted the flag only when the fixture expected `True`. Verified against `graph.py`: `route_by_score_node` sets `needs_user_confirm=True` on a low top score (`graph.py:419-422`) and `user_gate_node` returns `{}` once the user has answered (`graph.py:603-604`), so an **answered** low-score row still finishes with the flag `True`. An empirical probe of all 20 graph rows confirmed exactly 12 answered low-score rows observed `True` while the fixture declared `false`. The fixture rows were corrected to `true` (the real observable state — the assertion was NOT weakened) and the runner now asserts both directions with an explanatory message. The 3 high-score rows genuinely finish `False` and now have real coverage of the False direction. Gate/retrieval-layer rows never run the graph, so their flag field stays unasserted (documented in the runner comment).
2. **Same file — dead fixture-vs-constant assertions moved to one-time validation.** `_run_gate_row` asserted `expected_error_code == 400` / `expected_route == "gate_block"` and `_run_retrieval_row` asserted `expected_route == "keyword_hit"` — comparing fixture data to hardcoded constants, dead as behavioral coverage. These now live in a new `test_fixture_layer_contract` validation test (alongside `test_battery_covers_required_categories`); the per-row runners keep only behavioral assertions.
3. **Shared mocked-gateway `client` fixture extracted to `tests/conftest.py`.** `test_gate.py:26-102` and `test_edge_cases.py:22-80` carried copy-pasted `client` fixtures that had already drifted (different default answers, only one saved/restored gate globals, only one nulled `gate.claude`). One implementation now lives in conftest as `_mocked_gateway(tmp_path, *, peer=...)`; `test_gate.py` uses the default conftest `client` fixture unchanged, `test_edge_cases.py` keeps a thin wrapper. The dead `patch("gate.open", ...)` / `patch("gate.yaml.safe_load", ...)` no-ops in `test_edge_cases.py` (documented as dead in `test_gate.py`'s own comment) are deleted. **Important subtlety found during verification:** gate's rate limiter is a process-global 60 req/60 s bucket keyed per client IP, so each consuming file must use its OWN loopback peer — an intermediate version of this PR put both files on `(127.0.0.1, 51234)` and the full suite failed with 429-instead-of-409/200 in `test_gate_index_build.py` (whose own docstring warns about exactly this). `test_edge_cases.py` therefore uses `127.0.0.3`. The third "near-duplicate" named in the scan, `tests/test_gate_ops.py`, was **dropped**: its `_build_app()` builds a separate minimal FastAPI app with injected fakes and never imports `gate` — not a duplicate of this fixture.
4. **`tests/conftest.py` — `_disarm_agentic_write_execution` no longer swallows broken imports.** `except ImportError: return` on `import agentic.writer` meant a broken transitive import INSIDE `agentic/writer.py` would silently leave `EXECUTION_ENABLED` armed for the whole session. Now only a `ModuleNotFoundError` naming `agentic`/`agentic.*` (genuinely absent package) returns; anything else re-raises.
5. **Broad `pytest.raises(ValueError)` pinned with `match=`.** `tests/test_memory_policy.py` (`reason must not be empty`, `fact content exceeds max_content_chars`, `fact content must not be empty`, `tag exceeds max length`, `too many tags`), `tests/test_memory_store.py` (`reason must not be empty` x2), `tests/test_telegram_state.py` (`chat_id` — params span both chat-id validators; `provider must be grok or claude`), `tests/test_harness_atomic_write.py` (`Circular reference`). Messages verified against `memory/policy.py`, `memory/store.py`, `telegram/state.py`, and stdlib `json`.

- None of the six invariants or I6 module isolation is affected — no production module is touched; graph topology, gate policy, audit convergence, soul governance, and RAG-first entry are unchanged.
- Evidence: the diff is confined to `tests/`; the full suite (4837 tests) passes with the corrected fixture that now matches the graph's real observable routing state rather than an idealized one. If anything, coverage of the I3 user-gate flag is stronger: the False direction is now asserted on the high-score path, and the fixture no longer claims answered low-score queries clear the pause flag.

---

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — test-fixture rows contradicted real graph behavior; fixture corrected
- [ ] Breaking change
- [ ] Documentation Update

**Scope note:** test infrastructure only (`tests/`).

---

## Benefits / why

- The outcome battery now actually verifies the `needs_user_confirm` contract in both directions instead of silently skipping 25 of 30 rows on that axis.
- The fixture no longer mis-documents graph behavior: answered low-score queries keep `needs_user_confirm=True` in final state, and the runner comment explains why (user_gate returns `{}` on a decided confirm).
- One gateway-mock implementation instead of two drifting copies; the dead `gate.open`/`gate.yaml.safe_load` patches are gone; gate-global save/restore now applies to both consumers.
- A broken `agentic.writer` import now fails loudly at test time instead of leaving the write-execution flag armed for the session.
- `match=` pins keep the ValueError tests honest about WHICH validation fired, so a future refactor that raises the wrong error message breaks the right test.

---

## Risks to monitor

- The corrected fixture rows encode current `graph.py` behavior (flag never cleared after a decided confirm). If a future change deliberately clears the flag in `user_gate_node` or downstream nodes, those 12 rows fail — that is the intended signal, not a flake.
- Rate-limit budget sharing: the shared fixture's docstring and `test_edge_cases.py`'s wrapper both document the per-IP 60/min process-global limiter; any FUTURE file reusing `(127.0.0.1, 51234)` can starve `test_gate.py` / `test_gate_index_build.py` in full-suite runs. Detection: 429s where 200/409 were asserted.
- `_disarm_agentic_write_execution` now re-raises on broken transitive imports inside `agentic/writer.py` — an environment missing an agentic dependency will surface as a collection/setup error rather than a silent skip. That is deliberate (fail loud), but it changes the failure surface on partial installs.
- No effect on production runtime, offline posture, or audit behavior.

---

## Checklist

- [x] I have read the latest `docs/CyClaw Architecture Guide` (and any relevant Phase docs) and `SECURITY.md`
- [x] This change preserves all 6 security invariants and I6 module isolation (tests-only diff; evidence above)
- [x] Full sandbox validation has been run (`GROK_API_KEY=dummy python -m pytest tests/ -q --tb=short`) and passes with no regressions — 4837 tests, 0 failures, 0 errors, 350 skipped (pre-existing platform/service skips), exit code 0. `bash .claude/skills/CyClaw-Sandbox/verify.sh` not run: no core RAG/agentic path changed (tests only).
- [ ] For any agentic/fsconnect/harness change: two-phase audit, quota enforcement, governed delete/trash, and write guards have been verified — N/A (test-only)
- [x] Relevant architecture docs, threat model notes, or harness phase documentation have been updated if core behavior or topology changed — N/A, no core change
- [x] Commit messages follow the title prefix convention above (`[tests] - ...`)
- [ ] For large or complex changes: before/after invariant matrix — N/A (small, tests-only)

---

## Further comments

What changed, plainly: the test suite had four hygiene problems. (1) The big data-driven battery checked a routing flag only when it expected "pause", so rows that claimed "no pause needed" were never actually checked — and it turned out 12 of those rows claimed the wrong thing, because the real graph never clears the flag once a low retrieval score sets it. The fixture was fixed to match reality (verified by running the real mocked graph row-by-row), and the check now runs in both directions. (2) A few per-row assertions just compared the fixture to itself; they moved to a single fixture-validation test. (3) Two test files maintained separate copies of the same mocked-gateway setup that had already drifted apart; there is one shared implementation now, with each file on its own loopback IP so the rate limiter's per-IP budget can't starve a neighboring test file. (4) A safety fixture that disarms the agentic write path for tests used to ignore ANY import error, including a broken `agentic/writer.py`; now only a genuinely absent package is tolerated. Plus `pytest.raises(ValueError)` calls now pin the expected message.

Verification performed (from worktree root, system Python 3.12):
- `GROK_API_KEY=dummy python -m pytest tests/ -q --tb=short` → **exit 0**; junit: 4837 tests, 0 failures, 0 errors, 350 skipped.
- `ruff check --select E,F,I,B,C4,UP,S` on all touched test files → **All checks passed**.
- Empirical probe of all 20 graph fixture rows before editing: confirmed the scan's B1 claim exactly (12 answered low-score rows observe `needs_user_confirm=True`; 3 high-score rows observe `False`).
- An intermediate version of the shared fixture regressed `test_gate_index_build.py` (429 via shared rate-limit bucket); root-caused to the per-IP limiter and fixed by giving `test_edge_cases.py` its own peer — final full-suite run is green.

Unverified / out of scope:
- `tests/ci_rag_smoke.py` untouched (deliberately not renamed, per repo convention) and not run separately here.
- One flaky `PermissionError [WinError 5]` in `test_harness_memory.py::test_notes_add_forget_clear` (Windows `os.replace` temp-file lock) appeared on one full-suite run and passed on isolated rerun and in the final green run — pre-existing Windows flake, unrelated to this diff (that file's only change is a `match=` pin in a different test).
- Merge order: independent; no overlap with open PRs #1167 (fsconnect) or #1168 (network connector). Base commit: `ea8f8035`.

---

## Update 2026-08-28 — review pass fix (tip `7738f74f`)

One follow-up commit (rebased onto `origin/main` @ `6aed6e58` first; force-with-lease push):

- `7738f74f` `[fix] - support indirect parametrization in conftest client fixture` — the extracted conftest `client` fixture now takes `(request, tmp_path)` with `peer = getattr(request, "param", ("127.0.0.1", 51234))` passed through to `_mocked_gateway`. This rebase-proofs the known `tests/test_gate.py` conflict with #1173 (`kimi/scan-gate-request-path`), whose new test classes use `@pytest.mark.parametrize("client", [...], indirect=True)` — whichever PR merges second now resolves by keeping the conftest fixture and deleting the old local one. Default peer is unchanged, so existing consumers are unaffected.

Verification on the rebased tip: `ruff check tests/conftest.py tests/test_gate.py` clean; `GROK_API_KEY=dummy python -m pytest tests/test_gate.py tests/test_edge_cases.py tests/test_graph_outcomes.py -q --tb=short` → **228 passed**; indirect parametrization proven with a throwaway parametrized test (deleted, never committed); pre-push CI emulation → all PASS.


cgfixit:
@codex review

cgfixit:
@codex review

Issue #1134, opened 2026-08-27
NeMo Guardrail implementation plan


cgfixit:
> Note: this is roughly a 5 phase process (or less)

You are hardening CyClaw’s NeMo, model-boundary, tool-call and agentic-coding security.

Repository: cgfixit/CyClaw
Latest GitHub origin main branch

Your goal is to build a provider-independent, single-generation safety architecture covering local cached Qwen, offline fallback, online escalation, harness chat, agentic model calls, tool calls/results and model-authored code execution.

Do not treat NeMo as the authorization boundary. Deterministic identity, capability, routing, path, network, schema, approval and sandbox policy is authoritative. NeMo may only deny, redact, quarantine or require approval; it must never grant a capability, select an online route, expand a tool registry or override a deterministic denial.

1. Working protocol

1. Fetch origin/main, record the new base SHA and compare it.
2. If main advanced, re-audit every affected file and update the implementation assumptions before editing.
3. Read AGENTS.md, README.md, SECURITY.md, docs/THREAT_MODEL.md, all files under guardrails/ and docs/NeMo/, plus:
    * utils/guardrail_bridge.py
    * utils/sanitizer.py
    * utils/external_pre_hook.py
    * graph.py
    * gate.py
    * llm/client.py
    * harness/server.py, harness/prompts.py, harness/ollama.py, harness/web_search.py
    * agentic/context.py
    * agentic/real_repo_loop.py
    * agentic/executor/runner.py
    * agentic/real_repo_run_store.py
    * agentic/deepagent_github/
    * agentic/harness_optimizer/
    * relevant tests and CI workflows.
4. Create an isolated worktree and dedicated branch. Preserve unrelated user changes.
5. Hash soul.md before and after. Do not modify it. Never send it to an external provider or guard model.
6. Before each material change, record: observed failure, security hypothesis, proposed invariant and verification method.
7. Do not merge. Use small sequential draft PRs. Rebase each phase on the merged result of the prior phase; do not maintain concurrent branches editing the same security boundary.
8. Stop after three materially distinct failed hypotheses for the same problem and report the evidence rather than weakening the invariant.

2. Required security invariants

* Exactly one primary answer generation occurs per turn.
* Every model invocation crosses one registered model boundary.
* Every retrieved/external datum is provenance-labelled and treated as untrusted.
* Every effectful tool crosses one registered pre-tool and post-tool boundary.
* Guardrails can reduce permissions but cannot create them.
[...]

PR #1171, opened 2026-08-28
[docs] - fix remaining_work relocation links, stale open items, and isort first-party memory

cgfixit:
## Branch naming (required for agent-opened PRs)

Head branch `kimi/scan-docs-drift` — `kimi/` prefix matches the Kimi Code driver, kebab-case feature name.

## Title

`[docs] - fix remaining_work relocation links, stale open items, and isort first-party memory`

---

## Proposed changes

HEAD commit `ea8f8035` moved `remaining_work.md` into `docs/plans/` and left four inbound relative links resolving to the wrong place, plus related doc drift. This PR is docs-only plus one `pyproject.toml` isort config line:

- **`docs/plans/remaining_work.md`** — fixed the two header links broken by the relocation: `../zWork/remaining_work_STALE.md` and `../ARCHIVE_AND_ROADMAP.md` (both previously resolved under `docs/plans/docs/...`). Also removed two "Still open" items that have already shipped (see below) and corrected the `docs/TESTCLIENT_HTTPX_DEPRECATION.md` prose reference to its real location, `docs/analysis/TESTCLIENT_HTTPX_DEPRECATION.md` (pre-existing drift, verified via glob — the file only exists under `docs/analysis/`).
- **`docs/zWork/remaining_work_STALE.md`** — fixed `../../remaining_work.md` → `../plans/remaining_work.md`. Also fixed the same-file `docs/ARCHIVE_AND_ROADMAP.md` link at :29 → `../ARCHIVE_AND_ROADMAP.md` (same broken-relative-link class; previously resolved under `docs/zWork/docs/...`).
- **`docs/ARCHIVE_AND_ROADMAP.md`** — fixed the header link `../remaining_work.md` → `plans/remaining_work.md` and updated the link text accordingly.
- **`README.md:31`** — "Remaining Work" TOC entry pointed at `docs/zWork/remaining_work_STALE.md`, whose own header says "Not the live list"; repointed to `docs/plans/remaining_work.md`.
- **`.claude/skills/verify-deps/SKILL.md:41`** — the surface table claimed `requirements.txt`'s "Header declares itself DEPRECATED"; the header was reworded to "legacy compatibility surface". Skill line updated to match.
[...]

PR #1178, opened 2026-08-28
[docs] - consolidate CyClaw-Sandbox into one accurate skill, heal its verification scripts

cgfixit:
## Proposed changes

`.claude/skills/CyClaw-Sandbox/` shipped two competing skill documents since PR #1090: `SKILL.md` (818 lines — the one the skill runtime actually loads; its `description:` is what shows up in the available-skills listing) and `NEW_SKILL.md` (583 lines — **inert**: a stray instruction line sits above its YAML `---` delimiter, which breaks frontmatter parsing, and it has zero inbound references anywhere in the repo). `SKILL.md` made roughly a dozen confirmed-wrong claims about the current codebase — auth said "not landed" when it's fully wired, "10 telemetry env vars" against a 21-entry canonical map, "four `/api/agent/*` routes" against seven, "13 harness slash commands" against ~19, a "29 smoke checks" count that was never real, a stale graph-node count, and route ownership predating the `gate_ops`/`gate_auth`/`gate_memory` split. Its main automated ladder, `run_full_verification.py`, **failed on a clean tree**: a removed `retrieval.stemmer.PorterStemmer` import, a `gate.py`-only grep for routes that moved to `gate_ops.py`, and a `static/terminal.html`-only check for console logic that moved to the sibling `terminal.js` (CSP forces `script-src 'self'`).

Meanwhile the inert `NEW_SKILL.md` was factually *more* current on several of exactly these points (12 graph nodes, Auth Stages 3+4 wired) and covered real surfaces `SKILL.md` never mentioned (spend ledger, Numbat, sequence detection, Telegram/OpenTweet/unslop, MCP manifest pin, `/api/keys`) — accurate content sitting in a file nothing loads.

This PR, in three commits (code before docs):
1. **Heal the checkers** against current main — fix every stale check above in `run_full_verification.py`, extend its coverage per the user's explicit ask (all four terminal slash commands, `/api/keys`, `/api/auth/setup-status`, the full agent-route auth-gate set, the harness slash palette derived from `COMMANDS` instead of hardcoded), fix `test_terminal_consoles.py`'s four 400→422 assertions (closed `Literal` schema fields) plus two more bugs discovered live (a header-casing test bug — the headers are actually present — and a DROP-rejection check that didn't account for `sqlconnect.enabled: false` short-circuiting first), extend both `*_runtime_check.py` route subsets, and add the new ToolBroker (issue #1134) presence check plus two additive `harness_emulation.py` steps.
2. **Consolidate into one `SKILL.md`** — merges the accurate content from both documents, drops everything confirmed wrong, and replaces fragile hardcoded counts with either a real automated pin or an explicit "(derive)" instruction to read the number from the tree at verification time. Deletes `NEW_SKILL.md`.
3. **Sync the doc ripple** — `run.md`, `CyClaw-Sandbox.md`, `status.md`, `.claude/README.md`, `harness/README.md`, one new `.claude/settings.json` allowlist entry, and a dated `docs/audits/` report with the full drift table and verification evidence.

Resynced against `origin/main` (`a4ca399`, 28 commits ahead of this branch's prior base — see the plan/commit history) before implementation; confirmed the entire target surface (all `CyClaw-Sandbox` files, both console HTML/JS files, all four `gate*.py` files) was untouched by that drift, and folded in the two genuinely new surfaces that had landed (the ToolBroker gate, the `netconnect` connector) as additive coverage rather than a replan.

- **All 6 invariants + I6: untouched.** No graph node, edge, router, auth surface, sanitizer pattern, config value, or soul path changed. `gate.py` is not in this diff. `python3 .claude/skills/invariant-guard/check_invariants.py` — 47/47, unaffected.
- Scope is entirely the out-of-band verification skill (`.claude/skills/CyClaw-Sandbox/*`) and its own documentation ripple.

## Types of changes

- [x] Documentation Update

**Scope note:** Docs + audits (the skill's own scripts are verification tooling, not product code).

## Benefits / why

- A verification skill that reports false state is worse than having none. The live document claimed auth wasn't enforced, undercounted the telemetry contract by half, missed three of seven agent routes, and described a stale 13-command harness palette — an operator trusting a green run from this skill was trusting wrong facts.
- The automated ladder now actually runs clean: `run_full_verification.py` went from failing outright on several phases (confirmed via a scratch-clone run) to 222/225, with the one remaining gap a documented, reproduced-independent-of-this-PR environment artifact (see the dated report).
- Counts are drift-proofed going forward: every number in the new `SKILL.md` is either backed by a named automated pin or explicitly marked to derive from the live tree — the exact discipline that would have caught this drift originally.
- Directly answers the ask: both consoles' full slash-command palettes, tool/skill wiring, and REST surfaces are now verified, not just gestured at.

## Risks to monitor

- `run_full_verification.py` Phase 8's route check now concatenates `gate.py + gate_ops.py + gate_auth.py + gate_memory.py` source text rather than importing the app — a deliberate choice (matches this script's existing static-analysis style everywhere except the harness phase, whose own comment explains why harness alone gets a live `TestClient`); if a route's literal path string is ever refactored into an f-string or constant, this grep-based check stops seeing it, same limitation the original script always had.
- The Phase-10 membership checks (`{"goal","loop","hybrid_search"}` tool names, `{"ponytail","karpathy-guidelines","invariant-guard"}` skill names) are still spot-checks, not exhaustive — verified true today, will need re-verification whenever the harness registry changes shape.
- One pre-existing, unrelated finding surfaced during verification and is called out in the dated report rather than fixed here (out of scope): `docs/CyClaw Audit Bundle Export Script.py`, added by a commit already on `main` before this branch was cut, has a syntax error that will make a bare repo-wide `ruff check` report 2 errors regardless of this PR.

## Checklist

- [x] I have read the latest architecture docs and `SECURITY.md`
- [x] This change preserves all 6 security invariants and I6 module isolation (invariant-guard 47/47; no core file in this diff)
- [x] Full sandbox validation has been run (`GROK_API_KEY=dummy pytest tests/ -q --tb=short` — 216 focused contract tests green; see the dated report for the full live-verification transcript) and passes with no regressions
- [x] Relevant architecture docs updated (the skill itself, plus run.md/CyClaw-Sandbox.md/status.md/.claude/README.md/harness/README.md)
- [x] Before/after evidence included in "Further comments" below and in `docs/audits/2026-08-28-sandbox-skill-consolidation.md`

## Further comments

**ELI5:** This repo had two copies of its own "how do I test CyClaw" instruction sheet. The one actually in use was stale — it described an older version of the app (routes that moved, a smaller telemetry list, an auth system it thought was unfinished) and its main test script literally crashed on a fresh checkout. The other copy, which nothing was reading, had newer facts but a formatting mistake that made it invisible to the system. This PR merges them into one sheet that matches the app as it is today, fixes the crashing script, and — instead of hardcoding numbers that will just go stale again — tells the reader exactly where to look them up live next time.

**Before/after (see `docs/audits/2026-08-28-sandbox-skill-consolidation.md` for the full 13-row table + live evidence):**
- `run_full_verification.py`: failing on a clean tree (removed import, wrong grep target, stale HTML-only check) → 222/225 on a scratch clone, the 1 gap a documented environment artifact unrelated to this PR.
- `test_terminal_consoles.py` (live gate): 38/41 → 41/41.
- `harness_emulation.py` (live harness): full pass including 5 new agent-route auth-gate checks and 2 new steps.
- `doc-sync`: 0 drift. `invariant-guard`: 47/47.

No compensating controls needed — this is a verification-tooling accuracy fix, not a security-relevant behavior change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gmqfm8s7PEv4QvqLSyMMEA)_


cgfixit:
**Lint (Ruff + WPS) is failing, but not because of this PR.**

The check fails on `docs/CyClaw Audit Bundle Export Script.py:247-248` — a genuine syntax error (a ternary split across three lines without wrapping parens, so the parser sees an orphaned `if`/`else`):

```
invalid-syntax: Expected `:`, found newline
   --> docs/CyClaw Audit Bundle Export Script.py:247:52
invalid-syntax: Expected a statement
   --> docs/CyClaw Audit Bundle Export Script.py:248:17
```

This PR never touches that file. It's a pre-existing break on `main`: Lint has been red on every push to `main` since the merge of #1171 (`b962ad0`), including the current tip (`a4ca399`, this PR's base, and the even-newer `2e7374b`) — confirmed via the workflow run history, not assumed. I searched open PRs for an existing fix to port; none touches this file, so there's nothing to port yet.

Proposed patch (wraps the ternary in parens so it parses as the single continued expression it was clearly meant to be — semantics unchanged, `hashlib.sha256(...)` only runs when `hasattr` is false, same as before):

```diff
-            actual_hash = sha256_file.__wrapped__(zf.read(name)) 
-            if hasattr(sha256_file, "__wrapped__") 
-                else hashlib.sha256(zf.read(name)).hexdigest()
+            actual_hash = (
+                sha256_file.__wrapped__(zf.read(name))
+                if hasattr(sha256_file, "__wrapped__")
+                else hashlib.sha256(zf.read(name)).hexdigest()
+            )
```

Leaving it out of this PR's diff since it's unrelated to the CyClaw-Sandbox consolidation — happy to open a separate one-line fix PR if that's preferred over applying it directly.

---

PR #1196, opened 2026-08-28
[security] - harness: stop returning absolute artifact paths from the open /api/harness/runs

qwoifhqgbio:
Audit finding F-04. Independent of #1195 (harness `/static` mount + CSP nonce) — different concern, cut fresh from `main`, no shared lines.

## Proposed changes

`GET /api/harness/runs` returned `{"run_id": ..., "path": str(path)}` for every accepted optimizer artifact, where `path` is absolute and on a POSIX box embeds the operator's home directory and so their OS username. The route is **unauthenticated** — it carries no `dependencies=` at all.

Its sibling `GET /api/agent/runs/{run_id}` *is* guarded, and its docstring says why: *"the record names a branch, the changed file paths, and the clone's absolute location."* So the same class of value was gated on one route and public on the other.

**The fix drops the field instead of guarding the route.** That is deliberate, and it is the part of this PR worth reviewing:

- **Nothing consumed it.** The console renders the id alone — `sys(table(r.runs.map(x => [x.run_id]), ['run']))` in `static/harness.html`. Both sandbox verifiers assert only that the payload carries `runs` + `count`. No test asserted `path` except the ordering test updated here.
- **The route is open on purpose.** It sits in `tests/test_harness_auth.py`'s `OPEN` set, whose comment reads *"the console must be able to boot and tell the operator it needs a key."* Adding `dependencies=guarded` would break that contract, make the `/harness` console command require a key, and move the guarded count **29 → 30** — which would in turn make `CLAUDE.md`'s stated count stale.
- So guarding would change an auth contract and three documents in order to protect a value no caller reads. Removing the value is both smaller and stricter: it eliminates the data rather than gating it.

**Deliberately not adding a rate limit.** None of the nine `OPEN` routes carries one. Limiting only this route would introduce a new inconsistency rather than remove one, and with `path` gone the response is a list of variant-id stems.

**None of the 6 invariants or I6 is affected**, and **no route's authentication changes**. This touches only a response payload in the out-of-band harness layer.

- Guarded route set unchanged at **29**; `OPEN` set unchanged. `invariant-guard`: **47 passed, 0 failed**.
- No import added or removed, so I6 module isolation is untouched by construction.
- No graph edge, auth gate, sanitizer pattern, or soul path is involved.

## Types of changes

**Scope note:** Out-of-band harness layer. Core graph/gate/soul path untouched.

## Benefits / why

- An unauthenticated endpoint stops disclosing a filesystem path that identifies the operator's account, closing the asymmetry with the guarded sibling that protects exactly this.
- Data minimization over access control: there is now no sensitive value to leak if this route's guarding is ever reconsidered, and no auth contract had to move to get there.
- The open-console boot path (list runs before entering a key) keeps working exactly as documented.

## Risks to monitor

- **Payload shape change.** Any out-of-tree consumer reading `runs[].path` would break. I grepped the repo — console, tests, `harness/tools_view.py`, `harness/README.md`, and both `Cyclaw-Sandbox` verifiers — and nothing reads it. The harness API is loopback-only and single-operator, and `harness/README.md` documents routes without pinning payload shapes, so there is no published contract this violates. Flagging it because it is the only behavioral change in the diff.
- **The path is still on disk and still derivable** — `_RUNS_DIR` is a fixed repo-relative location. This removes gratuitous disclosure over an unauthenticated channel; it is not a secret-management control.
- **If someone later guards this route anyway**, the in-code comment explains why the field was removed rather than gated, so the two fixes are not applied redundantly.

## Checklist

- [x] This change preserves all 6 security invariants and I6 module isolation (`invariant-guard`: 47 passed, 0 failed)
- [x] For any harness change: no write guard, quota, or auth behavior altered — guarded set stays 29
- [x] Relevant docs checked — none pinned the payload shape, so none needed updating

## Further comments

**Verification** (Python 3.12.3, ruff 0.16.1):

| Check | Result |
|---|---|
| Full suite | **4827 passed**, 71 skipped, 1 failed |
| `invariant-guard` | 47 passed, 0 failed |
| `doc-sync` | 0 drift items |

The one failure is `test_fsconnect_quota.py::test_quota_recompute_fail_closed_on_unreadable_root`, **pre-existing and environmental** — the sandbox runs as uid 0, and root bypasses the directory permission bits the test relies on, so the expected `FsWriteRefused` never fires. Verified by reproducing it on clean `main`, and it passes in CI's non-root ubuntu leg.

**On the test change.** The ordering test's `assert all(set(r) == {"run_id", "path"} ...)` became `== {"run_id"}`, and a separately named `test_harness_runs_does_not_expose_absolute_paths` was added. The separate name is the point: the key-set assertion lives in a test *about mtime ordering*, where a future editor could reasonably relax it as incidental detail. The new test asserts no filesystem path appears anywhere in the serialized payload — not just inside the `runs` list — so a path reintroduced under a different key is caught too.

**ELI5.** A page on the local dev console listed your saved optimizer runs, and for each one it also handed back the full folder path on your machine — which usually contains your username. Anyone who could reach that page could read it, and no password was needed. Nothing on the page actually used that path; it just displayed the run's name. So the path is simply no longer sent. The alternative was to put a password on the page, but that page is intentionally password-free so the console can start up and tell you it needs a key.

---


cgfixit:
@codex review

PR #1197, opened 2026-08-28
[fix] - make config-guard docs, app.debug, and the spend ledger say what they mean

qwoifhqgbio:
## Proposed changes

Output of a `/CyClaw-Optimize` sweep of `main`. Three defects that share one shape: **a file states something an operator or auditor would act on, and the statement is not true.** Each is small, verified against the code, and independently revertable.

**1. `config-guard --strict` is documented as a merge gate but cannot pass.** `SKILL.md` says *"use it as a merge gate when you want the shipped defaults locked."* Verified on pristine `main`: it exits **2**, because C9 warns on the deliberately-armed posture (`app.mode: hybrid`, grok/claude enabled — `docs/THREAT_MODEL.md`'s eighth amendment). Wiring the documented command into CI breaks every build. The doc now states that plainly and names the default (non-strict) invocation as the real gate. The same table also omitted **C13**, which `check_config.py` implements — added.

**2. `config.yaml` ships `app.debug: true` with zero code consumers.** Repo-wide grep across `gate.py`, `graph.py`, `utils/`, `harness/`, `llm/`, `retrieval/`, `schemas/`: `cfg["app"]` is only ever read for `["mode"]`. It does not enable debug mode, loosen auth, or change logging (that's `logging.level`). Annotated rather than deleted, exactly the way `security.require_env` already is a few hundred lines down — so an auditor reading `debug: true` isn't left concluding the shipped server runs debug-enabled.

**3. The spend staleness alarm never fired where the money is spent.** `warn_if_priced_as_of_stale()` had exactly one non-test caller: `metrics.py`, an operator-invoked report. `record_external_usage()` — the path that prices live Grok/Claude calls — never consulted it, so a server billing against a >30-day-stale rate table emitted **no signal at all** until someone separately ran `metrics.py`. Now warns once per process from the recording path (warn-once, not per call: a per-call warning on the hot path is spam, and spam gets filtered exactly when it matters).

While there I corrected a related-but-different premise: `PRICED_AS_OF = "2026-08-19"` sits 8 days behind the `grok-4.3` row's `verified 2026-08-27` comment, which looks like drift. It isn't — it is the **oldest** verified date in the table, which is the *conservative* choice, because `rates_are_stale()` alarms off this one constant. Bumping it for a partial re-verification would mask the remaining rows going stale. That semantic is now documented on the constant and **pinned by a test**, so the tempting future "fix" fails loudly.

## Types of changes

- [x] Bugfix (staleness alarm unreachable from the recording path)
- [ ] New feature
- [ ] Breaking change
- [x] Documentation Update

Scope note: docs + audits, out-of-band spend ledger, config annotation. No core graph/gate/soul path.

**Invariant / Governance Impact:** none of I1–I6 affected. No graph node, edge, or router touched; no auth, sanitizer, or soul path; no dependency change. `invariant-guard` exits 0 (47/47). The only behavior change is one additional warn-once log line on the external-spend recording path.

## Benefits / why

- Removes a documented CI recipe that would break the build for anyone who followed it.
- Stops `debug: true` reading as an armed debug flag in a security-reviewed repo.
- Puts a financial-risk signal where the risk actually is: the server now says something when it prices calls against a stale table, instead of staying silent until a human runs a report.

## Risks to monitor

- The new warn-once fires from `record_external_usage`, so the first external call in a process with a stale table emits one extra WARNING. Intended, but it is a new log line on a hot path — watch that the latch holds and it does not become per-call.
- `PRICED_AS_OF`'s new test parses the `verified YYYY-MM-DD` provenance comments beside the rate rows. If a future row lands without that comment format, the test fails loudly rather than silently passing — that is deliberate, but it does couple the test to a comment convention.
- Annotating `app.debug` rather than deleting it keeps a dead key in the file. Deleting is the cleaner end state; it was left in place because removing a shipped config key is a wider change than this sweep should make unilaterally.

## Merge topology

- **Independent** — base `main`, no stack parent.
- Shares two files with #1194 (`.claude/skills/config-guard/SKILL.md`, `config.yaml`) in **non-adjacent regions**. Verified with a throwaway 3-way merge of both branches onto `origin/main`: zero conflict markers, and both sides' edits present afterward (this PR's C13 row and `--strict` warning, #1194's C5 `_CHARS_PER_TOKEN` update). Either merge order is safe.

## Checklist

- [x] Preserves all 6 invariants + I6 (`invariant-guard` 47/47)
- [x] `GROK_API_KEY=dummy pytest tests/` — only the pre-existing root-container `test_fsconnect_quota` artifact fails (it fails on unmodified `main` in that environment too; CI's non-root runner passes it)
- [x] `ruff` clean · `bandit` clean on the changed module · `config-guard` 0 failures · `doc-sync` 0 drift · config-guard's own `verify.sh` mutation self-tests pass
- [x] No new dependencies
- [x] Draft; one concern; diff reviewable in one sitting

## Further comments

Deliberately **not** included, though the sweep surfaced them:

- **`torch` missing from `pyproject.toml`** — `pip install -e .` resolves the CUDA wheel that `requirements.txt`'s `+cpu` pin exists to prevent. Real, but it is an install-contract change across four surfaces (`pyproject.toml`, `requirements.txt`, `constraints.txt`, `Dockerfile`) and both manifests already carry a written decision to defer it. Wants its own PR with real dry-runs, not a slot in a sweep.
- **Coverage `omit` excludes `agentic/{fsconnect,sqlconnect,netconnect}`** despite ~29 test files covering them — the two most write-capable surfaces in `agentic/` are invisible to `fail_under = 80`. Genuine, but fixing it likely moves the coverage floor, which needs measuring first rather than bundling here.
- **PowerShell launchers lack `Set-StrictMode`** while all nine `macos/*.sh` use `set -euo pipefail`. Plausible, but I cannot execute PowerShell in this environment, and adding strict mode blind to six installer scripts risks breaking the Windows install path — not something to ship unverified.
- **`agentic` ships `mode: write` + `writes_enabled: true`** behind `enabled: false`. Deliberate and documented across three comments; flagged only as a readability note, not a defect.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AU6cHzocQDJhzeUVc5a5iF)_


cgfixit:
@codex review

PR #1199, opened 2026-08-28
[fix] - rename memory.facts.enabled to facts.retrieval_enabled (legacy name honored); drop dead min_fts_score

qwoifhqgbio:
Audit findings F-03 and F-08 (second half). Grouped because both edit `config.yaml`'s `memory:` block and `memory/selftest.py` — splitting them by finding number would have put two branches into the same hunks.

Trial-merged clean against #1195, #1196, #1198 and against current `main`.

### F-03 — the flag's name promised something it never did

`memory.facts.enabled` gates **retrieval fusion and nothing else**. Verified: `create_proposal` and `apply_proposal` never consult it, and `GET /memory/facts` checks only the master switch. Meanwhile `stage_episode` *does* self-gate its write path on `episodes.enabled` — so the two data types behaved inconsistently, and the flag read as a master switch for facts that does not exist. An operator could legitimately see `{"facts_enabled": false, "active_facts": 3}` on `/memory/status`.

Renamed to `facts.retrieval_enabled`. `config.yaml`'s entry also gains the explanatory comment it was the only sub-switch in the whole `memory:` block to lack.

**Propose/apply behavior is unchanged**, deliberately. The staged rollout in `docs/memory/README.md` applies facts *before* fusion is enabled, and gating persistence on this flag would break that.

### The legacy name is still honored, and that is the load-bearing part

A plain rename is unsafe here. Nothing rejects unknown config keys — there is no `validate_memory_config`, and `gate.py`'s five validators don't cover `memory:` — while all four read sites use `.get()` with falsy defaults. So an operator who had turned fusion **on** would get: boot succeeds, fusion silently stops, no exception, no log, no audit event, and a `/memory/status` still reporting their facts as present. Silent degradation hitting exactly the people who opted in.

`memory/flags.py` resolves both names (new wins whenever present, even when `False`) and logs a **one-time** warning naming both keys when the legacy one supplies the value. Precedent for keeping a retired key readable and announced: `sync/config.py`'s `include_soul` deprecated no-op, and the `_unknown_keys` surfacing in `agentic/*/config.py`.

**Why a helper for a two-line read:** the flag has four readers and the gate in `retrieval/hybrid_search.py` *duplicates* the one in `retrieval_adapter.py`. One function stops them drifting — which is not hypothetical, see the mutation testing below.
- `hybrid_search` imports it **lazily**: `tests/test_memory_isolation.py` forbids a top-level `memory` import there, and reaching that line already means memory is on, so it loads nothing that wasn't about to load anyway.
- `harness/memory_notes.py` resolves the same two keys **inline** instead — `harness/` stays decoupled from the memory package, and its echo is read-only so it doesn't warn.

`mirror.py`'s status field is renamed `facts_enabled` → `facts_retrieval_enabled`. Nothing in-repo consumes it (no test, no doc, no console), and leaving it would have made it the last surviving use of the wrong word.

### F-08 (second half) — `min_fts_score` was dead

Five repo-wide occurrences, **zero reads**. `fuse_memory_hits` reads `max_hits`/`rrf_k`/`source_prefix` and explicitly discards the bm25 value as `_bm25_rank`. No test referenced it; `tests/test_memory_fusion.py` covered all three siblings and omitted this one.

**Deleted rather than wired, and the reason matters.** Wiring was technically possible — `store.py` computes `bm25(facts_fts) AS rank` and returns it as a float. But SQLite's `bm25()` is **negative, lower-is-better** (`ORDER BY rank` ASC). The obvious `rank >= min_fts_score` filter with the shipped `0.0` default would have discarded **every hit**, silently disabling fusion for anyone who enabled it. A knob whose shipped default means "no filtering" under one reading and "filter everything" under the other should be removed, not wired on spec.

**None of the 6 invariants or I6 is affected.** No graph edge, auth gate, sanitizer pattern, or soul path is touched.

- `invariant-guard`: **47 passed, 0 failed**.
- Memory's lazy-import boundary is preserved — `tests/test_memory_isolation.py` **7 passed**. The new `memory/flags.py` import in `hybrid_search` is function-local, and `memory/` still imports no forbidden package.
- No route's authentication changes; no HTTP route is added or removed.
- `/memory/status`'s payload changes one field name (see Risks).

## Types of changes

**Scope note:** Optional default-off memory subsystem + its docs. Core graph/gate/soul path untouched.

## Benefits / why

- The flag's name now matches its behavior, so an operator reading `config.yaml` cold no longer expects it to gate persistence.
- Upgrading with an old config keeps working *and* says so, instead of silently losing recall.
- A config knob that did nothing — and would have done the wrong thing if naively wired — is gone.
- The gate is now actually tested, which it wasn't (see below).

## Risks to monitor

- **`/memory/status` renames a field** (`facts_enabled` → `facts_retrieval_enabled`). Nothing in-repo reads it, and the route is API-key-gated and loopback-only, but an out-of-tree operator script parsing that JSON would need updating. This is the only wire-format change in the diff.
- **The legacy shim is a compatibility surface with no removal date.** Following `include_soul`'s precedent it should be dropped on a named future version rather than silently; until then both keys are live and `memory/flags.py` is the single place that decides.
- **The warning fires once per process.** A long-running server logs it at first fusion attempt, not at boot — there is no `validate_memory_config` to hook (the plan proposed one at `IMPLEMENTATION_PLAN.md:350`; it was never implemented). An operator who never triggers fusion never sees it.
- **The gate is still duplicated** between `retrieval_adapter` and `hybrid_search`. The shared helper means they resolve identically, but the two `if` statements remain, and the new tests exist specifically to catch them diverging.

## Checklist

- [x] This change preserves all 6 security invariants and I6 module isolation (`invariant-guard` 47/0; `test_memory_isolation` 7 passed)
- [x] For any memory change: propose/apply gating is unchanged by design; no write path altered
- [x] Relevant docs updated (`docs/memory/README.md`, `docs/memory/IMPLEMENTATION_PLAN.md`)

## Further comments

**Verification** (Python 3.12.3, ruff 0.16.1):

| Check | Result |
|---|---|
| Full suite | **4841 passed**, 71 skipped, 1 failed |
| `invariant-guard` | 47 passed, 0 failed |
| `config-guard` | 0 failures (1 warning, pre-existing C9 — verified identical on clean `main`) |
| `doc-sync` | 0 drift items |
| `python -m memory.selftest` | **ALL PASS** (11 checks, incl. `fusion` and `fusion_disabled_noop`) |

The single failure is `test_fsconnect_quota.py::test_quota_recompute_fail_closed_on_unreadable_root` — pre-existing and environmental, reproduced on clean `main`; this sandbox runs as uid 0 and root bypasses the `chmod` the test relies on. It passes in CI's non-root leg.

**The test gap this closes.** Nothing previously asserted that the facts flag gates fusion *at all* — `test_fusion_disabled_is_identity` flips the **master** switch instead. Combined with the duplicated gate, a rename that updated only one of the two call sites would have passed CI green. This adds 14 cases across both gates: new key, legacy key, both set (new wins), neither, no `facts` block, and warn-once behavior.

**Proved non-vacuous by mutation** — each of these was applied, confirmed to fail, and reverted:

| Mutation | Result |
|---|---|
| Drop the legacy fallback | 3 tests fail |
| Remove new-key precedence over legacy | 3 tests fail |
| Leave `hybrid_search`'s gate reading only the old key | 1 test fails |

That third one is exactly the half-done-rename scenario, and it was green before this PR.

**Manual upgrade proof** — a config with *only* the legacy key set, memory fully on: fusion still returns memory hits, the warning fires exactly once across two calls, and `/memory/status` reports the renamed field.

**ELI5.** A setting called `facts.enabled` sounded like the on/off switch for the whole "remembered facts" feature, but it only ever controlled one narrow thing: whether saved facts get mixed into search results. Saving and reading facts ignored it entirely. It's now called `facts.retrieval_enabled`, which is what it does. Because renaming a setting would otherwise silently turn the feature off for anyone already using it — with no error message at all — the old name still works and prints a one-time notice telling you to rename it. Separately, a nearby setting that claimed to filter weak search matches was doing nothing whatsoever, and had it been hooked up the obvious way it would have filtered out *everything*; it's removed.

---

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -214,7 +214,10 @@ memory:
   enabled: false                    # master switch
   db_path: "data/memory/cyclaw_memory.db"
   facts:
-    enabled: false
+    # Gates retrieval fusion ONLY -- never persistence, apply, or read (those
+    # are memory.enabled + propose_apply.enabled). Legacy name `enabled` is
+    # still honored with a startup warning; see memory/flags.py.
+    retrieval_enabled: false
     max_content_chars: 8192
     max_active: 10000
   episodes:
@@ -227,7 +230,6 @@ memory:
     enabled: false                  # fuse FTS fact hits into hybrid_search
     max_hits: 3
     rrf_k: 60
-    min_fts_score: 0.0
     source_prefix: "memory:fact:"
   propose_apply:
     enabled: false                  # /memory/propose|apply|reject
--- a/docs/memory/IMPLEMENTATION_PLAN.md
+++ b/docs/memory/IMPLEMENTATION_PLAN.md
@@ -324,7 +324,7 @@ memory:
   enabled: false                    # master switch
   db_path: "data/memory/cyclaw_memory.db"
   facts:
-    enabled: false
+    retrieval_enabled: false        # fusion exposure ONLY; not persist/apply/read
     max_content_chars: 8192
     max_active: 10000
   episodes:
@@ -337,7 +337,6 @@ memory:
     enabled: false                  # fuse FTS hits into hybrid_search
     max_hits: 3
     rrf_k: 60                       # match corpus RRF k
-    min_fts_score: 0.0              # bm25-ish; tune in selftest
     source_prefix: "memory:fact:"   # SearchResult.source prefix
   propose_apply:
     enabled: false                  # admin propose/apply routes useful only if true
@@ -700,7 +699,7 @@ Memory is **not** OOB like telegram. It is an optional core feature. Isolation m
 1. Deploy code (defaults off) → behavior identical to today.  
 2. Set `memory.enabled: true` + `episodes.enabled: true` → episodes start staging.  
 3. Set `propose_apply.enabled: true`, set `CYCLAW_API_KEY`, propose/apply facts.  
-4. Optionally enable `retrieval_fusion.enabled: true` after verifying FTS quality.  
+4. Optionally enable `facts.retrieval_enabled: true` + `retrieval_fusion.enabled: true` after verifying FTS quality — both are required to fuse.  
 5. Export HTML only if needed.
 
 ---
--- a/docs/memory/README.md
+++ b/docs/memory/README.md
@@ -13,10 +13,20 @@ Every switch in `config.yaml` → `memory:` is **false**. With defaults, behavio
 
 1. `memory.enabled: true` + `episodes.enabled: true` — stage query episodes (hashed query by default).
 2. `propose_apply.enabled: true` + set `CYCLAW_API_KEY` — use `/memory/propose` then `/memory/apply`.
-3. `facts.enabled: true` + `retrieval_fusion.enabled: true` — FTS fact hits fuse into `hybrid_search` as `retrieval_mode="memory"`.
+3. `facts.retrieval_enabled: true` + `retrieval_fusion.enabled: true` — FTS fact hits fuse into `hybrid_search` as `retrieval_mode="memory"`.
 4. `export_html.enabled: true` — `GET /query/export/html` (auth-gated).
 
 `consolidation.enabled` is a **stub** and must stay false in v1.
+
+Step 3 comes after step 2 on purpose: facts are proposed, applied and verified
+**before** they are exposed to retrieval. `facts.retrieval_enabled` gates only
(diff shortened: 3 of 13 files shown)


cgfixit:
@codex review

PR #1214, opened 2026-08-29
[docs] - repo-wide doc-sync: reconcile READMEs, manuals, and dependency-skill docs with current code

qwoifhqgbio:
## Proposed changes

Repo-wide documentation reconciliation, requested as a "thorough /doc-sync and cleanup" over **every** feature/functionality README plus the manuals, with a verify-deps configuration check and full run. Method: all 37 READMEs, `CLAUDE.md`, `AGENTS.md`, `PROJECT_RULES.md`, and the two dependency-skill SKILL.md files were audited claim-by-claim against code/config (three parallel read-only audits; every finding re-verified against the live file before its edit). **46 findings fixed across 14 files; 24 READMEs verified clean and untouched.**

The headline findings:

1. **`CLAUDE.md` §5/§6 and `AGENTS.md` claimed ruff is "CI-enforced." It is not.** `lint.yml` runs Ruff with job- AND step-level `continue-on-error` — its own header says a red result can never block merge. An agent trusting the manuals would ship a lint regression believing CI catches it. Both manuals now say "advisory," in lockstep. ⚠️ **Owner decision flagged, not made here:** if you *want* ruff blocking, that's a small `lint.yml` change riding its own PR (repo precedent: CI-workflow changes are their own reviewable concern).
2. **The executor-sandbox understatement fixed in #1210 existed at three more sites** — root `README.md`'s tree comment and containment bullet (which cited `runner.py` for a claim `runner.py` now contradicts), and `agentic/README.md`'s module row. All now describe the required fail-closed `production_sandbox()` with the honest residual limits.
3. **`README.md` said "every gate below `agentic.enabled` defaults to `false`"** — contradicted by `config.yaml` (three cloud-provider switches ship armed) and by the README's own armed-cloud note 200 lines later.
4. **`PROJECT_RULES.md` was the drift-heaviest file in the repo** (7 findings): it required `soul.md` to exist before boot (the code self-heals; `CLAUDE.md` §4 explicitly says "do not add a boot crash"), counted 16 coverage sources (17 — `opentweet` missing), prescribed `pydantic-settings` (never a CyClaw pattern — chromadb transitive only), pointed at a nonexistent `/sandbox-runtime-verification` skill, and listed only 3 of the 6 I6 core modules.
[...]

PR #1216, opened 2026-08-29
[fix] - collapse raced unique violations into typed auth errors; stop the harness 500ing on a duplicate user

qwoifhqgbio:
## Proposed changes

Audit finding F-05. Verified by reading the code rather than taken on the finding's word — and it is **worse than reported**, in a way that needs no race at all.

### 1. Both account-creating writes are check-then-act, and only one of three defends it

`create_user` and `create_device_token` are both SELECT-then-INSERT. The `self._lock` guarding each orders only threads sharing *that* manager — the gateway, the harness and the `cyclaw-user` CLI each hold their own.

**The repo already knew.** `bootstrap_if_empty` carries the comment *"Gateway and harness each hold their own AuthManager. Both can observe an empty table and race the INSERT; the PK on username makes the loser a unique violation"* — and it was the **only** one of the three that wrapped its INSERT.

**The helper already existed too.** `_is_unique_violation` normalizes `sqlite3.IntegrityError` and psycopg's SQLSTATE `23505` (the two backends share no base class) and had exactly one caller. This PR takes it from 1 call site to 3. Nothing new was invented.

What the raw backend error did before:

| Surface | Before | After |
|---|---|---|
| `POST /auth/users` (gate) | unhandled **500** — `_raise_auth_error`'s ladder ends in a bare `raise exc`, and `gate.py` registers only a `RequestValidationError` handler | **409** `AUTH_USER_EXISTS` |
| `cyclaw-user add` | exit **1** — a backend error is not an `AuthError`; the documented contract is `0`/`2`/`3` | exit **2** |
| Postgres connection | `connect()` sets `autocommit=False`; nothing rolled back, so the failed INSERT left the long-lived connection raising `InFailedSqlTransaction` until something else cleared it | rolled back, reusable |

### 2. The harness 500s on an ordinary duplicate — no race required

`harness/server.py`'s `harness_create_user` called `manager.create_user(...)` bare. With only a `RequestValidationError` handler registered there too, **every** repeat create returned 500, while `gate_auth.py`'s equivalent route returned a clean 409. The two HTTP surfaces disagreed on the same manager error and nothing asserted either one.

Reproduced before fixing: the new test returns **500** against the unfixed handler and **409** against the fixed one.

**Invariant / Governance Impact:** none. No graph node, edge or router; no sanitizer pattern, soul path, config value or dependency. `invariant-guard` **47/47**. This *narrows* observable behavior (a 500 becomes a typed 409); it grants no new capability and relaxes no check.

## Types of changes

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation Update

**Scope note:** Auth subsystem (`utils/authn_manager.py`) + the harness console's auth route. No core graph/soul path.

## Benefits / why

- **An operator adding a user that already exists gets an answer instead of a stack trace.** On the harness console that was the *normal* path, not an edge case.
- **The two HTTP surfaces stop disagreeing.** They now return the same status and the same typed code for the same underlying error.
- **Postgres stops getting poisoned.** A failed INSERT with no rollback under `autocommit=False` breaks every subsequent statement on that shared connection — an availability bug, not a cosmetic one.
- **Exit codes stay an API.** `CLAUDE.md` §4 says so explicitly; exit 1 was outside the contract.

## Risks to monitor

- **`except Exception` is broad.** Deliberately mirrors `bootstrap_if_empty`'s existing shape, and every non-unique-violation is re-raised unchanged — but it does mean an unrelated backend fault now passes through a `rollback()` first. That is the correct move for a failed write regardless of cause.
- **The 409 body names the username** (`user already exists: <name>`), so it confirms existence. Not a new disclosure: `POST /auth/users` requires an authenticated admin-or-operator actor, and both roles can already read the full roster via `GET /auth/users`. Unauthenticated callers get 401 first.
- **`create_device_token` is in scope on purpose.** It has the identical shape against the partial unique index `idx_device_tokens_live_label`. Widening a "surgical" diff by ~6 lines was the lesser evil against knowingly shipping a half-fix one method away from a verified-identical race — see the scope note below.
- **The race window is not closed, only its symptom.** These are still check-then-act; the DB constraint remains the real arbiter. Closing the window properly means `INSERT ... ON CONFLICT`, which is a larger change to two backends' SQL and is not what this PR is.

## Checklist

- [x] Preserves all 6 security invariants and I6 module isolation — `invariant-guard` **47 passed, 0 failed**
- [x] Full suite green: **5020 collected, exit 0, zero failures** (baseline on `main` at `6092171`: 5014 collected, zero failures — this branch is +6, all new)
- [x] No new external network dependencies or online-LLM assumptions
- [x] For any harness change: no write guard, quota, or auth *policy* altered — only the error mapping of an already-refused operation
- [x] Relevant docs updated — **deliberately none**, see Further comments
- [x] `ruff check --select E,F,I,B,C4,UP,S .` clean · `config-guard` 0 failures · `dep-guard` 0/0 · `doc-sync` 0 drift

### Every new test was mutation-checked

A test that only ever passes is indistinguishable from one that never runs, so each was run against the reverted fix:

| Test | Against the unfixed code |
|---|---|
| `test_raced_duplicate_create_is_still_auth_user_exists` | `sqlite3.IntegrityError: UNIQUE constraint failed: users.username` |
| `test_raced_duplicate_create_leaves_the_connection_usable` | same |
| `test_raced_duplicate_token_label_is_still_auth_token_label_exists` | `sqlite3.IntegrityError: UNIQUE constraint failed: device_tokens.username, device_tokens.label` |
| `test_raced_duplicate_create_is_409_not_an_unhandled_500` | 500 |
| `test_duplicate_create_is_409_not_an_unhandled_500` (harness) | **500** — the plain-duplicate bug, reproduced |

### How the race is forced

Deterministically, not with threads: the pre-check SELECT is blinded to **one username**, which is exactly what a writer committing in the window looks like from inside the manager, and leaves the DB constraint as the only remaining defence. Scoped to one username on purpose — the same statement resolves the session's own user, so blinding it wholesale 401s the request before it ever reaches `create_user`. (Found that the hard way; the first draft did exactly that and got a confusing 401.)

`test_duplicate_create_is_409_with_a_typed_code` is a **characterization** test, not a proof of this fix — `gate_auth.py` already mapped `AuthUserExists` to 409. It is here because `grep AUTH_USER_EXISTS tests/` returned **zero hits** before this PR, which is precisely why the harness copy could diverge and 500 unnoticed.

### Deliberately not done

- **No CLI-layer raced test.** The manager test pins the exception *type*, and `test_authn_cli.py:68-73` already pins that an `AuthUserExists` maps to exit 2. A raced-CLI test would exercise the same contract through more scaffolding.
- **No doc updates.** `CLAUDE.md`'s route table would ordinarily earn a line here, but PR #1214 was sweeping 14 doc files including that one while this was written (it has since merged). Avoiding the shared-file trap in `CLAUDE.md` §4 was worth more than the table row; flagging it here instead.

### ELI5

Two places in CyClaw create things that must be unique — a user account and a named device token. Both first ask "does this already exist?" and then create it. Between those two steps another program can slip in, because the lock that guards them only covers one program at a time and CyClaw runs three that share the database.

When that happened, the database itself refused the duplicate — correctly — but nobody caught its complaint. The web server turned it into a generic "500 server error", the command-line tool exited with a code that isn't in its documented list, and on PostgreSQL the failed write jammed the shared connection so *later* unrelated commands started failing too.

Separately, and much simpler: on the harness console, creating a user that already exists returned a 500 **every single time** — no race needed, the error just wasn't caught there at all — while the main server returned a polite "409, that name is taken". Now both say the same thing, and the tests were each checked by undoing the fix and confirming they fail.

---


cgfixit:
Oh wow is this that same issue from a week or so ago just deeper rooted or does that same type of logical failure mode appear in more than one decision tree?

cgfixit:
Oh and also isn't ollama inherently single channel or was that lm studio? There's probably an upstream reason if it's confirmed this is a separate but similar style mistake

qwoifhqgbio:
Both good questions, and the second one has a correction in it. Checked rather than recalled.

## 1. Same class, recurring — and this file already fixed it once, differently

You're right that it recurs. `utils/authn_manager.py` now carries **three** answers to the same hazard, and PR #940 got there first — its comment is still in the file, twenty lines above the code this PR touches:

> *"Last-admin guard, folded INTO each mutating statement (PR #940 review, check-then-act race): a separate SELECT-then-WRITE pair leaves two windows open — two threads through this instance if the check runs outside `self._lock`, and **two PROCESSES regardless** (the HTTP admin surface and a concurrent `cyclaw-user role/disable`, each with its own AuthManager and its own lock)."*

That describes this PR's bug exactly, a year of commits before it. The three strategies in one file:

| Where | Strategy | Result for the loser |
|---|---|---|
| `set_role` / `disable` / `delete`, `_sql_claim_login` (#940) | fold the condition **into** the SQL — compare-and-swap | `rowcount == 0`, write never happens |
| `bootstrap_if_empty` | catch the unique violation, treat as a no-op | write refused by the PK |
| `create_user` / `create_device_token` (**this PR**) | catch the unique violation, re-raise typed | write refused by the PK |

**But the difference is principled, not laziness — I checked before concluding it.** #940 *had* to fold the check into the statement because "at least one enabled admin must remain" is a `COUNT(*)` over the table, which no DB constraint can express. For `create_user` the constraint already exists (`username TEXT PRIMARY KEY`), so the database was **already** serializing correctly and refusing the duplicate. Nothing was racing through. What was broken was only the *classification* of that refusal: the raw `sqlite3.IntegrityError` / psycopg `UniqueViolation` matched no branch of `_raise_auth_error`'s ladder and escaped as a 500.

So: same failure mode, different layer. #940 fixed *"the wrong thing happens."* This fixes *"the right thing happens and is then reported as a crash."*

### It is not confined to auth

The same shape shows up in at least five subsystems, each with its own fix:

- **`utils/numbat_emitter.py`** — size-check then rename. Two writers both rotated, and the loser's `os.replace` **overwrote the archive** rather than appending. Fixed in #1208 with an `O_CREAT|O_EXCL` lock plus a size *and inode* re-check under it.
- **`agentic/fsconnect/writer.py`** — name-based TOCTOU. Closed on POSIX with held file descriptors; on Windows, where it cannot be closed, **writes are hard-refused** rather than shipped with a residual window.
- **`agentic/deepagent_github/repo_workspace.py`** — `dir_fd`, documented "zero TOCTOU window."
- **`agentic/executor/manifest.py`** — approval bound to a content digest, so a post-approval edit invalidates it.
- **`harness/web_search.py:213`** — a DNS TOCTOU explicitly documented as residual and accepted.

The repo has effectively grown a pattern library for this, and in one case chose to *refuse the feature* rather than accept the window. What it does **not** have is anything that would have caught the gap here — no linter or guard flags a SELECT-then-INSERT pair, which is why two of the three sites in this one file went unnoticed until now.

## 2. Ollama — half right, and it doesn't apply here

**Half right:** it is single-channel in CyClaw, but by *our* configuration, not inherently. `macos/ollama-mlx.env:28` pins `OLLAMA_NUM_PARALLEL=1`, deliberately:

> *"Metal decode is memory-bandwidth bound. Parallel slots multiply KV RAM (NUM_PARALLEL × CONTEXT_LENGTH) and stall both /query and /loop."*

`docs/! How-To-Guides/OLLAMA_SETUP.md:400` says outright *"Do **not** raise `OLLAMA_NUM_PARALLEL`"*, and `tests/test_ollama_mlx_env.py:31` pins the value. Ollama itself supports parallel slots; we turn them off for RAM on a 48 GB box. (I can't tell you what LM Studio's default was — that predates what I can verify from this tree, so I'm not going to guess.)

**But it doesn't apply here**, and this is the part worth being clear about: **the auth path never touches the LLM.** `utils/authn_manager.py` imports stdlib plus `utils.authn`, `utils.authn_store` and `utils.errors` — nothing else. The concurrency driving this race is entirely upstream of any model call:

1. FastAPI's threadpool inside `gate.py` — several threads, one `AuthManager`, one lock;
2. `gate.py` and `harness/server.py` as **separate processes**, each with its own `AuthManager` and its own lock;
3. `cyclaw-user` as a **third process** on the same DB file.

`OLLAMA_NUM_PARALLEL=1` serializes *inference*. It does nothing to two processes writing `users`. So the race is fully live at that setting — which is what makes your instinct about an upstream reason worth stating explicitly and then setting aside: there is one, and it's the wrong upstream.

---

Nothing to change in this PR from either question, unless you want the stronger #940 treatment (`INSERT … ON CONFLICT DO NOTHING` + rowcount) applied to these two inserts as well. I'd argue against it: the PK already does the serializing, so it would add SQL divergence between the two backends to buy nothing the `except` clause doesn't already deliver. Happy to be overruled.

---

qwoifhqgbio:
# Fix the harness same-origin port-blindness — the audit finding I stopped short of.

Research pass complete — real code plus the official definitions, every load-bearing claim spot-checked in the tree. No code changed; this is the findings report for your go/no-go. One correction to my own earlier rating is included, because the research contradicted me.

## The gap, precisely

`harness/server.py:619` `_enforce_same_origin` checks `urlparse(origin).hostname not in _LOOPBACK_HOSTS` — **hostname only**. The official definition of an origin is the **scheme / host / port** tuple (MDN Same-origin policy, which lists same-host-different-port explicitly as a *failure* case, with no localhost exception). So the harness accepts three things it shouldn't call same-origin:

| Origin sent to `:8790` | harness today | per the SOP tuple |
|---|---|---|
| `http://127.0.0.1:8787` (the gate console) or `:3000` (any dev server) | **pass** | cross-origin |
| `https://localhost:8790` (wrong scheme) | **pass** | cross-origin |
| `http://localhost:notaport` (malformed port) | **pass** — `.port` is never read, so the lazy `ValueError` never fires | gate/gate_auth refuse this explicitly, fail-closed |

That third row is a divergence beyond the audit's original finding. Meanwhile `gate.py:1453` and `gate_auth.py:268` compare all three dimensions against the request's own URL — the #1205 fix. `git show a0f9d7b --stat` confirms **`harness/server.py` was simply not in that PR**; the harness was left behind, and two claims rotted with it:

- `docs/THREAT_MODEL.md:1013-1015` still says the gate change is *"matching `harness/server.py`'s `_enforce_same_origin`, which already covered the harness for this"* — **no longer true.**
- `gate.py:1401-1405`'s own comment describes the harness's exact current behavior as the bug that was fixed: *"it used to ask only 'is the Origin's host loopback?' … Too lax: it ignored the port outright."*

## Severity — correcting my own earlier call

I rated this MEDIUM-HIGH in the audit, reasoning that `auth_open` routes carry no CSRF to compensate. The research shows that overstated it, because **a different arm of the same function stops the realistic attack today**:

1. **`Sec-Fetch-Site` rejects `same-site`.** Line 639 allows only `same-origin`/`none`. A page on `http://127.0.0.1:3000` fetching `:8790` is cross-origin but **same-site**, so a modern browser stamps `Sec-Fetch-Site: same-site` → 403 before the Origin arm is ever consulted. The port blindness only bites for a client that sends `Origin` *without* `Sec-Fetch-Site` — old browsers, or non-browser clients that are not CSRF vectors anyway.
2. **The JSON routes can't be reached simply.** `login`/`bootstrap-password` take Pydantic JSON bodies; `Content-Type: application/json` forces a preflight the harness cannot answer (no CORS middleware), and a `text/plain` downgrade 422s — verified against the installed FastAPI's body handling.
3. **The bodyless `guarded` POSTs** (`/api/web/inject`, `/api/chat/cancel`, `/api/agent/runs/{id}/push`, …) *are* CORS-simple, but the per-process CSRF token — unreadable cross-origin — remains as the last barrier.
4. **One spec-level caveat, flagged not asserted:** SameSite is *site*-scoped and ports are not part of a site, so `SameSite=Strict` does **not** withhold the session cookie between two loopback ports (RFC 6265bis reading, not browser-tested). That's why the layers above matter.

**Net: a defense-in-depth regression plus a false parity claim in the threat model — not a demonstrated live exploit.** Real, cheap to fix, not urgent-severity. My audit line was wrong to imply the CSRF gap made this hot; the Sec-Fetch-Site arm was the compensating control I missed.

## The fix, if you want it (one hunk + one test)

Replace the Origin arm (`harness/server.py:648-667`) with the gate_auth comparison shape: hostname equals `request.url.hostname` AND hostname in `_LOOPBACK_HOSTS` AND `parsed.port == request.url.port` (with the lazy `.port` read wrapped so a malformed port fails closed) AND `parsed.scheme == request.url.scheme`. No wildcard logic needed — the allow-set is the fixed loopback tuple `TrustedHostMiddleware` already uses. The Sec-Fetch-Site arm stays as-is.

Exactly **one existing test pins the old behavior**: `tests/test_harness_auth.py:409` (`test_loopback_origin_is_allowed`) sends `Origin: http://127.0.0.1:8790` against a client whose `base_url` has no port — all three parametrized cases fail under a strict check for exactly the reason `tests/test_gate.py:1327-1340`'s rewritten equivalent documents ("an Origin case must state the port its base_url actually connected on"). That test rewrite is part of the fix, with `test_gate_auth.py:387-477`'s four regression cases as the template. Everything else survives: cross-origin 403s, malformed-Origin 403s, absent-headers-allowed (curl/PowerShell/the sandbox verifier — and per MDN, browsers always send Origin on POST, so the absent-Origin allowance costs nothing against browser attacks).

**Your calls, not mine:**
- Ship it at all? (Parity + closing the malformed-port divergence + making THREAT_MODEL true again vs. no demonstrated exploit.)
- `GET /api/auth/setup-status` carries **no same-origin check at all** (rate limit only) — matches the gate's route table posture, so probably deliberate; confirm or I fold it in.
- The cookie's hardcoded `secure=False` vs gate_auth's `secure=tls_enabled` — harmless today (harness never serves TLS, verified), separate one-liner if wanted.
- Does anything in your setups front the harness with a port-rewriting proxy? A strict check would 403 it (same residual class as the `api_key_optional` forwarding-header rule).

## Current workaround for the SELECT-then-INSERT class: the guard is live

While this finding waits on your decision, the other commissioned piece is shipped as **#1219** (draft): a static guard so the check-then-act class this PR fixes gets caught by CI instead of re-discovered by review — which is how all three historical instances were found (#940, then the audit's two).

- **Rule:** in `utils/authn_manager.py`, a SELECT followed by an INSERT in the same function must wrap the INSERT in a `try` with a real `except` (a `try/finally` doesn't count — it re-raises) or fold the conflict into the SQL (`ON CONFLICT`/`OR IGNORE`/`OR REPLACE`). The pre-check SELECT is the confession that a violable constraint exists.
- **Scope is that one module, as a measured decision:** a full inventory of every INSERT-executing function in the repo found all instances of the class there, and found three structural look-alikes elsewhere (`personality._load_soul`, `memory._insert_fact_conn`, `stage_episode`) that are safe only because of schema facts — exonerating those mechanically needs a DDL oracle, deliberately not built until the class appears in a second module. The guard's docstring names them.
- **Interplay with this PR:** on today's `main` the guard fires on exactly `create_user` and `create_device_token` — verified live — so those two names ride a temporary accepted set whose comment orders its own deletion once this PR merges. The assertion is subset-shaped, so **merging #1216 turns the guard quiet, not red** (verified by swapping this branch's `authn_manager.py` in: zero findings, all tests green). Merge order doesn't matter; the accepted entries come out afterwards either way.
- The detector itself is pinned by six synthetic-source tests (bug shape, real handler, the `try/finally` trap, folded SQL, no-precheck, and the `self._sql_*` f-string indirection), so the mutation checks are permanent tests rather than a claim in a PR body.

---

cgfixit:
## Review

Gate, harness, and `cyclaw-user` each hold their own `AuthManager` lock on one users table. `create_user` / `create_device_token` are still SELECT-then-INSERT. The PK / live-label unique index already refuses the loser. What was broken is classification of that refusal:

| Surface | Before | After |
|---|---|---|
| `POST /auth/users` (gate) | unhandled 500 on a raced unique violation | 409 `AUTH_USER_EXISTS` |
| harness `POST /api/auth/users` | 500 on an *ordinary* duplicate — no race required | 409 `AUTH_USER_EXISTS` |
| `cyclaw-user add` | exit 1 (not an `AuthError`) | exit 2 |
| Postgres | failed INSERT left `autocommit=False` connection in `InFailedSqlTransaction` | rollback, reusable |

Reusing `_is_unique_violation` (1 call site → 3) is the right shape. Do not fold these inserts into `#940`-style compare-and-swap / `ON CONFLICT`: last-admin is a `COUNT(*)` no constraint can express; username uniqueness already is a PK. Unifying them would add SQLite/Postgres SQL divergence for no extra safety.

Default `auth.enabled: false` is why this stayed latent. It is not a credential-store failure and not a double-create. The write was refused; one HTTP surface reported it as a crash.

### Leave in this PR
- Check-then-act window stays open; the constraint remains the arbiter. Correct scope.
- Broad `except Exception` + rollback-first matches `bootstrap_if_empty`. Non-unique faults re-raise.
- 409 body naming the username is not new disclosure (`POST /auth/users` is already admin/operator; roster is readable).
- Harness maps `AuthUserExists` only. No harness `create_device_token` route found; token path is manager + CLI. Acceptable.
- Tests that blind `_sql_get_user` / `_sql_get_live_token_by_label` are ugly and deterministic. Prefer that over threaded flake.

### Out of scope — do not land here
Harness `_enforce_same_origin` is still hostname-only (`#1205` covered gate / `gate_auth` only). THREAT_MODEL currently claims harness parity that is false. Compensating control today is `Sec-Fetch-Site`. Separate PR.

Follow-up after merge: `#1219` allowlist entries for `create_user` / `create_device_token` must be deleted so the guard is live, not grandfathered.

cgfixit:
Pending task:

The repo has effectively grown a pattern library for this, and in one case chose to refuse the feature rather than accept the window. What it does not have is anything that would have caught the gap here — no linter or guard flags a SELECT-then-INSERT pair, which is why two of the three sites in this one file went unnoticed until now.

PR #1218, opened 2026-08-29
[fix] - make every memory-flag echo strict, finishing the sweep #1207 started

qwoifhqgbio:
## Proposed changes

From the merge-train completeness audit of `79c1a03..8a3fb2e`. This is the **highest-severity finding** in that range, and it is the sibling pass #1207 explicitly deferred in its own body.

#1199 put the facts flag behind `memory/flags.facts_retrieval_enabled`, whose docstring states the reason plainly: *"Three readers of one flag drift; one function does not."* #1207 made the harness console's `facts` echo strict to match. **The siblings turned out to be the larger half.**

### `memory/mirror.py.status_dict` — the body of `GET /memory/status`

Internally inconsistent inside a single dict literal: the shared helper for `facts`, `is True` for the master key, then hand-rolled `bool()` for the other five.

| flag | echo (before) | its real gate |
|---|---|---|
| `episodes.enabled` | `bool()` | `memory/store.py` `is not True`; `graph.py` `is True` |
| `retrieval_fusion.enabled` | `bool()` | `retrieval/hybrid_search.py` + `memory/retrieval_adapter.py` `is not True` |
| **`propose_apply.enabled`** | `bool()` | `gate_memory.py` `is True` |
| **`export_html.enabled`** | `bool()` | `gate_memory.py` `is True` |
| `consolidation.enabled` | `bool()` | `memory/consolidation.py` `is not True` |

The two bolded ones are the **write and export gates**.

### `harness/memory_notes.rag_flags` — the worst single case

The master `enabled` key was loose. **Six** independent gates read that key strictly — `memory/store.py`, `memory/retrieval_adapter.py`, `retrieval/hybrid_search.py`, `graph.py`, `gate_memory.py`, `memory/mirror.py` — and this echo was the last loose reader left in the repo.

### Why a quoted YAML value is the realistic trigger, not a contrived one

The `memory:` block has **no validator**. There is no `validate_memory_config`, and `gate.py`'s validators do not cover it, so `enabled: "false"` boots silently. Under `bool()` every non-empty string is truthy, so both consoles reported a capability **ON** while its gate refused it — and `"true"`, the value an operator writes when they mean to switch something *on*, fails identically:

| configured | console said | gate actually |
|---|---|---|
| `"false"` | on | **off** |
| `"true"` | on | **off** |
| `"no"` | on | **off** |

**Invariant / Governance Impact:** none. No graph node, edge or router; no sanitizer pattern, soul path, route, config value or dependency. This *narrows* two status payloads to match gates that were already strict. `invariant-guard` **47/47**, `doc-sync` 0 drift.

## Types of changes

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation Update

**Scope note:** Optional memory subsystem status surface + the out-of-band harness console echo. No core graph/gate/soul path.

## Benefits / why

- **Neither console can now claim a write or export capability its route 404s.** That is the same silent-and-misleading failure shape #1199 and #1207 existed to remove, on the endpoint that is the operator's only feedback for a block nothing validates.
- **It closes a gap this repo had already committed to closing.** #1207 shipped `facts` and said the dict "stays half-strict until a separate pass." No such pass existed in the train.
- **One less special case.** After this, every reader of every `memory:` flag — six gates and two echoes — treats the value the same way.

## Risks to monitor

- **Stricter parsing surfaces latent config errors as "off".** An operator running with a quoted `propose_apply: "true"` was already getting 404s from those routes; after this the console stops disagreeing and reports `false`. Their behavior is unchanged — only the reporting becomes truthful — but a flip in the status payload could read as a regression if the quoting is not noticed. That is the intended outcome, flagged so it is not mistaken for one.
- **Two consoles change their payload at once.** `GET /memory/status` and the harness `/api/memory` echo both shift for the same configs. Nothing consumes these programmatically in-tree (`test_terminal_contract` and `test_harness_console_contract` check routes, not payload values), but an external script keying off a quoted config would see the change.
- **The remaining loose `bool()` calls in the repo were not swept.** This PR is scoped to the memory flag set, where the divergence-from-gate was verified case by case. A repo-wide `bool()` → `is True` sweep would be a different, much larger change with no such verification behind it.

## Checklist

- [x] Preserves all 6 security invariants and I6 module isolation — `invariant-guard` **47 passed, 0 failed**; `test_memory_isolation` and `test_harness_isolation` green
- [x] Full suite green: **5036 passed, 73 skipped, 0 failed** (baseline on `main`: 4946/73/0 — this branch is +90, all new)
- [x] No new external network dependencies introduced
- [x] For any harness change: no write guard, quota, or auth behavior altered — this is a read-only status echo
- [x] Relevant docs updated — `doc-sync` 0 drift

### The tests assert agreement, not a fixed value

90 parametrized cases over ten value shapes — `"false"`, `"true"`, `"no"`, `"yes"`, `""`, `0`, `1`, `None`, `False`, `True`. Each asserts `echo is (value is True)`, i.e. that the echo **agrees with the gate**, rather than pinning an expected constant. The contract is the mirror, so the mirror is what must hold; that also means the tests keep working if a gate's own semantics are ever deliberately changed.

Mutation-checked: restoring `bool()` fails them, including on the `"true"` and `"yes"` shapes that fail in the *inverse* direction from the one a reviewer would guess.

### Found by, not fixed in, the audit

This came out of a completeness survey of the 7-PR merge train, hunting specifically for fixes applied at one call site but not at identical siblings. Six other findings from that survey are reported separately rather than fixed here — including a port-blind third copy of the same-origin predicate in `harness/server.py`, which is a security change and warrants its own decision rather than riding along in this diff.

### ELI5

CyClaw's memory feature has several on/off switches, and two dashboards that show you which are on. In a YAML config file, `enabled: true` and `enabled: "true"` look almost identical — the quotes are easy to add by accident.

The actual switches only accept the unquoted `true`. The dashboards used a looser check where **any** text at all counted as "on" — so if you quoted the value, the switch was off but the dashboard cheerfully said on. Worse, quoting `"true"` is what someone does when they are *trying* to turn something on, so they would be told it worked when it hadn't.

Two of the affected switches control writing and exporting your saved memories. And nothing else in CyClaw checks this config file for mistakes, so the dashboard was the only place you'd ever find out — and it was the thing lying.

Both dashboards now read the switches exactly the way the switches read themselves.

---

### DIFF
--- a/harness/memory_notes.py
+++ b/harness/memory_notes.py
@@ -87,10 +87,13 @@ def rag_flags(cfg: Mapping[str, Any] | None) -> dict[str, bool]:
     episodes = block.get("episodes")
     fusion = block.get("retrieval_fusion")
     return {
-        _ENABLED_KEY: bool(block.get(_ENABLED_KEY)),
+        # `is True` throughout, matching _facts_retrieval_flag and the gates all
+        # four mirror. The master key is the worst offender: six independent
+        # gates read it strictly, and this echo was the only loose reader left.
+        _ENABLED_KEY: block.get(_ENABLED_KEY) is True,
         "facts": _facts_retrieval_flag(facts),
-        "episodes": bool(episodes.get(_ENABLED_KEY)) if isinstance(episodes, dict) else False,
-        "retrieval_fusion": bool(fusion.get(_ENABLED_KEY)) if isinstance(fusion, dict) else False,
+        "episodes": episodes.get(_ENABLED_KEY) is True if isinstance(episodes, dict) else False,
+        "retrieval_fusion": fusion.get(_ENABLED_KEY) is True if isinstance(fusion, dict) else False,
         "writable_from_harness": False,
     }
 
--- a/memory/mirror.py
+++ b/memory/mirror.py
@@ -17,11 +17,16 @@ def status_dict(cfg: Mapping[str, Any]) -> dict[str, Any]:
         "enabled": enabled,
         "db_path": mem.get("db_path", "data/memory/cyclaw_memory.db"),
         "facts_retrieval_enabled": facts_retrieval_enabled(mem),
-        "episodes_enabled": bool((mem.get("episodes") or {}).get("enabled")),
-        "retrieval_fusion_enabled": bool((mem.get("retrieval_fusion") or {}).get("enabled")),
-        "propose_apply_enabled": bool((mem.get("propose_apply") or {}).get("enabled")),
-        "export_html_enabled": bool((mem.get("export_html") or {}).get("enabled")),
-        "consolidation_enabled": bool((mem.get("consolidation") or {}).get("enabled")),
+        # `is True`, not bool(): every gate these five report on accepts only
+        # the literal True, so a quoted YAML value ("true"/"false" -- both
+        # truthy strings) made this echo disagree with the gate. The memory
+        # block has no validator, so /memory/status is the operator's only
+        # feedback that a quoted value took effect as "off".
+        "episodes_enabled": (mem.get("episodes") or {}).get("enabled") is True,
+        "retrieval_fusion_enabled": (mem.get("retrieval_fusion") or {}).get("enabled") is True,
+        "propose_apply_enabled": (mem.get("propose_apply") or {}).get("enabled") is True,
+        "export_html_enabled": (mem.get("export_html") or {}).get("enabled") is True,
+        "consolidation_enabled": (mem.get("consolidation") or {}).get("enabled") is True,
         "active_facts": 0,
         "episodes": 0,
     }
--- a/tests/test_config_flag_echo_strictness.py
+++ b/tests/test_config_flag_echo_strictness.py
@@ -55,12 +55,20 @@ def _cfg(block: dict) -> dict:
     return {"memory": block}
 
 
+def _block(tmp_path, **extra: object) -> dict:
+    # A real path under tmp_path, never the sqlite ":memory:" sentinel:
+    # memory.store.connect treats db_path as a filesystem path and CREATES it,
+    # so ":memory:" produces a file literally named ":memory:" in the CWD --
+    # a path Windows cannot check out (`error: invalid path`). Learned the hard
+    # way; tests/test_repo_hygiene.py now guards the repo against it.
+    return {"enabled": True, "db_path": str(tmp_path / "flags.db"), **extra}
+
+
 @pytest.mark.parametrize("value", _SHAPES)
 @pytest.mark.parametrize("section", sorted(_MIRROR_KEYS))
-def test_mirror_status_echo_agrees_with_the_gate(section: str, value: object):
+def test_mirror_status_echo_agrees_with_the_gate(section: str, value: object, tmp_path):
     """GET /memory/status must not claim a capability its gate refuses."""
-    block = {"enabled": True, "db_path": ":memory:", section: {"enabled": value}}
-    out = status_dict(_cfg(block))
+    out = status_dict(_cfg(_block(tmp_path, **{section: {"enabled": value}})))
     assert out[_MIRROR_KEYS[section]] is (value is True), (
         f"/memory/status reported {section}={out[_MIRROR_KEYS[section]]!r} "
         f"for a configured value of {value!r}; every gate on this key accepts "
@@ -69,8 +77,8 @@ def test_mirror_status_echo_agrees_with_the_gate(section: str, value: object):
 
 
 @pytest.mark.parametrize("value", _SHAPES)
-def test_mirror_master_enabled_echo_agrees_with_the_gate(value: object):
-    out = status_dict(_cfg({"enabled": value, "db_path": ":memory:"}))
+def test_mirror_master_enabled_echo_agrees_with_the_gate(value: object, tmp_path):
+    out = status_dict(_cfg({"enabled": value, "db_path": str(tmp_path / "flags.db")}))
     assert out["enabled"] is (value is True)
 
 
--- a/tests/test_config_flag_echo_strictness.py
+++ b/tests/test_config_flag_echo_strictness.py
@@ -1,0 +1,89 @@
+"""Every memory-flag echo must agree with the gate it reports on.
+
+PR #1199 renamed `memory.facts.enabled` to `facts.retrieval_enabled` and put the
+resolution behind `memory/flags.facts_retrieval_enabled`, whose module docstring
+states the reason: *"Three readers of one flag drift; one function does not."*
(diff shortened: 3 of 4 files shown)


qwoifhqgbio:
**CI: three Windows jobs failed on `ab2d6f4` — runner/checkout failure, not this diff.**

Failing checks, all in run [33265667097](https://github.com/cgfixit/CyClaw/actions/runs/33265667097):

- `windows-latest`
- `windows-2022 / Windows PowerShell 5.1 installer`
- `deepagents-harness (windows-latest)`

### Why it is not this PR's

**They died before any test body ran.** All three started at 17:27:15 and failed at 17:27:25–17:27:29 — 10 to 14 seconds. The `windows-latest` log is 143 lines total and ends in the checkout/auth-cleanup phase with:

```
##[error]The process 'C:\Program Files\Git\bin\git.exe' failed with exit code 128
```

A job that reached the suite would be thousands of lines.

**The same jobs pass on the same base.** #1216 was cut from the identical base (`6092171`) and pushed twelve minutes earlier; its `windows-2022 / Windows PowerShell 5.1 installer` and `deepagents-harness (windows-latest)` both came back **success**.

**Nothing in this diff is platform-sensitive.** It changes `bool(x)` to `x is True` in two dict literals (`memory/mirror.py`, `harness/memory_notes.py`) and adds one test module. No path handling, no subprocess, no filesystem semantics — nothing that could distinguish Windows from the Linux and macOS legs, which are green or still running on this same commit.

### No fix ported, because there is nothing to port

This is not a code failure with an available fix — it is a runner-side git failure during checkout. Skipping, quarantining, or working around it would be inventing a problem to solve.

### Re-run

Attempting the one permitted re-run now returns `403 This workflow is already running` — the run is still live (`ci (3.12)`, `ubuntu-latest`, `macos-latest` and others in progress). I will re-run the failed jobs once the workflow completes, and that is the single re-run spent on this failure; a second failure would be real and I would treat it as this PR's.

Local verification on this exact commit for reference: **5036 passed, 73 skipped, 0 failed** (baseline on `main`: 4946/73/0), `invariant-guard` 47/47, `ruff` clean, `doc-sync` 0 drift.

---

qwoifhqgbio:
**Correction: my previous comment was wrong. The Windows failures were this PR's fault, and they are now fixed in `19a2b2f`.**

I called it a runner/checkout failure. It was a checkout failure, but the reason was a file I committed.

### The actual error

The re-run failed identically, so I read the full checkout step instead of the log tail:

```
[command] git checkout --progress --force refs/remotes/pull/1218/merge
##[error]error: invalid path ':memory:'
##[error]The process 'git.exe' failed with exit code 128
```

There was a zero-byte file in the repo root **literally named `:memory:`**. Windows forbids `:` in a path component, so `git checkout` of the merge ref aborts and every Windows job dies before running anything.

### Where it came from

My new test fixture passed sqlite's `:memory:` sentinel as `memory.db_path`. But `memory/store.connect` treats `db_path` as a filesystem path and **creates** it — so instead of an in-memory database I got a real file named `:memory:`, and my `git add -A` committed it.

### Why my diagnosis was wrong

Two mistakes, both mine:

1. **I read only the tail of the log.** I fetched the last 40 lines of a 143-line log; `error: invalid path ':memory:'` sat just above that window. What I saw was auth-cleanup noise and a bare exit 128, which looks exactly like a runner fault.
2. **My corroborating evidence was true but irrelevant.** #1216 does pass the same Windows jobs on the same base — because #1216 does not contain this file. "A sibling PR is green" says nothing about a defect introduced by *this* diff. I treated it as though it did.

The "10–14 seconds, before any test body ran" observation was accurate and pointed at checkout. I should have read the checkout step rather than inferring infrastructure from its absence.

### The fix

- The fixture now uses a real path under `tmp_path`. Verified after the change: a full suite run leaves no untracked file behind.
- `tests/test_repo_hygiene.py` gains a guard over the tracked path list — Windows-forbidden characters `<>:"|?*` and the reserved device names (`CON`, `PRN`, `AUX`, `NUL`, `COM1-9`, `LPT1-9`). It runs on **every** platform, so a Linux-only machine catches this class before a Windows runner does. That module's docstring gives the reason it belongs in `tests/`: the three test legs are release gates, while `lint.yml` is advisory end to end.

Mutation-checked against the real artifact — re-staging the file fails the guard with `:memory:  (forbidden on Windows: :)`, and removing it passes.

Checked the sibling branches: `claude/authn-unique-violation`, `claude/memory-metadata-prompt-isolation` and `main` carry no such path. Only this branch was affected.

Suite on `19a2b2f`: **5037 passed, 73 skipped, 0 failed.**

---

cgfixit:
# load /fable-protocol and /ponytail

Ignore this: 2:54

cgfixit:
## Review

#1207 made the harness `facts` echo `is True`. The siblings were the larger half. `memory:` has no validator, so a quoted YAML value boots. Under `bool()` every non-empty string is truthy — including `"true"` and `"false"` — so both consoles reported write/export capabilities **on** while `gate_memory` (`is True`) 404'd them.

| Echo | Before | Gate |
|---|---|---|
| `GET /memory/status` (`mirror.status_dict`) | helper for facts, `is True` for master, `bool()` for the other five | `is True` / `is not True` |
| harness `rag_flags` | master `enabled` still `bool()` | six gates read master strictly |

`propose_apply` and `export_html` are the write and export gates. Status was the only operator feedback and it was the thing lying. Behavior of the gates does not change; only the echo becomes truthful.

### What the diff does (4 files, +171/−8)
- `memory/mirror.py` + `harness/memory_notes.py`: `bool(...)` → `is True` on every remaining memory-flag echo.
- `tests/test_config_flag_echo_strictness.py`: 10 value shapes × both echoes; asserts `echo is (value is True)` (agreement with the gate, not a pinned constant).
- `tests/test_repo_hygiene.py`: tracked-path guard for Windows-forbidden characters and reserved device names.

The first Windows red on this branch was a file literally named `:memory:` (fixture passed sqlite’s sentinel as `db_path`; `memory.store.connect` treats that as a filesystem path). Fixed in `19a2b2f`. Current head’s Windows jobs are green.

### Leave in this PR
- No `validate_memory_config`. Correct scope. A validator is a separate PR if wanted.
- No repo-wide `bool()` → `is True` sweep. Only the memory flag set was verified against its gates.
- Quoted `propose_apply: "true"` will flip the status payload from on → off. The route already 404'd. That is the intended honesty, not a regression.

### Out of scope — do not land here
Harness `_enforce_same_origin` still hostname-only (#1205 missed `harness/server.py`). Separate PR, same residual as #1216.

Independent of #1216 / #1217 / #1219.

PR #1219, opened 2026-08-29
[tests] - guard authn_manager against new SELECT-then-INSERT pairs that drop the unique violation

qwoifhqgbio:
## Proposed changes

The check-then-act class #1216 fixes had appeared **twice in one file** and been found by review both times (#940 caught the last-admin variant a year of commits earlier; the audit caught these two). Nothing in the repo flags the pattern, which is why it kept being re-discovered instead of caught. This makes the third discovery automatic.

### The rule

In `utils/authn_manager.py`, a function that executes a SELECT and later an INSERT must either wrap the INSERT in a `try` with a **real `except` handler** — a `try/finally` with no handler does not count, it re-raises, so the raw backend error still escapes — or fold the conflict into the SQL itself (`ON CONFLICT` / `OR IGNORE` / `OR REPLACE`).

The reasoning that keeps it honest: **the pre-check SELECT is the confession.** Its author believed a uniqueness constraint exists; therefore the violation is reachable by a concurrent writer (the gateway, the harness, and `cyclaw-user` each hold their own `AuthManager` on the same DB file); therefore the INSERT must be defended. An INSERT with no pre-check makes no such claim and is out of scope.

### Why the scope is one module, stated as a decision rather than left as a gap

A full inventory of **every INSERT-executing function in the repo** (authn_manager, authn_store, personality, ratelimit, memory/store, vector_store, sqlconnect) found:

- Every instance of the bug class lives in `utils/authn_manager.py` — the module three OS processes write concurrently.
- Three functions elsewhere are **structurally identical to the bug yet safe only because of schema facts**: `personality._load_soul` (autoincrement PK, no UNIQUE column), `memory/store._insert_fact_conn` (rowid alias nobody supplies), `memory/store.stage_episode` (a plain, non-unique index).
- Two more hide their conflict handling where a call-site rule can't see it: `ratelimit._persist` (the `ON CONFLICT` is inside a *method-call return*), `memory.apply_proposal` (the INSERT is reached through a helper).

Exonerating those mechanically requires a DDL oracle plus indirection resolution — a static analyzer, not a repo guard. The guard's docstring names all of them so the next reader knows widening the scope is a real project and why it wasn't built: it earns its keep only if the class ever appears in a second module.

### The two accepted names, and their expiry

On today's `main` the guard fires on exactly `create_user` and `create_device_token` — **the two real #1216 bugs, verified live**. They ride a temporary accepted set whose comment orders its own deletion once #1216 merges. The assertion is subset-of-accepted, so #1216 landing turns the guard **quiet, not red** — no merge-order coupling. A companion test additionally refuses the half-fixed state (one of the pair fixed without the other).

**Invariant / Governance Impact:** none. Test-only — zero production diff; no route, graph edge, sanitizer pattern, soul path, config value or dependency. `invariant-guard` 47/47.

## Types of changes

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation Update
- [x] Other — a static guard pinning a bug class that has recurred

**Scope note:** One new test module. No production file.

## Benefits / why

- **The class stops depending on review to be found.** #940, then the audit, each found one instance by human reading. The next `create_*` method written in the bug shape fails CI on every platform instead.
- **The detector can't rot silently.** Its behavior is pinned by six synthetic-source tests — the bug shape, a real handler, the `try/finally` trap, folded SQL, no-precheck inserts, and the `self._sql_*` f-string indirection the module actually uses — so the mutation checks are permanent tests, not a one-off run described in a PR body.
- **It documents the survey.** The three DDL-exonerated look-alikes are named in the docstring; the next auditor who greps for this pattern gets the triage for free.

## Risks to monitor

- **The accepted set must actually be deleted when #1216 merges.** The guard cannot force that (deleting it automatically would mean failing `main` on #1216's merge, which is worse). The entries' comment orders it, and I am watching both PRs.
- **Module-scoped means new modules are uncovered.** Deliberate — see above. If a second module grows hand-written SQL with uniqueness pre-checks, that is the trigger to widen.
- **`ast.walk` descends into nested functions**, so a SELECT in an inner def counts toward its enclosing function. `authn_manager` has no such nesting today; if it grows some, the guard errs toward flagging, which is the safe direction.

## Checklist

- [x] Preserves all 6 security invariants and I6 module isolation — test-only; `invariant-guard` **47 passed, 0 failed**
- [x] Full suite green: **4954 passed, 73 skipped, 0 failed** (baseline on `main` at `6092171`: 4946/73/0 — this branch is +8, all new)
- [x] No new external network dependencies — the detector is stdlib `ast` + `pathlib`
- [ ] agentic/fsconnect/harness write path — n/a
- [x] Relevant docs updated — n/a; the guard's docstring is the documentation
- [x] `ruff check --select E,F,I,B,C4,UP,S .` clean · working tree left no stray files after the suite run (the #1218 `:memory:` lesson)

### Verified against both trees, not one

| Tree | `find_unguarded_pairs(...)` returns |
|---|---|
| `main` (`6092171`) | `['create_device_token', 'create_user']` — exactly the accepted set |
| #1216's branch file swapped in | `[]` — guard quiet, all 8 tests still pass |

The second row is the important one: it proves the guard and #1216 compose in either merge order.

### ELI5

Twice now, someone wrote code that asks "does this name already exist?" and then creates it — and twice, a reviewer had to notice that another program can sneak in between the question and the creation. The fix pattern is known and sits in the same file. This adds a small automatic check that reads that file's code on every test run and fails if anyone writes the ask-then-create shape again without the protection. It was tested by feeding it the real bug (it catches it), the real fix (it stays silent), and four tricky in-between shapes.

---

### DIFF
--- a/tests/test_check_then_act_guard.py
+++ b/tests/test_check_then_act_guard.py
@@ -1,0 +1,302 @@
+"""Static guard: a SELECT-then-INSERT pair must handle the unique violation.
+
+The bug class (PR #1216): a function pre-checks existence with a SELECT, then
+INSERTs bare. The in-process lock around the pair orders only threads sharing
+one manager -- the gateway, the harness and the cyclaw-user CLI each hold their
+own AuthManager on the same DB file -- so a writer landing between the two
+statements turns the constraint into a raw backend error
+(sqlite3.IntegrityError, or psycopg's UniqueViolation; the two share no base
+class). Raw, it escaped gate_auth's error ladder as a 500 and broke the CLI's
+documented exit-code contract. The pre-check SELECT is the confession: its
+author believed a uniqueness constraint exists, so the violation is reachable
+and the INSERT must either sit in a ``try`` with a real handler or fold the
+conflict into the SQL itself (``ON CONFLICT`` / ``OR IGNORE`` / ``OR REPLACE``).
+
+Scope is ``utils/authn_manager.py`` ONLY, and that is a measured decision, not
+an oversight. A full-repo inventory of every INSERT-executing function
+(authn_manager, authn_store, personality, ratelimit, memory/store,
+vector_store, sqlconnect) found every instance of the bug class in this one
+module -- and found three functions elsewhere that are STRUCTURALLY identical
+to the bug yet safe only because of schema facts a cheap rule cannot know:
+
+  * ``utils/personality.py`` ``_load_soul`` -- same-table SELECT then bare
+    INSERT, but ``soul_versions`` has an autoincrement PK and no UNIQUE column.
+  * ``memory/store.py`` ``_insert_fact_conn`` -- capacity-check SELECT then
+    bare INSERT, but ``facts.id`` is a rowid alias nobody supplies.
+  * ``memory/store.py`` ``stage_episode`` -- ``idx_episodes_query_hash`` is a
+    plain index, not UNIQUE.
+
+Flagging those would be false positives; exonerating them mechanically needs a
+DDL oracle (parse every backend's schema, map tables to violable constraints,
+match INSERT targets and SELECT pre-check columns against them) plus resolution
+of SQL hidden behind method calls (``ratelimit._upsert_sql()``) and helpers
+(``memory.apply_proposal``). That is a static analyzer, not a repo guard; build
+it only if the bug class ever appears in a second module. Here, every statement
+flows through ``self._sql_*`` attributes whose shape is fixed (a string literal,
+or an f-string whose leading fragment carries the verb), so a ~60-line detector
+reaches zero false positives.
+
+Lives in tests/ for the same reason tests/test_repo_hygiene.py gives: the three
+test legs are release gates, while lint.yml is advisory end to end, so a guard
+that must block a merge has to run here.
+"""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+_TARGET = Path(__file__).resolve().parent.parent / "utils" / "authn_manager.py"
+
+# Violations the guard accepts on the CURRENT tree, each because a fix already
+# exists on an unmerged branch. These are the two real bugs this guard was
+# built from; both are fixed on claude/authn-unique-violation (PR #1216), which
+# wraps each INSERT in try/except + _is_unique_violation. The entries exist
+# only so this guard can land independently of that PR's merge order -- once
+# #1216 is on main the pairs read as guarded, these names stop matching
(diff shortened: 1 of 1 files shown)


cgfixit:
## Review

The check-then-act class in `utils/authn_manager.py` was found by review twice (#940 last-admin CAS; #1216 unique-violation mapping). Nothing in CI flagged the next `create_*` written the same way. This is the detector, not the fix.

Rule, one module: a function that `execute`s a SELECT and later an INSERT must wrap the INSERT in a `try` that has a real `except` (try/finally alone does not count) or fold the conflict into the SQL (`ON CONFLICT` / `OR IGNORE` / `OR REPLACE`). The pre-check SELECT is treated as the confession that a uniqueness constraint exists.

Scope is `authn_manager` on purpose. A repo-wide inventory put every instance of the *bug class* there. Three look-alikes elsewhere (`personality._load_soul`, `memory._insert_fact_conn`, `stage_episode`) are structurally identical but safe only because of schema facts a cheap AST rule cannot know. Widening needs a DDL oracle; do not build it until the class appears in a second module.

### What the diff does (1 file, +302)
- `tests/test_check_then_act_guard.py`: `find_unguarded_pairs` over `utils/authn_manager.py`, resolving `self._sql_*` literals and f-string verbs.
- Live assertion is subset-of-accepted, so #1216 landing makes the guard quiet rather than red. Either merge order works.
- Six synthetic sources pin detector behavior: bug shape, real handler, try/finally trap, folded SQL, no-precheck INSERT, `_sql_*` indirection.
- Companion test refuses a half-fix (one of `create_user` / `create_device_token` without the other).

On current main the detector returns exactly `{create_user, create_device_token}` — the two #1216 bugs.

### Leave in this PR
- The accepted set is temporary. After #1216 merges, **delete** `_ACCEPTED_UNTIL_1216_MERGES`. The guard cannot delete itself without failing main on #1216’s land.
- The detector only checks that an `except` exists. A `try/except: rollback(); raise` still escapes as a raw 500. Typed mapping to `AuthUserExists` remains #1216’s job. Acceptable for a ~60-line AST rule; do not pretend this replaces that mapping.
- Module-scoped. A second module growing hand-written uniqueness pre-checks is the trigger to widen.

### Out of scope
Do not fold #1216’s production fix into this PR. Do not scan `personality` / `memory` / `ratelimit` here.

Follow-up after #1216: empty the allowlist in a one-line commit so the guard is live, not grandfathered.

PR #1306, opened 2026-09-04
[infra] - Drop two things nothing calls: the tzdata runtime pin and an orphan deepagent shim

cgfixit:
## Proposed changes

Output of a `/CyClaw-Optimize` pass run under `/ponytail` discipline. Both changes are the same finding in two places: something declared, shipped and maintained that has **no caller**.

**1. `tzdata==2026.2` was a runtime dependency for a test-only import.**

Nothing outside `tests/` imports `zoneinfo`. `opentweet/runner.py:34,168,170` uses `datetime.now().astimezone()`, which reads the host's local offset and needs no IANA tzdb. The pin's own comment named the OpenTweet scheduling *tests* as the reason it existed — yet it shipped in the wheel and the Docker image.

- `pyproject.toml` — moved from `dependencies` into `[project.optional-dependencies].test`
- `requirements.txt` — removed
- `requirements-test.txt` — added
- `constraints.txt` — **pin kept**, comment corrected. That file is a version ceiling, not an install list.
- `environment.yml` — **untouched.** conda-forge splits the package (`python-tzdata`) and the conda env is a test surface; its existing comment explains the split.

After the move, `tzdata` sits in exactly the shape `pytest-cov` already has (`pyproject` ✓ / `constraints` ✓ / `requirements.txt` ✗ / `req-test` ✓).

**2. `GitHubCodingRunner.build_optional_deepagent` had no caller anywhere in the repo.**

Its body was a single forwarded call to `build_deepagent_github`, and it was the sole reason the runner imported `DeepAgentBuildResult`, `build_deepagent_github`, `ProposerWorkspaceTools` and `SkillRegistry`. Removing it also removes the runner's import edge into the retired `deepagent_github` builder.

The builder itself is **untouched** — the owner decision of 2026-07-31 keeps that code in place.

**Invariant / Governance Impact:** none of the 6 invariants or I6 isolation are affected. No graph node, edge, entry point, gate, sanitizer pattern or soul path is touched. Neither file changed is in the core six. `invariant-guard` re-run: **47 passed / 0 failed**.

---

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Invariant / Governance refinement

[...]

PR #1307, opened 2026-09-04
[docs] - Reconcile 33 doc claims the code contradicts, including one that sends agents to a directory that does not exist

cgfixit:
## Proposed changes

Output of a thorough `/doc-sync` pass. **Code is the source of truth; every fix here moves the doc, never the behavior.** Nothing in this PR changes a config value, a graph edge, a route, or `settings.json`.

**The headline:** `doc_sync.py` exits `0 — "0 drift item(s) found"` on this tree, while all 33 items below were live. Its scope gaps are addressed in a separate PR.

**The worst single line**, `docs/ARCHIVE_AND_ROADMAP.md`, told agents *not* to write session notes to `docs/work/SESSION_NOTES.md` and to use `.claude/session-notes/` instead — attributing that to `PROJECT_RULES.md`. Every clause was false: `.claude/session-notes/` does not exist, `PROJECT_RULES.md:118` names `docs/work/SESSION_NOTES.md` as the active log, and `CLAUDE.md` §7/§10 both point there. An agent following it wrote nowhere.

Seven commits, one per doc family, each reviewable alone:

| Commit | Family | Items |
|---|---|---|
| `a302f3c` | CLAUDE.md | 6 |
| `fd211b1` | AGENTS.md | 1 |
| `6c19103` | PROJECT_RULES.md | 4 |
| `f8e0853` | README.md | 2 |
| `7a3ceca` | `.claude/commands/` + skill/codex mirrors | 10 |
| `f158e4f` | `docs/` | 9 |
| `5465baa` | guardrails honesty annotation | 1 |

Selected items:

- **CLAUDE.md contradicted itself on `lint.yml`.** `:761` called "the whole `lint.yml` job" advisory; its F/B/S step carries no `continue-on-error` and does block — as `:543` and `:615` already said. The same sentence's blocking-gate list was also incomplete: `verify-skills` is the **only** job in `ci.yml` with job-level `continue-on-error`.
- **`gate_auth.py` was described as "the three `/auth/*` endpoints".** It registers **15**, and the route table 60 lines above lists them all.
- **`audit.md` documented a pipe that silently does nothing.** `grep … | python3 -m metrics` — `metrics.py` has no argparse and never reads stdin; `print_metrics()` resolves the log itself from `cfg["logging"]["audit_file"]`, so the pipe re-reads the whole file and reports **unfiltered numbers as if they were filtered**.
- **`status.md` showed `Health status: healthy`.** `gate.py:1255` emits `ok` or `degraded` and never `healthy` — an agent comparing to the template would flag a healthy server as anomalous.
- **Four separate "10-node graph" claims.** `graph.py` has 12 `add_node()` calls; the two missing are `pre_action_hook_grok` and `pre_action_hook_claude`.
- **Three docs pointed at `/run-cyclaw` and `/sandbox-runtime-verification`** for the install step. Neither exists anywhere on disk.
- **`PROJECT_RULES.md:109`** asserted "the stop hook blocks `--force-with-lease`" as a repo control. `.claude/settings.json` registers no `Stop` hook. **This is precisely the drift `doc_sync.py`'s D6 check exists to catch and missed** — D6 only reads `CLAUDE.md`.
- **`THREAT_MODEL.md` had two amendments both headed "Ninth"**, the second at `##` level outside the sequence. Renumbered to Fourteenth, demoted to `###`, moved in with its siblings — line count unchanged at 1046, a pure move.
- **`THREAT_MODEL.md:65`** said the rate limit covers every route "except the two" probes. `gate.py:514,516` leave the `/static` mount and `GET /` unlimited too, so it is four.

**Invariant / Governance Impact:** **none.** No graph node, edge, entry point, gate, sanitizer pattern, config value or soul path is touched. The only non-Markdown files changed are two `guardrails/` module **docstrings** (no code, no imports). `invariant-guard` re-run: **47 passed / 0 failed**.

---

## Types of changes

- [ ] Breaking change
- [x] Documentation Update

**Scope note:** Docs + audits. Core graph/gate/soul path untouched.

---

## Benefits / why

Drift in this repo is not cosmetic — agents follow these files literally. Three of the fixes above would each cause an agent to do the wrong thing: write session notes into the void, report a healthy server as broken, or report unfiltered audit numbers as a filtered window. The rest close the gap between what the manuals promise and what the code does, which is what makes the manuals worth reading at all.

The `guardrails/` annotation is your "keep, but mark it honestly" call: `boundary.py`'s docstring promised consumers "in later phases", but Phases 2a/3/4/5 all shipped without adopting it (Phase 5 went out as `utils/tool_broker.py`). Nothing is deleted — a reader can now tell that from the file instead of inferring live architecture.

---

## Risks to monitor

- **This PR should merge before the `doc-sync` checker PR.** That PR teaches D6 to read `PROJECT_RULES.md`; run against `main` as it stands today, the sharpened check would correctly fail on the very line this PR fixes.
- Two items are judgement calls rather than clear-cut, flagged rather than hidden:
  - The `ci.yml` blocking-gate rewording reads `continue-on-error` from the workflow. I cannot read branch-protection **required checks**, so if "blocking gate" was meant in that narrower sense, the new wording is still true but describes a different thing. Reworded to be unambiguous either way.
  - `docs/README.md` says "Pointers only", so its incompleteness may be deliberate; I added the M5 doctrine row because `doc-sync`'s D7 treats that file as load-bearing.
- Four items were **deliberately not actioned** — they need a code/`settings.json` change, which `CLAUDE.md` §7 puts on the always-ask-first list. See the comment below.

---

## Checklist

- [x] Preserves all 6 security invariants and I6 module isolation — `invariant-guard` 47 passed / 0 failed
- [x] Sandbox validation run — see table
- [x] Docs-only PR; Benefits + Risks provided as required

**Verification run:**

| Gate | Result |
|---|---|
| `doc_sync.py` | 0 drift items |
| `doc-sync/verify.sh` | OK — detection self-test PASS, both D7 adjacency sub-tests PASS |
| `invariant-guard/check_invariants.py` | 47 passed, 0 failed |
| `ruff check --select F,B,S .` | All checks passed |
| `config-guard/check_config.py` | 0 failures, 1 pre-existing warning (C9 armed posture) |
| `pytest tests/test_guardrails_boundary.py tests/test_qwen_registry.py tests/test_terminal_contract.py` | exit 0 |
| Full `pytest tests/ -q` (same tree, PR #1306 run) | exit 0 |

Every path, route count, line number and file added to a tree listing in this PR was confirmed to exist on disk.

## Merge topology

- **Independent** — cut from `origin/main`, shares no file with #1306.
- **Merge this before the `doc-sync` checker PR** (ordering reason above).

https://claude.ai/code/session_01G1rWdgSUd4dWgbNBV9FsSh

---
_Generated by [Claude Code](https://claude.ai/code/session_01G1rWdgSUd4dWgbNBV9FsSh)_


cgfixit:
## Four items deliberately NOT actioned — they need a code / `settings.json` decision

`CLAUDE.md` §7 puts these on the always-ask-first list, and the doc-sync contract is to flag a mismatch that needs a code change rather than paper over it with a doc edit. No PR in this session touches any of them.

**A. `.claude/settings.json`'s `UserPromptSubmit` hook points at a script that does not exist.**

```
python3 .claude/skills/fable-protocol/context_gate.py 2>/dev/null || true
```

`ls .claude/skills/fable-protocol/` returns **only `SKILL.md`**. The `2>/dev/null || true` tail means it fails silently on *every* prompt rather than erroring, so nothing surfaces it. `.claude/README.md:120-129` already documents this exact condition and explicitly defers it for the same reason.

The doc is right and the tree is wrong, so there is nothing to fix in a doc pass. **Decision needed: unwire the hook, or restore `context_gate.py`?**

**B. No `Stop` hook is wired, and the intent is ambiguous.**

`6c19103` reconciles `PROJECT_RULES.md:109` to reality (the enforcement, where it exists, comes from an external session runtime — matching what `CLAUDE.md` §10 already said). But if the intent is that *this repo* should enforce committer-email and force-push policy itself, that is a `settings.json` change and the doc fix is the wrong direction. Say so and I'll swap it.

**C. `.claude/hooks/session-start-sync-check.sh` exists but is not registered.**

No drift here — `CLAUDE.md:841-843` correctly hedges with "if wired", and `.claude/README.md:63-66` states the situation accurately. Raising it only because the script has sat unwired long enough to deserve a wire-or-delete call.

**D. `fable-5.1-cc` has no `.claude/commands/` wrapper.**

It is the only skill without one, while `CLAUDE.md:828-835` documents `/fable-5.1-cc` as a usable entry point. This PR takes the doc-only direction — `.claude/README.md`'s "the wrapper set now covers all skills" claim is corrected. Creating the wrapper is a new file on the command surface, so that direction is yours to choose.

---

One more thing worth a decision, from the same session's ponytail scan rather than the doc pass:

**`guardrails/boundary.py` + `qwen_registry.py` (issue #1134 Phase 1) have no consumer.** ~200 lines plus `qwen_manifest.yaml`, reachable only from their own tests. Phases 2a/3/4/5 all shipped and none of them adopted the types — Phase 5's name gate went out as `utils/tool_broker.py`, and `guardrails/profiles.py` mirrors `GuardrailStage` by hand with an explicit "do not import it here". You chose **keep, but mark it honestly**, which is what `5465baa` does. Deleting it remains available later; the annotation makes that call easier to revisit because the file now says what it is.

---

PR #1349, opened 2026-09-06
[fix] - keep torch pinned in the macOS stripped constraints copy

cgfixit:
## Proposed changes

`pip install ... --ignore-installed PyYAML` reinstalls **every** resolved package, not just PyYAML: `--ignore-installed` is a bare pip flag (`-I`), and the `PyYAML` after it is parsed as one more requirement. The macOS install recipe removed the torch line from its constraints copy entirely (`grep -v '^torch=='`), so that reinstall re-resolved torch with no constraint and replaced the explicit `torch==2.13.0` installed one line earlier with PyPI's newest release.

Reproduced 2026-09-06 in a Claude Code sandbox via `.claude/skills/cyclaw-gotchas/driver.sh venv` (same recipe shape as the installer's Darwin branch): the driver installed `torch==2.13.0`, ran the stripped-constraints install, and the venv ended with `torch 2.14.0+cu130`. `pip show torch` reports it as pulled in by `sentence-transformers` (`torch>=1.11.0`).

The fix rewrites the constraints copy with only the `+cpu` suffix dropped, so the pin survives:

```sh
sed 's/^\(torch==[0-9][0-9.]*\)+cpu$/\1/' constraints.txt > /tmp/constraints-macos.txt
```

Portable BRE (no GNU-only flags; `+` is literal), version-agnostic so the pin lives only in `constraints.txt`. Applied to every copy of the recipe so they stay textually parallel and the docs' "same thing CI's macos-latest leg runs" claim stays true:

- `macos/install-cyclaw.sh` Darwin branch: the surface where the pin was actually lost (uses `--ignore-installed`).
- `.claude/skills/cyclaw-gotchas/driver.sh`: where it was caught; scratch file renamed `constraints-notorch.txt` to `constraints-plain-torch.txt` to match what it now contains.
- `.github/workflows/ci.yml` and `pip-audit.yml` macOS legs: **not broken today** (neither passes `--ignore-installed`, so pip keeps the 2.13.0 it installed). Changed so the constraints copy still bounds torch if a transitive ever asks for a newer one, instead of silently floating.
- `README.md`, `setup-guide.md`, `CLAUDE.md` (§4 trap text + §8 recipe), `docs/HARNESS_MACOS.md`: operator/agent copies of the same lines.
- `tests/test_macos_scripts.py`: new `test_installer_macos_constraints_copy_keeps_torch_pinned` extracts the installer's exact `sed` expression, runs it against the real `constraints.txt`, and asserts the pin survives without `+cpu` and every other line passes through unchanged. Fails against the previous installer (verified by mutation).
- `.claude/skills/cyclaw-gotchas/SKILL.md`: records the trap with its evidence.

Linux/Windows are unaffected: their constraints keep `torch==2.13.0+cpu`, so the reinstall re-fetches the pinned CPU wheel.

**Invariant / Governance Impact:** none. No core path, graph, gate, soul, sanitizer, or import-structure change. `python3 .claude/skills/invariant-guard/check_invariants.py` 47 passed; dep-guard, config-guard, doc-sync, env-drift all clean.

## Types of changes

- [ ] New feature
- [ ] Breaking change
- [x] Documentation Update

Scope: Infrastructure / CI only + macOS installer + docs.

## Benefits / why

macOS is the primary target and the only surface where the documented torch pin did not hold. An operator following the installer or any of the three doc recipes ran whatever torch PyPI served that day, so the CVE-2025-32434 posture and the reproducibility the `+cpu` pin exists for were not actually enforced there. After this change every install surface ends on `torch==2.13.0`.

## Risks to monitor

- If a future pin move changes the `torch==X.Y.Z+cpu` line shape in `constraints.txt` (a comment on the same line, a different local tag), the `sed` no longer matches and the new test fails loudly rather than the pin silently disappearing. That is the intended failure mode.
- BSD `sed` on stock macOS: the expression uses only basic-regex groups and a literal `+`; `bash -n` passes and the test runs the expression through bash on all three CI OSes.
- `pip-audit.yml`/`ci.yml` macOS legs now carry a `torch==2.13.0` constraint for the already-installed torch. pip treats an installed, matching version as satisfied, so no re-download.

## Checklist

- [x] Preserves all 6 security invariants and I6 (no core files touched)
- [x] `GROK_API_KEY=dummy pytest tests/test_macos_scripts.py tests/test_setup_from_clone.py tests/test_installer_python_contract.py tests/test_macos_fsconnect_setup.py tests/test_uninstall_cyclaw.py` — 88 passed, 4 skipped
- [x] `ruff check --select F,B,S .` clean; `bash -n` on both scripts; both workflows load as YAML; `cyclaw-gotchas/verify.sh` OK
- [x] No new dependency or network assumption
- [x] Docs updated where the recipe is quoted

## Merge topology

- independent (cut from `origin/main`, base `main`)
- merge order: this PR first; the sibling `[fix] - exclude the fsconnect skip-cache staging sibling from Dropbox sync` shares no files with it

https://claude.ai/code/session_011c5YiuNSqrDEb1FqN2e3gx

---
_Generated by [Claude Code](https://claude.ai/code/session_011c5YiuNSqrDEb1FqN2e3gx)_

### DIFF
--- a/.claude/skills/cyclaw-gotchas/SKILL.md
+++ b/.claude/skills/cyclaw-gotchas/SKILL.md
@@ -58,7 +58,8 @@ session 2026-09-06):
    fallback costs ~2 GB and ~7 minutes; disk had 30 GB free.
 4. Installs `requirements.txt` + `requirements-test.txt` with the `torch==`
    and `--extra-index-url` lines stripped (same shape as the macOS recipe in
-   CLAUDE.md §8), constrained by a torch-stripped copy of `constraints.txt`.
+   CLAUDE.md §8), constrained by a copy of `constraints.txt` that keeps the
+   torch pin minus its `+cpu` suffix (see the `--ignore-installed` gotcha).
 
 Re-running is idempotent: it exits 0 immediately once
 `import torch, chromadb, langgraph, pytest` succeeds.
@@ -139,6 +140,16 @@ expire it.
   deps live in 3.11's `dist-packages`; on 2026-09-06 `import torch` failed for
   3.10, 3.11, 3.12 and 3.13 alike. Treat the venv step as mandatory every
   session; the container is rebuilt from a generic image, not from this repo.
+- **`--ignore-installed PyYAML` reinstalls everything, not just PyYAML.**
+  `--ignore-installed` is a bare pip flag; `PyYAML` after it is one more
+  requirement. Every install line in this repo that carries it therefore
+  re-resolves and reinstalls torch too, and the only thing holding the
+  version is the constraints file. The macOS recipe used to strip the torch
+  line from its constraints copy, so that reinstall floated torch to PyPI's
+  newest: this driver installed `torch==2.13.0` and ended with `2.14.0+cu130`
+  (2026-09-06; `macos/install-cyclaw.sh`'s Darwin branch had the same shape).
+  Rule: the constraints copy keeps `torch==X.Y.Z` with only `+cpu` removed;
+  `tests/test_macos_scripts.py` pins it.
 - **`pkill -f "python gate.py"` kills your own shell, and `pgrep -f gate.py`
   reports it as still running.** Both match the `bash -c` wrapper that is
   executing the command (exit 144 from `pkill`; a phantom "server still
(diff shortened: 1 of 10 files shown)


cgfixit:
@codex review

